PR #9, opened 2022-03-28
✨ Update package names and other references

laurentsimon:
Update package names and other references of the original repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,1 +1,1 @@
-# slsa-provenance
+# slsa-verifier
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/gossts/slsa-provenance
+module github.com/slsa-framework/slsa-verifier
 
 go 1.17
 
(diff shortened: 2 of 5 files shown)


joshuagl (approved):

TomHennen:
This only verifies provenance that comes from slsa-github-generator, right?

Would it be possible to change the name to slsa-github-verifier to make that clear?

MarkLodato:
> This only verifies provenance that comes from slsa-github-generator, right?
> 
> Would it be possible to change the name to slsa-github-verifier to make that clear?

Personally I think that it makes more sense to have a single verifier rather than one per producer. Right now it only verifies the one, but I think we should expand it in the future to support more.

TomHennen:
> > This only verifies provenance that comes from slsa-github-generator, right?
> > Would it be possible to change the name to slsa-github-verifier to make that clear?
> 
> Personally I think that it makes more sense to have a single verifier rather than one per producer. Right now it only verifies the one, but I think we should expand it in the future to support more.

That makes sense.  Perhaps the README.md could be updated to indicate the current support?

PR #11, opened 2022-03-29
fix(codeql): fix branch wildcard

joshuagl:
* is a special character in YAML, so we must use quotes
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet


laurentsimon (approved):
Thanks!

PR #13, opened 2022-03-29
✨ Option to verify branch

laurentsimon:
The default branch is set to `main`, and can be overwritten via `--branch xxx`. We check that the branch in the provenance information of the form `refs/heads/xxx`


pkg/provenance.go
```diff
@@ -393,8 +395,148 @@ func VerifyProvenance(env *dsselib.Envelope, expectedHash string) error {
...
+	if version != 1 {
+		return "", fmt.Errorf("%w", errorInvalidVersion)
+	}
+
+	// RefType.
+	refType, err := getAsString(parameters, "ref_type")
+	if err != nil {
+		return "", err
+	}
+	// RefType.
```
asraa, line 522: nit: fix comment

pkg/provenance.go
```diff
@@ -393,8 +395,148 @@ func VerifyProvenance(env *dsselib.Envelope, expectedHash string) error {
...
+	}
+
+	switch refType {
+	case "branch":
+		return ref, nil
+	case "tag":
+		return baseRef, nil
+	}
+
+	return "", fmt.Errorf("%w: %s", errorInvalidDssePayload,
```
asraa, line 540: can you combine this?
```
fmt.Errorf("%w: %s %s", errorInvalidDssePayload, "unknown ref type", refType)
```

asraa (changes requested):

pkg/provenance.go
laurentsimon: done.

pkg/provenance.go
```diff
@@ -393,8 +395,137 @@ func VerifyProvenance(env *dsselib.Envelope, expectedHash string) error {
...
+	if !ok {
+		return "", fmt.Errorf("%w: %s", errorInvalidDssePayload, "parameters type payload")
+	}
+
+	bRef, ok := payload["base_ref"]
+	if !ok {
+		return "", fmt.Errorf("%w: %s", errorInvalidDssePayload, "parameters type base_ref")
+	}
+
+	baseRef, ok = bRef.(string)
```
asraa, line 483: can you `getAsString` the baseRef?

laurentsimon: that does not work. getAsString needs a map, the input is just a string. Right?

asraa: getAsString(payload, 'base_ref')?

laurentsimon: done, thanks!

asraa (approved):

Issue #7, opened 2022-03-25
Feature: add `--tag` option


laurentsimon:
Consumes may want to verify the tag of the repo that was used. This is useful to avoid version swapping, e.g. an attacker using an old vulnerable version of the same software (e.g., v1.2.3) and claiming to be v1.2.4.

Idea: support `--tag=v1`, `--tag=v1.2` and `--tag=v1.2.3` assuming semver is used

Note that this theoretically could be left to the caller, but it may be a common case we want to support

PR #18, opened 2022-04-01
✨ Add tag and version verification

laurentsimon:
Add support for:
1. tag verification thru `--tag` option. This must be an exact match
2. versioned tag verification thru `--versioned-tag`. This uses semantic versioning matching on the tag. I've avoided `--version` because this may be used to report the version of our tool.

I separated the `--tag` and `--versioned-tag` because I wanted to make it explicit that there is versioning match.
Please advise if we should do it differently.

closes https://github.com/slsa-framework/slsa-verifier/issues/7


main.go
```diff
@@ -96,14 +117,32 @@ func main() {
 	flag.StringVar(&provenancePath, "provenance", "", "path to a provenance file")
 	flag.StringVar(&binaryPath, "binary", "", "path to a binary to verify")
 	flag.StringVar(&source, "source", "", "expected source repository that should have produced the binary, e.g. github.com/some/repo")
-	flag.StringVar(&branch, "branch", "main", "expected branch the binary was compiled from. Default: main")
+	flag.StringVar(&branch, "branch", "main", "[optional] expected branch the binary was compiled from")
```
asraa, line 120: remove the default param for main here

asraa (changes requested):

main.go
laurentsimon: why? Is this not supposed to be the default branch we should verify by default? In fact if we want that my code is wrong because it uses a pointer :-)

laurentsimon: I guess you mean this should not be an option paramater?

asraa: Yeah, it's confusing it's optional but has a default. maybe just specify that this will be default main and you don't need to check if it's nil

laurentsimon:
comments addressed, PTAL

asraa (approved):

PR #63, opened 2022-05-24
✨ Allow main branch only for trusted builder and e2e tests repos

laurentsimon:
This is part of trying to reduce the time of compilation https://github.com/slsa-framework/slsa-github-generator/issues/74

This PR allows the trusted builder and e2e repos to build using a trusted builder referenced at `@main`. This is an exception that is needed to use the builder binary generated during a release: the builder needs to build itself from main. For the e2e tests,, we also want to tests at main.
For any other repos, this is not allowed, and trusted builder referenced at a version/tag is required.


laurentsimon:
In order to fix the unit tests, we need to merge https://github.com/slsa-framework/slsa-github-generator/pull/86 and generate the provenance with the trusted builder pinned with a tag. This ie because starting with this PR, we only accept tagged's trusted builders (except for the old ones which we're going to deprecate as well once we fix the other unit tests)

laurentsimon:
Ready for review, PTAL

main_test.go
```diff
@@ -31,259 +32,260 @@ func Test_runVerify(t *testing.T) {
 		{
 			name:     "valid main branch default",
 			artifact: "./testdata/binary-linux-amd64-workflow_dispatch",
-			source:   "github.com/asraa/slsa-on-github-test",
+			source:   "github.com/laurentsimon/slsa-verifier-test-gen",
```
ianlewis, line 35: Was changing this necessary for some reason?

pkg/provenance.go
```diff
@@ -44,24 +44,32 @@ const (
...
-	errorMismatchHash         = errors.New("binary artifact hash does not match provenance subject")
-	ErrorMismatchBranch       = errors.New("branch used to generate the binary does not match provenance")
-	ErrorMismatchRepository   = errors.New("repository used to generate the binary does not match provenance")
-	ErrorMismatchTag          = errors.New("tag used to generate the binary does not match provenance")
-	ErrorMismatchVersionedTag = errors.New("tag used to generate the binary does not match provenance")
-	ErrorInvalidSemver        = errors.New("invalid semantic version")
-	errorInvalidVersion       = errors.New("invalid version")
+	ErrorInvalidDssePayload      = errors.New("invalid DSSE envelope payload")
+	errorRekorSearch             = errors.New("error searching rekor entries")
+	errorMismatchHash            = errors.New("binary artifact hash does not match provenance subject")
```
ianlewis, line 63: maybe group the exported and private errors? e.g.

```
var (
    ErrorInvalidDssePayload = ...
    ...

    errorRekorSearch = ...
    ...
)```

File: pkg/provenance.go [hunk 3]
```diff
@@ -402,6 +418,31 @@ func VerifyWorkflowIdentity(id *WorkflowIdentity, source string) error {
 	return nil
 }
 
+// Only allow `@refs/heads/main` for the builder and the e2e tests that need to work at HEAD.
+// This lets us use the pre-build builder binary generated during release (release happen at main).
+// For other projects, we only allow semantic versions that map to a release.
+func verifyTrustedBuilderRef(id *WorkflowIdentity, ref string) error {
```
ianlewis, line 424: This seems like something you'll want a test or two for.

.github/workflows/release.yml
```diff
@@ -0,0 +1,21 @@
+name: Verifier realeaser
+
+on: 
+  # For manual tests.
+  workflow_dispatch:
+  push:
+    tags:
+      - "*" # triggers only if push new tag version, like `0.8.4` or else
```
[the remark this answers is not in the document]
laurentsimon: done.

.github/workflows/release.yml
```diff
@@ -0,0 +1,21 @@
+name: Verifier realeaser
```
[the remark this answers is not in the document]
laurentsimon: thanks.

pkg/provenance.go
laurentsimon: the tests exist under provenance_test.go (Test_verifyTrustedBuilderRef).

pkg/provenance.go
```diff
@@ -44,24 +44,32 @@ const (
 	certOidcIssuer   = "https://token.actions.githubusercontent.com"
 )
 
-// TODO: remove builder.yml
+var (
+	trustedBuilderRepository = "slsa-framework/slsa-github-generator"
```
asraa, line 48: nit: replace slsa2_provenance.yml with trustedBuilderRepository prefix

asraa (approved):
+1 that we need a generator test repository that we have perms to trigger workflows

pkg/provenance.go
laurentsimon: good catch, thanks

main_test.go
ianlewis: I see. We need to re-generate them by building in a github action but I guess doing it in this repo is problematic? Maybe a workflow that is only triggered by `workflow_dispatch` would work?

We probably want to have a readme in the `testdata` directory to document how to regenerate it.

laurentsimon: Agreed on all these steps. What repo name shall we create?

ianlewis: Under ossf-tests in fine probably

Issue #74, opened 2022-05-31
Print verified provenance on stdout for piping into policy engines


asraa:
Print other info in std err

PR #76, opened 2022-06-01
feat: print verified provenance on stdout

asraa:

Fixes https://github.com/slsa-framework/slsa-verifier/issues/74

See title.


laurentsimon (approved):
Awesome!

Do you want to update the doc as well?

Issue #64, opened 2022-05-24
Add tests in main_tests.go for the builder's ref


laurentsimon:
We added an exception to allow the builder's ref to be built at HEAD https://github.com/slsa-framework/slsa-verifier/pull/63
We need to add the corresponding e2e test.

PR #78, opened 2022-06-03
e2e : Add tests in main_tests.go for using builder ref at head for e2e example-package repo

asraa:

Fixes https://github.com/slsa-framework/slsa-verifier/issues/64


laurentsimon (approved):
Thanks

Issue #73, opened 2022-05-31
Create testdata subfolders for generator version 0.0.2 release


asraa:
https://github.com/slsa-framework/slsa-github-generator/releases

PR #80, opened 2022-06-06
fix: add v0.0.2 subfolder for builder testdata

asraa:

https://github.com/slsa-framework/slsa-verifier/issues/73


laurentsimon (commented):
I'm wondering:
1. For provenance_test.go: do we need per-version tests, or it's OK to only tests at HEAD?
2. For main_test.go: shall we re-generate tests so that they come from `slsa-framework/package-example`, and use a loop:
```
for v in versions:
  artifact = sprintf(./testsdata/%v/binary-blabla, v)
  source = github.com/slsa-framework/example-package
```

This way it will work for all versions. Wdut?

RELEASE.md
```diff
@@ -8,6 +8,7 @@ This is a  document to describe the release process for the verifier.
 - [Verify provenance](#verify-provenance)
 - [Update documentation](#update-documentation)
 - [Update builders](#update-builders)
+- [Add tests for new builder release version](#update-tests)
```
laurentsimon, line 11: should this not be in the builder's RELEASE.md? We already have it mentioned, but you can add the link to the workflows u created

laurentsimon (approved):
Thanks

RELEASE.md
asraa: moving it!

main_test.go
```diff
@@ -18,6 +19,9 @@ func pString(s string) *string {
 	return &s
 }
 
+// Versions of the builders to test
+var generatorVersions = []string{"v0.0.2"}
```
laurentsimon, line 23: how about v1.0.0?

laurentsimon: we're going to need it before we cut the release

asraa: If I added it right now, it'll fail because the data isn't there yet

laurentsimon (approved):

Issue #86, opened 2022-06-08
Enhancement: Expose an option to print the full provenance, don't print by default


asraa:
Per feedback:

```
I wonder if the verifier should just print something unambiguous like 'PASSED'.  As I understand it, the verifier has already checked the expected source, branch, and tag.  What is a user expected to do with the decoded provenance statement?  Why not just cut that part (it can still be an option that the tool can support) and instead just tell the user the binary looks good?
Laurent Simon
Laurent Simon
10:38 AM Jun 6
we want to let users inspect the content if they want, e.g. if the have policies to enforce. PTAL at the `Uses cases` section further below for example. If you think this is too verbose in general, maybe we could not print by default, and have an option for users to turn it on. Without printing, users need to be aware how to extract the provenance from the provenance file (DSSE), which we're trying to avoid. Let us know what you think. /cc @[asraa@google.com](mailto:asraa@google.com)
Show less
Tom Hennen
Tom Hennen
11:01 AM Jun 6
Oh yes, I think an option to print makes a lot of sense. By default, especially for what you're trying to get across in this part of the blog post, I bet it's probably not needed (and potentially confusing).
Laurent Simon
Laurent Simon
11:16 AM Jun 6
+1, thanks for the feedback
Joshua Lock
Joshua Lock
3:42 AM Today
Agree with the suggestion to print a result by default and add an option to print the full provenance JSON.
```

PR #87, opened 2022-06-08
feat: add an option to print provenance

asraa:
see title

Fixes https://github.com/slsa-framework/slsa-verifier/issues/86

Also prints the Rekor URL of the verified entry. this was some feedback given by Brandon earlier. 

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,8 @@ $ go run . --help
     	path to an artifact to verify
   -branch string
     	expected branch the binary was compiled from (default "main")
+  -output-provenance
+    	output the verified provenance
   -provenance string
     	path to a provenance file
   -source string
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -61,6 +61,8 @@ $ go run . --help
     	path to an artifact to verify
   -branch string
     	expected branch the binary was compiled from (default "main")
+  -output-provenance
```
laurentsimon, line 64: shall we rename to `--display-provenance` or `--print-provenance`. `output-provenance` may be interpreted as a file to output the provenance.
Otherwise, shall we rename `--provenance` to `input-provenance`

I think the former looks better. Wdut?

laurentsimon (approved):

README.md
asraa: Oh, changed to `--print-provenance`

laurentsimon: LGTM

laurentsimon:
Last note: Do we want to remove all the other things printed (tlog found, etc) and only print OK when `--print-provenance` is not given as argument?

laurentsimon:
> Last note: Do we want to remove all the other things printed (tlog found, etc) and only print OK when `--print-provenance` is not given as argument?

I think this is already done (it's printed on stdout). Ignore.

PR #98, opened 2022-06-18
Refactor - Use interface for rekor

naveensrinivasan:
- Use rekor interface instead of concrete implementation so that it can
  be mocked and tested.

Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>


naveensrinivasan:
@laurentsimon 👀

pkg/mockentries.go
```diff
@@ -0,0 +1,124 @@
...
+import (
+	reflect "reflect"
+
+	runtime "github.com/go-openapi/runtime"
+	gomock "github.com/golang/mock/gomock"
+	entries "github.com/sigstore/rekor/pkg/generated/client/entries"
+)
+
+// MockEntriesClientService is a mock of ClientService interface.
+type MockEntriesClientService struct {
```
ianlewis, line 12: Maybe provide a Makefile target or instructions somewhere on how to regenerate these?

pkg/mockentries.go
```diff
@@ -0,0 +1,124 @@
+package pkg
```
ianlewis, line 1: Put generated mocks in a separate package.

ianlewis:
I'm going to close this as it's pretty old. Feel free to resubmit the PR if we need this.

PR #101, opened 2022-06-19
GitHub dependency graph

naveensrinivasan:
- https://github.blog/2022-06-17-creating-comprehensive-dependency-graph-build-time-detection/

- https://github.com/actions/go-dependency-submission

Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/dependency.yml
+++ b/.github/workflows/dependency.yml
@@ -1,0 +1,35 @@
+name: Go-Dependency-Submission
+on:
+  push:
+    branches:
+      - main
+
+permissions: read-all
+
+jobs:
+  go-action-detection:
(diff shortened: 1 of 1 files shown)


laurentsimon:
Sorry fort he late reply, I missed this PR.
Can you explain what benefits this Action provides? We have dependabot alerts already on (if not we can enable it). IFAIU, a project with a lock file does not need to upload their dependency list to GitHub, since GitHub can already scan the lock file. The Action seems intended for projects without lock file.

ianlewis:
I'm going to close this as it's pretty old. Feel free to resubmit the PR if you think we need this.

PR #128, opened 2022-07-07
fix deprecated libs

### LINKED ISSUE #61: Need to move dependency on google/trillian/merkle to transparency-dev/merkle
see https://github.com/sigstore/rekor/issues/787

naveensrinivasan:
- Refactored to move away from github.com/google/trillian/merkle/logverifier"
and github.com/google/trillian/merkle/rfc6962

- Included go mod tidy for validation

- Fixes #61

### DIFF
--- a/.github/workflows/pre-submit.yml
+++ b/.github/workflows/pre-submit.yml
@@ -24,3 +24,6 @@ jobs:
           go build -mod=vendor
           # Tests
           go test -mod=vendor -v ./...
+          # go mod tidy
+          go mod tidy
+          
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/docker/go v1.5.1-1
 	github.com/go-openapi/runtime v0.24.1
 	github.com/google/go-cmp v0.5.8
-	github.com/google/trillian v1.4.1
+	github.com/google/trillian v1.4.1 // indirect
 	github.com/in-toto/in-toto-golang v0.3.4-0.20211211042327-af1f9fb822bf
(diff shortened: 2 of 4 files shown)


.github/workflows/pre-submit.yml
```diff
@@ -24,3 +24,6 @@ jobs:
           go build -mod=vendor
           # Tests
           go test -mod=vendor -v ./...
+          # go mod tidy
+          go mod tidy
```
laurentsimon, line 28: why do we need to do that? We want to test the config used to build the builders. This may change the go.mod.
I think this should be a pre-submit verifying that go.mod is clean, not a command we use before building.

Did I understand the use case properly?

naveensrinivasan: OK, I understand. The `go.sum` wasn't updated so if not here where should this change be?

Even better why store the `vendor` folder in git similar to `k8s`. That will make the builds hermetic.

naveensrinivasan: Removed that go mod tidy.

naveensrinivasan: What if one of your deps gets deleted? How do you handle that?

laurentsimon: We don't handle it, and I think we'd use the copy from the Go module proxy if this became a problem?

naveensrinivasan: Ok.

laurentsimon (approved):
LGTM. Let's wait for @asraa to take a final look on this one before merging

naveensrinivasan:
> LGTM. Let's wait for @asraa to take a final look on this one before merging

👍

PR #129, opened 2022-07-07
refactor: refactor packages and create verification options

asraa:

It was complicated to organize OCI image verification without organizing these options and functions.

* Creates verification options for the provenance. In a follow up, I'll organize verification options on the builder identity
* Splits pkg library files into rekor/builder/provenance.

cc @ianlewis @laurentsimon


laurentsimon (approved):
Thanks

ianlewis (approved):
LGTM

PR #132, opened 2022-07-08
fix: Fixes v1.0.0 versions and below to correctly get entries by UUID

asraa:

Fixes https://github.com/slsa-framework/slsa-github-generator/issues/427


asraa:
@laurentsimon I had to make these fixes from v1.0.0: I needed to update rekor (and then also cosign) to pull in new sharding libraries, and then update logic in the entry we needed to verify.

Because v1.0.0 is a tag, I can't easily merge in.

Because v1.0.0 and below verifiers only use search index, they are effectively totally broken now.

laurentsimon:
> @laurentsimon I had to make these fixes from v1.0.0: I needed to update rekor (and then also cosign) to pull in new sharding libraries, and then update logic in the entry we needed to verify.
> 
> Because v1.0.0 is a tag, I can't easily merge in.

Is it do-able to backport the change by re-writing the commit history and re-running the release? Just curious how much work this would be.

Or we could create a new branch off v1.0.0 to backport the changes then re-run the release. Not ideal because verification would need to pass a branch... I'm not sure if we can have a branch v1.0.1 and a tag v1.0.0 :) Probably not?

I'm wondering how we can best respond to a similar problem in the future.

> Because v1.0.0 and below verifiers only use search index, they are effectively totally broken now.

is this because the artifact hash is not enough to determine the UUID to search rekor?

laurentsimon (approved):

asraa:
> Or we could create a new branch off v1.0.0 to backport the changes then re-run the release. Not ideal because verification would need to pass a branch... I'm not sure if we can have a branch v1.0.1 and a tag v1.0.0 :) Probably not?

I think yes, we should create a new branch. If we call this branch v1.0.0, and then delete the tag and create a new release for v1.0.0, would that work?

In the future, we should create a branch for each release I think. If a release has a major issue like this, then we should always be using verifier at the highest patch version anyway, right?

PR #140, opened 2022-07-11
fix: properly handle sharded uuids returned from rekor

asraa:

* fix: Correctly resolved the sharded uuid returned by rekor.
* chore: update go to v1.18. required for the pkg in rekor needed for the fix

This ports the rekor fix back to version v1.0.0. I will create a new tag for v1.0.1 after this commit.


laurentsimon (approved):
Thanks!

PR #141, opened 2022-07-11
fix: properly handle sharded uuids returned from rekor

asraa:

* fix: Correctly resolved the sharded uuid returned by rekor.

Same as #140 except for main.


laurentsimon (approved):
Thanks!

PR #143, opened 2022-07-12
fix: another sharded uuid reference

asraa:

Verification of old builders (relying on rekor search) will fail because the params.entryuuid is a shard+uuid


naveensrinivasan (commented):
How do you test this change? Manually to approve it.

PR #147, opened 2022-07-13
feat: support oci image verification

asraa:

Support OCI image verification in the GHA verifier.

Added initial testing, for valid signatures, invalid builder IDs and invalid source repositories.


naveensrinivasan (commented):
What is the process to verify the attached binaries?

The goal is to make sure the e2e tests are reliable.

asraa:
Yes @naveensrinivasan I'm still trying to figure out a testing strategy

> What is the process to verify the attached binaries?

I don't understand the question. Binaries are already e2e tested.

> The goal is to make sure the e2e tests are reliable.

Yeah. Ideally I'd be able to attach manifests for OCI images and use those to verify. I will probably need to implement a fake registry and a fake signed image for cosign. do you have other ideas?

naveensrinivasan:
> > Binaries are already e2e tested.
> 
> How were these specific binaries already e2e tested? I am trying to understand, so pardon my ignorance.

🤦 Now, I see them as renamed and not new binaries. Thanks, it makes sense. I thought these were new binaries.

main.go
```diff
@@ -82,7 +92,13 @@ func main() {
 		"print the verified provenance to std out")
 	flag.Parse()
 
-	if provenancePath == "" || artifactPath == "" || source == "" {
+	if (provenancePath == "" || artifactPath == "") && ociImageRef == "" {
```
ianlewis, line 95: I feel like we should have different commands for containers vs. artifacts

The combinations of how flags are used isn't really very user friendly and I feel like it will just get worse over time.

main.go
```diff
@@ -12,13 +13,17 @@ import (
 	"log"
 	"os"
 
+	"github.com/google/go-containerregistry/pkg/name"
```
ianlewis, line 16: nit: pick a better identifier for this package as "name" is pretty commonly used as a variable etc. Maybe `crname`?

main.go
ianlewis: ...not that I expect that to be implemented in this PR. Just a general comment.

asraa: For v2, I'll try to write a one-pager on the SLSA verifier API. There are a few issues on this (https://github.com/slsa-framework/slsa-verifier/issues/180) and I am worried without writing it up first I'll get it wrong.

asraa:
Currently working with example-test packages 
```
$ go run . -oci-image ghcr.io/slsa-framework/example-package.e2e.container.schedule.main.default.slsa3@sha256:93e556b16d8be1283f3825aa0ee595250649f56e70e18a3089d92ba29cd81e85 -source github.com/slsa-framework/example-package
Verified build using builder https://github.com/slsa-framework/slsa-github-generator/.github/workflows/generator_container_slsa3.yml@refs/heads/main at commit 8837317a8c8021fa59bb05260f363c7deb4ca63d
```

For some reason, `crane manifest`'ing the image doesn't show me where the cosign attesation layer is. Trying to retrieve some testadata manifests.

main.go
ianlewis: Just to be clear, I was simply thinking something like `slsa-verifier verify-artifact ...` and `slsa-verifier verify-image ...` something like that. Right now it's down to the combination of what flags users use. Commands allow us to have different flags for different CLI commands.

asraa: Ah, yes I agree too we should have sub-commands separating the two.

laurentsimon: SGTM

asraa:
Got tests in. Waiting for the queue of PRs https://github.com/slsa-framework/slsa-verifier/pull/187

and then I'll implement the `VerifyImage` on the builder interface.

asraa:
@ianlewis @laurentsimon This is ready now.

I had to make a few changes for testing -- remote and local images behave slightly different. See the `container` package.

I will make an issue to add more testing: malicious tests, tag verification. Just wanted to start the functionality.

cli/slsa-verifier/main.go
```diff
@@ -61,17 +70,18 @@ func main() {
 		os.Exit(1)
 	}
 
-	verifiedProvenance, _, err := runVerify(artifactPath, provenancePath, source,
+	verifiedProvenance, _, err := runVerify(artifactReference, artifactPath, provenancePath, source,
```
laurentsimon, line 73: Maybe let's separate the functions to avoid special checks depending on options passed: `runVerifyArtifactPath()` and `runVerifyArtifactReference()`?

Trying to find a way to simplify the code, at the expense of a little bit of code duplication. Mixing two different logic is prone to mistakes - happened a lot in openssl in the past.

cli/slsa-verifier/main_test.go
```diff
@@ -23,18 +30,10 @@ func pString(s string) *string {
...
-	"v1.1.1": {"go"},
-	"v1.2.0": {"generic"},
-}
-
 const TEST_DIR = "./testdata"
 
-func Test_runVerify(t *testing.T) {
-	t.Parallel()
+func Test_runVerifyGoAndGeneric(t *testing.T) {
+	// t.Parallel()
```
laurentsimon, line 36: is the comment intentional? if so, add a comment.

cli/slsa-verifier/main_test.go
```diff
@@ -23,18 +30,10 @@ func pString(s string) *string {
...
-	"v0.0.2": {"go"},
-	"v1.1.1": {"go"},
-	"v1.2.0": {"generic"},
-}
-
 const TEST_DIR = "./testdata"
 
-func Test_runVerify(t *testing.T) {
-	t.Parallel()
+func Test_runVerifyGoAndGeneric(t *testing.T) {
```
laurentsimon, line 35: Shall this be more generic and be called `Test_runVerifyArtifactPath`, and the other `Test_runVerifyArtifactReference`?

cli/slsa-verifier/main_test.go
```diff
@@ -457,3 +444,179 @@ func Test_runVerify(t *testing.T) {
 		})
 	}
 }
+
+func TestContainerVerification(t *testing.T) {
```
laurentsimon, line 448: rename to `Test_runVerifyArtifactReference`?

cli/slsa-verifier/main_test.go
```diff
@@ -457,3 +444,179 @@ func Test_runVerify(t *testing.T) {
...
+			if err != nil {
+				return "", err
+			}
+		default:
+			return "", errors.New("must verify either an image index or image")
+		}
+		return strings.TrimPrefix(h.String(), "sha256:"), nil
+	}
+
+	// t.Parallel()
```
laurentsimon, line 492: add a comment why need to disable parallelism

cli/slsa-verifier/main_test.go
```diff
@@ -457,3 +444,179 @@ func Test_runVerify(t *testing.T) {
...
+			artifact:    "container_workflow_dispatch",
+			source:      "github.com/slsa-framework/example-package",
+			pversiontag: pString("v1"),
+			err:         serrors.ErrorInvalidSemver,
+		},
+	}
+	for _, tt := range tests {
+		tt := tt // Re-initializing variable so it is not changed while executing the closure below
+		t.Run(tt.name, func(t *testing.T) {
+			// t.Parallel()
```
laurentsimon, line 572: same comment about the comment :)

cli/slsa-verifier/main_test.go
```diff
@@ -457,3 +444,179 @@ func Test_runVerify(t *testing.T) {
...
+			getVersions := func() []string {
+				res := []string{}
+				builder := "container"
```
laurentsimon, line 575: should this be more specific, in case we later support ko, jib, etc?

cli/slsa-verifier/main_test.go
```diff
@@ -457,3 +444,179 @@ func Test_runVerify(t *testing.T) {
...
+
+				builderDir, err := ioutil.ReadDir(filepath.Join(TEST_DIR, builder))
```
laurentsimon, line 577: If logic to read the directories into `res` is shared between container test function and non-container tests function, let's create a function to do it.

cli/slsa-verifier/main_test.go
```diff
@@ -457,3 +444,179 @@ func Test_runVerify(t *testing.T) {
...
+	// t.Parallel()
+	tests := []struct {
+		name        string
+		artifact    string
+		source      string
+		branch      string
+		ptag        *string
+		pversiontag *string
+		pbuilderID  *string
+		builderID   string
```
laurentsimon, line 501: It's my fault, but maybe we should rename `builderID` to `outBuilderID`.

verifiers/internal/gha/verifier.go
```diff
@@ -80,11 +72,72 @@ func (v *GHAVerifier) VerifyArtifact(ctx context.Context,
...
+	env, cert, err := VerifyProvenanceSignature(ctx, rClient, provenance, artifactHash)
+	if err != nil {
+		return nil, "", err
+	}
+
+	return verifyEnvAndCert(env, cert, provenanceOpts, builderOpts)
+}
+
 // VerifyImage verifies provenance for an OCI image.
 func (v *GHAVerifier) VerifyImage(ctx context.Context,
```
laurentsimon, line 96: here it's called "image", in the CLI it's called "reference". Let's try to use the same terminology throughout the code.

verifiers/internal/gha/verifier.go
```diff
@@ -34,23 +38,11 @@ func (v *GHAVerifier) IsAuthoritativeFor(builderID string) bool {
 	return strings.HasPrefix(builderID, "https://github.com/")
 }
 
-// VerifyArtifact verifies provenance for an artifact.
-func (v *GHAVerifier) VerifyArtifact(ctx context.Context,
-	provenance []byte, artifactHash string,
+func verifyEnvAndCert(env *dsse.Envelope,
```
laurentsimon, line 41: remind me: do we verify that the leaf cert is signed by Fulcio?

laurentsimon (commented):
LGTM overall. Excited to see this one land!!

verifiers/internal/gha/verifier.go
laurentsimon: of course we do https://github.com/slsa-framework/slsa-verifier/blob/main/verifiers/internal/gha/rekor.go#L375
Closing.

container/container.go
```diff
@@ -0,0 +1,27 @@
...
+
+var GetImageDigest = func(imageReference string) (string, error) {
+	ref, err := crname.ParseReference(imageReference)
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimPrefix(ref.Context().Digest(ref.Identifier()).DigestStr(), "sha256:"), nil
+}
+
+var RunCosignImageVerification = func(ctx context.Context,
```
laurentsimon, line 20: Why not have this under `gha/`?

verifiers/internal/gha/builder.go
```diff
@@ -19,8 +19,9 @@ var (
 )
 
 var defaultTrustedReusableWorkflows = map[string]bool{
-	trustedBuilderRepository + "/.github/workflows/generator_generic_slsa3.yml": true,
-	trustedBuilderRepository + "/.github/workflows/builder_go_slsa3.yml":        true,
+	trustedBuilderRepository + "/.github/workflows/generator_generic_slsa3.yml":   true,
```
laurentsimon, line 22: we may bucket these into container and not; and verify that the trusted builder is as expected (container vs artifact)

verifiers/internal/gha/verifier.go
```diff
@@ -80,11 +72,72 @@ func (v *GHAVerifier) VerifyArtifact(ctx context.Context,
...
+	/* Retrieve any valid signed attestations that chain up to Fulcio root CA. */
+	roots, err := fulcio.GetRoots()
+	if err != nil {
+		return nil, "", err
+	}
+	opts := &cosign.CheckOpts{
+		RootCerts: roots,
+	}
+
+	atts, _, err := container.RunCosignImageVerification(ctx,
```
laurentsimon, line 110: I would move this function into `gha/` in container.go or image.go

verifiers/internal/gha/verifier.go
```diff
@@ -80,11 +72,72 @@ func (v *GHAVerifier) VerifyArtifact(ctx context.Context,
...
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "unexpected error parsing envelope from OCI registry %s", err)
+			continue
+		}
+		cert, err := att.Cert()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "unexpected error getting certificate from OCI registry %s", err)
+			continue
+		}
+		verifiedProvenance, builderID, verifyErr = verifyEnvAndCert(env, cert, provenanceOpts, builderOpts)
```
laurentsimon, line 136: do we want to verify the subject? Our builder uses `image:tag` as subject, so this should be safe?
/cc @ianlewis

cli/slsa-verifier/main.go
```diff
@@ -9,25 +9,28 @@ import (
...
+	tag               string
+	versiontag        string
+	printProvenance   bool
 )
 
 func main() {
 	flag.StringVar(&builderID, "builder-id", "", "EXPERIMENTAL: the unique builder ID who created the provenance")
 	flag.StringVar(&provenancePath, "provenance", "", "path to a provenance file")
 	flag.StringVar(&artifactPath, "artifact-path", "", "path to an artifact to verify")
+	flag.StringVar(&artifactReference, "artifact-reference", "", "reference to an OCI image to verify")
```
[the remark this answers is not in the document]
asraa: I think we should, should I do that in this PR though? If it's pretty easy to review I can split the commands

cli/slsa-verifier/main.go
```diff
@@ -38,7 +41,13 @@ func main() {
 		"print the verified provenance to std out")
 	flag.Parse()
 
-	if provenancePath == "" || artifactPath == "" || source == "" {
+	if (provenancePath == "" || artifactPath == "") && artifactReference == "" {
+		fmt.Fprintf(os.Stderr, "either 'provenance' and 'artifact-path' or '' must be specified\n")
+		flag.Usage()
+		os.Exit(1)
+	}
+
```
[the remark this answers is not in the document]
asraa: Good catch, yes. I'll add a check for either -artifact or -provenance populated when artifact-reference is non-empty

cli/slsa-verifier/main_test.go
asraa: resolving, I think this snuck in through a bad merge or something

cli/slsa-verifier/main_test.go
asraa: These can be made in parallel: i'm not worried about rate limiting, most will be using offline bundle verification

cli/slsa-verifier/main_test.go
asraa: I can name this "generic_container", yes.

asraa: also, going to move the builder names to top-level variables, instead of in the funcs
```
var ARTIFACT_REFERENCE_BUILDERS = []string{"generic_container"}
var ARTIFACT_PATH_BUILDERS = []string{"go", "generic"}
```

container/container.go
asraa: Other verifiers may want it: If GCB is also signing containers, then I assume it will also want to run image verification, right?

verifiers/internal/gha/builder.go
asraa: Good call!

container/container.go
laurentsimon: but GCB won't be using cosign's format, at least we don't know that. Their first version wont use it for sure.
I see what you mean, though. This would be useful for any builder that integrates with Sigstore / Fulcio via OIDC, correct?

Makes sense, let's keep it here then.

cli/slsa-verifier/main.go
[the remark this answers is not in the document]
laurentsimon: up to you, I don't mind either way

verifiers/internal/gha/builder.go
asraa: curious: do you think this should be captured in BuilderOpts (internally generated depending on whether the artifact was a blob or container image)

verifiers/internal/gha/verifier.go
asraa: my rationale for putting container externally was that it may be used across verifiers. WDYT about putting it in verifiers/container/?

verifiers/internal/gha/verifier.go
asraa: hmmm the subject digest is verified against the computed artifact digest. is this what you mean? the artiact digest is resolved and put into the `ProvenanceOpts`

verifiers/internal/gha/builder.go
laurentsimon: i guess we could. But it's probably sufficient to select the correct "bucket" based on which function we're executing (`VerifyContainer` vs `VerifyArtifact`)?

verifiers/internal/gha/verifier.go
laurentsimon: you convinced me sharing is beneficial in another comment already :p. Under `verifiers/container/cosign.go` SGTM.

verifiers/internal/gha/verifier.go
laurentsimon: sorry. What I mean is that we don't verify the `subject.name`, but maybe we could since our generator puts the name of the container (including tags) in the name. That would provide a binding for container tags as well.

verifiers/internal/gha/builder.go
asraa: yeah, that makes sense.

cli/slsa-verifier/main.go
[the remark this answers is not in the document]
asraa: yes, it means repository:tag. technically it also supports the actual digest, so maybe I should say -artifact-image (and image reference resolution will be implicit)?

verifiers/internal/gha/verifier.go
asraa: Going to move over to image. I'm not totally sure I can test verification via image and implicit tag (latest), image and explicit tag (reference), or digest locally, but i'll see if we can in e2e tests.

cli/slsa-verifier/main.go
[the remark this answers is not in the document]
laurentsimon: so 'reference' means the image name + sha256?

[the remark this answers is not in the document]
asraa: Changed to artifact-image for now. Leaving unresolved for feedback from Ian

cli/slsa-verifier/main.go
asraa: Right now there's not difference in options, but there is a difference in artifact hash. Going to do this in the same follow-up PR as separating into subcommands. creating issue

verifiers/internal/gha/verifier.go
asraa: Interesting, gotcha. In case the tags are different, we need to compare the resolved digest of the containers... which imo comes down to verifying digest anyway...

We could compare repository name?

I'm a little worried because I have to mock this out for tests anyway because local filepaths supply the pathname (not container name) as input.

It is easy to do though, we can pipe it through `ProvenanceOpts` and verify when we verify subject digest.

laurentsimon: There seems to be a few things to discuss, so maybe let's create an issue to discuss further?

asraa: Done!
https://github.com/slsa-framework/slsa-verifier/issues/208

cli/slsa-verifier/main.go
```diff
@@ -38,7 +41,19 @@ func main() {
...
 
-	if provenancePath == "" || artifactPath == "" || source == "" {
+	if (provenancePath == "" || artifactPath == "") && artifactImage == "" {
+		fmt.Fprintf(os.Stderr, "either 'provenance' and 'artifact-path' or 'artifact-image' must be specified\n")
+		flag.Usage()
+		os.Exit(1)
+	}
+
+	if artifactImage != "" && (provenancePath != "" || artifactPath != "") {
+		fmt.Fprintf(os.Stderr, "'provenance' and 'artifact-path' should not be specified when 'artifact-image' is provided\n")
```
laurentsimon, line 51: note: that won't be true for GCB-built containers. Not something for this PR, though.

cli/slsa-verifier/main.go
```diff
@@ -64,17 +79,18 @@ func main() {
...
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "FAILED: SLSA verification failed: %v\n", err)
 		os.Exit(2)
 	}
 
 	fmt.Fprintf(os.Stderr, "PASSED: Verified SLSA provenance\n")
-
 	if printProvenance {
-		fmt.Fprintf(os.Stdout, "%s\n", string(verifiedProvenance))
+		for _, verified := range verifiedProvenance {
```
laurentsimon, line 91: This won't be a valid JSON output anymore, will it?
I wonder if we should simplify and print a single valid provenance. Later I agree we need t output all, but we need to decide the format. Wdut?

cli/slsa-verifier/main.go
```diff
@@ -88,39 +104,53 @@ func isFlagPassed(name string) bool {
...
-		artifactHash, provenanceOpts, builderOpts)
+	var provenance []byte
+	if provenancePath != "" {
+		provenance, err = os.ReadFile(provenancePath)
+		if err != nil {
+			return nil, "", err
+		}
+	}
+
+	return verifiers.Verify(ctx, artifactImage, provenance, artifactHash, provenanceOpts, builderOpts)
```
laurentsimon, line 138: do we want to have a `VerifyArtifact()` and `VerifyImage()`?

verifiers/internal/gha/verifier.go
```diff
@@ -80,11 +73,76 @@ func (v *GHAVerifier) VerifyArtifact(ctx context.Context,
...
+		cert, err := att.Cert()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "unexpected error getting certificate from OCI registry %s", err)
+			continue
+		}
+		verifiedProvenance, builderID, verifyErr = verifyEnvAndCert(env,
+			cert, provenanceOpts, builderOpts,
+			defaultContainerTrustedReusableWorkflows)
+		if verifyErr == nil {
+			return verifiedProvenance, builderID, nil
```
laurentsimon, line 143: does this mean we're always returning a maximum of 1 provenance? This means it's fine to print only a single provenance when passing `-print-provenance`.

laurentsimon (approved):
Thanks. Left a few comments.

cli/slsa-verifier/main.go
laurentsimon: Note: the code currently prints one character per line, I think

verifiers/container/container.go
```diff
@@ -0,0 +1,15 @@
+package container
+
+import (
+	"strings"
+
+	crname "github.com/google/go-containerregistry/pkg/name"
+)
+
+var GetImageDigest = func(image string) (string, error) {
```
laurentsimon, line 9: Note: would this break if we verify in parallel different images that require changing this global variable?
I don't think we're going to need that any time soon.. unless the unit tests have to.

cli/slsa-verifier/main.go
asraa: Interesting, OK. Maybe need to redefine the interface for `VerifyImage`

verifiers/container/container.go
asraa: Hmmm The purpose of this global is just for testing -- only testing will use the modified vars (and always will)

> I wonder whether we this should be a func GetImageDigest() and have an interface that this function implement in a structure. 

Actually, that can work. We can do a file-based impl and a registry-based implementation of the container. Then we can instantiate based on the params (was image str a local file or a container reference)

cli/slsa-verifier/main.go
asraa: Can do in the sub-command splitting issue!

asraa: Noted in the issue

cli/slsa-verifier/main.go
asraa: Oh no! Good point. This should have been one provenance each. Will simplify to a single valid provenance.

verifiers/internal/gha/verifier.go
asraa: Yes, correct. Updating. Will also create an issue to support multiple attestations returned on the container.

verifiers/container/container.go
laurentsimon: It's fine to do that in a follow-up PR. Up to you.

PR #159, opened 2022-07-21
debug: add error messages for debugging with rekor

asraa:

When seeing rekor entry errors, we only see `could not find a matching valid signature entry`. this augments the error messages with any unexpected errors.


laurentsimon (approved):
Thank you!

Issue #133, opened 2022-07-08
Add generic testsdata for its release


laurentsimon:
We need to populate https://github.com/slsa-framework/slsa-verifier/tree/main/testdata with tests data for the generic builder.
We typically do this as part of https://github.com/slsa-framework/slsa-github-generator/blob/main/RELEASE.md#update-verifier

It may make sense to create a folder for each builder under each `testsdata/vX.Y.Z`.

Some tests cases may be kinda general while others may be more builder-specific. Or maybe we just duplicate each of these for each builder, this way we're certain everything works.

@asraa Wdut? 

/cc @ianlewis

asraa:
> It may make sense to create a folder for each builder under each testsdata/vX.Y.Z.

+1

> Some tests cases may be kinda general while others may be more builder-specific.

Yes, right. For things like rekor errors we can keep them at the top-level for testdata/. Some things are general, e.g. checking builder pinning, but those can also be duped for each builder.

So the structure could be
```
testdata/
   - general tests
   go/
     v1.0.0/
     v1.1.1/
     general go tests
   generic/
      v1.0.0/
      general generic tests
```

laurentsimon:
LGTM. Let's also document the workflows (in example-package) to use to generate these... maybe they can all be in the same (that's a detail, I'm just curious :))?

ianlewis:
@asraa Do we still need to do anything for this issue? or can it be closed?

This issue is the last task referenced by https://github.com/slsa-framework/slsa-github-generator/issues/544 and I think we'd like to get that closed.

PR #162, opened 2022-07-22
test: add testing for generic builder multi subject

asraa:

Linking issue on multi subject verification https://github.com/slsa-framework/slsa-github-generator/issues/609

Part of https://github.com/slsa-framework/slsa-verifier/issues/133


main_test.go
```diff
@@ -362,7 +384,7 @@ func Test_runVerify(t *testing.T) {
 						t.Fatal(err)
 					}
 					for _, f := range builderDir {
-						if f.IsDir() {
+						if f.IsDir() && semver.Compare(minversion, f.Name()) <= 0 {
```
naveensrinivasan, line 387: A comment here would be helpful as to the reason for `semver.Compare`

main_test.go
```diff
@@ -343,17 +363,19 @@ func Test_runVerify(t *testing.T) {
...
-				}
-				for _, f := range testdataDir {
-					if f.IsDir() {
-						// These are the builder subfolders
-						builders = append(builders, f.Name())
+				builders := tt.builders
+				if len(builders) == 0 {
+					testdataDir, err := ioutil.ReadDir("./testdata")
+					if err != nil {
+						t.Fatal(err)
```
naveensrinivasan, line 372: A suggestion would be to have `t.Errorf` instead of `t.Fatal`

main_test.go
```diff
@@ -343,17 +363,19 @@ func Test_runVerify(t *testing.T) {
...
-				if err != nil {
-					t.Fatal(err)
-				}
-				for _, f := range testdataDir {
-					if f.IsDir() {
-						// These are the builder subfolders
-						builders = append(builders, f.Name())
+				builders := tt.builders
+				if len(builders) == 0 {
+					testdataDir, err := ioutil.ReadDir("./testdata")
```
naveensrinivasan, line 370: A suggestion would be to have `./testdata` as a constant.

The same  is being used here in next few lines below `ioutil.ReadDir(fmt.Sprintf("./testdata/%s", builder))`

main_test.go
asraa: I think that's fair, I didn't realize that `ReadDir` would return an empty list and I could still continue execution. Thanks!

laurentsimon (approved):
Awesome, thanks!

Issue #163, opened 2022-07-22
SLSA verifier as a service


laurentsimon:
To make the verifier accessible to everyone easily, we could have a REST API to verify as a service.

asraa:
It would also be beneficial for a service to have a configuration of options: e.g. if we have a stable service, we can more easily pin a trust root for the sigstore signing service (e.g. a hosted sigstore ecosystem).

Moreover, we can do things like enforce certain properties of builds with a service configuration rather than CLI options

laurentsimon:
yep I agree having option will be useful. I was thinking it could even be aligned with https://github.com/slsa-framework/slsa-verifier/issues/158?

Do you see use cases where the service and the CLI would significantly differ? I was thinking we could try to have them be more-or-less the same interface. That improves usability for a user who first tries the service, then wants to use the CLI and vice-versa. I would not say this is a hard constraint, though, but it'd be nice.

Was that the sort of configuration you had in mind; or something different?

ianlewis:
If we actually go through with creating a service, I wonder if it would make sense to define the service using grpc and expose it as a JSON service through [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway).

PR #164, opened 2022-07-22
feat: Create a verifier as a service

laurentsimon:
See https://github.com/slsa-framework/slsa-verifier/issues/163

This is an experimental feature, obviously. This PR does:
1. Re-factor code to be able to share the verification code between the cli and the service
2. Implement a simple REST API `/v1/verify` which takes as input the same as the CLI (except for the artifact which is replaced by its hash):
```yaml
{
    "source": "github.com/ossf/scorecard",
    "artifactHash": "02e7ac9a2a252a4e57cb2050953e787d095a04b955a41878068d60885bd31a0b",
    "tag": "v4.4.0",
     versionedTag: "bla",
    "printProvenance": true,
    "provenanceContent": "..."
}
```

Example: `curl -s 127.0.0.1:8000/v1/verify -d @./request`

and returns:
```yaml
{
   "error": "some caught error"
   "validation": "failure" / "success"
   "provenanceContent": "base64 intoto"
}
```

We may want to change some of the names. I have not included unit tests yet. Will do in a follow-up PR. (need to mock the http client).

Let me know what you think about the input / output format. I'm wondering if we should unify the output for both the cli and the service. It would force us to have proper error / log handling


cmd/verify.go
```diff
@@ -0,0 +1,52 @@
+package cmd
```
laurentsimon, line 1: this function was part of main.go, I just moved it in here. No change to the code itself

naveensrinivasan (commented):
This is large PR.

Can you please add some explanation as to what are these changes for? This will help others with the review process.

Can you please explain as to why are these test files part of this PR?

naveensrinivasan:
What is the idea behind the service? Is it for SLSA org to host this service? It is not clear

laurentsimon:
> This is large PR.

Only the 3 files under `service/` are new. The rest is renaming files.

> Can you please add some explanation as to what are these changes for? This will help others with the review process.

Please see https://github.com/slsa-framework/slsa-verifier/issues/163

> Can you please explain why are these test files part of this PR?

the folder has been renamed, there is no change to these files.

laurentsimon:
> What is the idea behind the service? Is it for SLSA org to host this service? It is not clear

Please see https://github.com/slsa-framework/slsa-verifier/issues/163

naveensrinivasan:
> > What is the idea behind the service? Is it for SLSA org to host this service? It is not clear
> 
> Please see #163

Thanks for editing the #163 with additional details. Also, a suggestion would be is to include a summary in the PR, which helps anyone new to understand the context before jumping into details about the issue.

.github/workflows/pre-submit.yml
```diff
@@ -20,7 +20,16 @@ jobs:
           set -euo pipefail
 
           go mod vendor
-          # Build
+          
+          # Build cli
+          cd cli
```
asraa, line 25: simplify to remove the `cd .../` with `go build -mod=vendor ./cli` and `go build -mod=vendor ./service`?

service/verify.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/slsa-framework/slsa-verifier/cmd"
+	"github.com/slsa-framework/slsa-verifier/verification"
+)
+
+var errInvalid = errors.New("invalid")
+
+type v1Query struct {
+	// Compulsory fields.
+	Source       string `json:"source"`
+	ArtifactHash string `json:"artifactHash"`
```
asraa, line 21: for containers, we may want to support this as a oneof artifacthash/dsse envelope or container image reference.

I would rather see grpc service like definitions, but save for a later review. just commenting on the idea for now.

service/verify.go
```diff
@@ -0,0 +1,170 @@
...
+
+	"github.com/slsa-framework/slsa-verifier/cmd"
+	"github.com/slsa-framework/slsa-verifier/verification"
+)
+
+var errInvalid = errors.New("invalid")
+
+type v1Query struct {
+	// Compulsory fields.
+	Source       string `json:"source"`
```
asraa, line 20: +1. I think one big mistake in cosign's API was that the source was never pinned. We should never try signature verification without a pinned subject.

service/verify.go
```diff
@@ -0,0 +1,170 @@
...
+
+var (
+	validationSuccess = validation("success")
+	validationFailure = validation("failure")
+)
+
+type v1Result struct {
+	Version         uint       `json:"version"`
+	Error           *string    `json:"error,omitempty"`
+	Validation      validation `json:"validation"`
```
asraa, line 40: Maybe an exit status int? easier than a string for version compatibility in case we extend this later.

asraa (commented):
Overall, I think it's useful to have a docker container service: it would be easy for users to run in their own clusters, re-package with a container image, etc. 

The limitation with signing the result is important too. Users should be able to specify a signer (e.g. we can use  in-memory signers for ephemeral test instances, or GCP signers that can rotate keys).

.github/workflows/pre-submit.yml
laurentsimon: I tried that but it failed `go: build output "cli" already exists and is a directory`... any way around that?

service/verify.go
laurentsimon: SG.

service/verify.go
laurentsimon: string can also be extended, right?
I'm not against using code, just trying to see the additional benefits it gives us.

I'm thinking we may want an `ErrorMsg`, `ErrorCode`... but I'm not sure if Golang handles that easily in practice...unless we wrap all errors with our own structure. Wdut?

laurentsimon:
> The limitation with signing the result is important too. Users should be able to specify a signer (e.g. we can use in-memory signers for ephemeral test instances, or GCP signers that can rotate keys).

+1, we need to take care of this, in particular for a public service.

laurentsimon:
> Overall, I think it's useful to have a docker container service: it would be easy for users to run in their own clusters, re-package with a container image, etc.
> 
> The limitation with signing the result is important too. Users should be able to specify a signer (e.g. we can use in-memory signers for ephemeral test instances, or GCP signers that can rotate keys).

if you have pointers for this, please drop them in. I'd like to take a look, even if it's not part of this PR. Thanks

.github/workflows/pre-submit.yml
asraa: `-o verifier`? it's weird i can't find any docs on that... i'm pretty sure the behavior for `./cmd/XYZ` is that it writes to XYZ. But maybe not.

laurentsimon: `-o` did the trick. Thanks

laurentsimon:
> Currently, the CLI tool will be go install'd as `cli`. I think you should move it to `./cmd/slsa-verifier` to make it be named as `slsa-verifier`. Service can move under `/experimental/service`. Should the current `./cmd/verify.go` be moved into `verification/` as another module?

that should work. I'll ping once I've made the changes. Thanks

fyi @ianlewis

laurentsimon:
> > Currently, the CLI tool will be go install'd as `cli`. I think you should move it to `./cmd/slsa-verifier` to make it be named as `slsa-verifier`. Service can move under `/experimental/service`. Should the current `./cmd/verify.go` be moved into `verification/` as another module?
> 
> that should work. I'll ping once I've made the changes. Thanks
> 
> fyi @ianlewis

changes made. I've kept `cli` for the code that defines the cli. I've mades all the other changes. PTAL

PR #170, opened 2022-07-25
feat: fix renovatebot tags

laurentsimon:
Unless the tags are a full semver, renovatebot seems unable to display the version in PR description.


asraa (approved):

laurentsimon:
rekor is upset at us today :-) presubmits failing. WOUld it help if we ran these test *not* in parallel?

asraa:
> rekor is upset at us today :-) presubmits failing. Would it help if we ran these test not in parallel (if this is a rate limiting issue)?

Oops! I think I just missed the commenting out at the top of the test invocation. Yes definitely. Next PR I'll add in retries

PR #182, opened 2022-08-03
feat: Create a verifier as a service

laurentsimon:
See https://github.com/slsa-framework/slsa-verifier/issues/163

This PR was originally started in #164 . There was a large file accidentally pushed in one of the commits. After removing it, the PR got automatically closed. Hence this new PR.

This is an experimental feature, obviously. This PR does:
1. Re-factor code to be able to share the verification code between the cli and the service
2. Implement a simple REST API `/v1/verify` which takes as input the same as the CLI (except for the artifact which is replaced by its hash):
```yaml
{
    "source": "github.com/ossf/scorecard",
    "artifactHash": "02e7ac9a2a252a4e57cb2050953e787d095a04b955a41878068d60885bd31a0b",
    "tag": "v4.4.0",
     versionedTag: "bla",
    "printProvenance": true,
    "provenanceContent": "..."
}
```

Example: `curl -s 127.0.0.1:8000/v1/verify -d @./request`

and returns:
```yaml
{
   "error": "some caught error"
   "validation": "failure" / "success"
   "provenanceContent": "base64 intoto"
}
```

We may want to change some of the names. I have not included unit tests yet. Will do in a follow-up PR. (need to mock the http client).

Let me know what you think about the input / output format. I'm wondering if we should unify the output for both the cli and the service. It would force us to have proper error / log handling


README.md
```diff
@@ -23,20 +23,20 @@ You have two options to install the verifier.
 
 #### Option 1: Install via go
 ```
-$ go install github.com/slsa-framework/slsa-verifier@v1.2.0
+$ go install github.com/slsa-framework/slsa-verifier@v1.2.1
```
asraa, line 26: 1.2.1 doesn't exist yet, maybe it does for the builder?

cli/experimental/service/README.md
```diff
@@ -0,0 +1,15 @@
+# SLSA verifier as a service
+
+This document is WIP.
+
+Command to run the service locally:
+
+```bash
+$ docker build -f service/Dockerfile .
```
asraa, line 8: Should this README be under `experimental` dir given the path to the dockerfile?

asraa: I think it needs to be run from slsa-verifier head, with path `cli/experimental/service/Dockerfile`, otherwise go mod vendor won't work. Then the `cd service` and the `COPY` in the Dockerfile also needs to be changed.

cli/experimental/service/README.md
```diff
@@ -0,0 +1,15 @@
...
+$ docker run -p 8000:8000 slsa-verifier-rest:latest # This did not work for me.
+$ docker run --network=host slsa-verifier-rest:latest
+```
+
+```bash
+$ curl -s 127.0.0.1:8000/v1/verify -d @./request.txt
```
asraa, line 14: move request.txt to a testdata folder?

asraa (approved):
lgtm to keep iterating on

README.md
laurentsimon: my bad. Fixed.

cli/experimental/service/README.md
laurentsimon: i Have not renamed the folder in the example, thanks for catching. I take it means the folder re-arrangement is OK..

cli/experimental/service/README.md
laurentsimon: done.

experimental/rest/service.go
```diff
@@ -0,0 +1,169 @@
...
+type v1Query struct {
+	// Compulsory fields.
+	Source       string `json:"source"`
+	ArtifactHash string `json:"artifactHash"`
+	DsseEnvelope string `json:"provenanceContent"`
+	// Optional fields.
+	Tag             *string `json:"tag"`
+	Branch          *string `json:"branch"`
+	VersionedTag    *string `json:"versionedTag"`
+	PrintProvenance *bool   `json:"printProvenance"`
```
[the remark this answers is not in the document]
laurentsimon: Yes,, someone may want to check for repository-id and what not. It's an optional argument, so is not needed. Is that OK?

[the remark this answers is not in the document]
laurentsimon: We will iterate, so I'm going to merge. This PR is just to get the ball rolling with something to hack on.

PR #184, opened 2022-08-03
feat: Add `-builder-id` option to CLI

laurentsimon:
This PR:
- Add `-builder-id` as an optional argument. This will be useful for users to select a builder, especially when we add GCB support. For GitHub builders, we verify that the repo is a github.com repository.
- Verify the builderID in both the certificate and the provenance. We omit the tag during verification. I think we can add support that later if people ask. I imagine we'd check whether the user provided a tag, and if so, we'd verify it too. Otherwise, only the part before the tag.
- Add a BuilderOpt
- Add verification of material source URI and configSource URI, and that they match (hardening)


laurentsimon:
note to myself: need to add tests for runVerify() in main_test.go.

main.go
```diff
@@ -66,6 +72,7 @@ func verify(ctx context.Context,
 }
 
 func main() {
+	flag.StringVar(&builderID, "builder-id", "", "EXPERIMENTAL: the unique builder ID")
```
asraa, line 75: nit: "unique builder ID specified in the provenance"? 

is this not a provenance options?

main.go
```diff
@@ -48,13 +52,15 @@ func verify(ctx context.Context,
 	}
 
 	// Verify the workflow identity.
-	if err := pkg.VerifyWorkflowIdentity(workflowInfo, source); err != nil {
+	builderID, err := pkg.VerifyWorkflowIdentity(workflowInfo, builderOpts,
```
asraa, line 55: i think we will need to revisit this eventually and generalize this to "VerifySignerIdentity" so that it can both check github workflow signed certificates against the source/builder, but also support general other builders that sign with other types of certifiates and keys. wdyt?

pkg/builder.go
```diff
@@ -15,44 +15,69 @@ var (
...
 	if len(workflowPath) < 2 {
-		return fmt.Errorf("%w: %s", errorMalformedWorkflowURI, id.JobWobWorkflowRef)
+		return "", fmt.Errorf("%w: workflow uri: %s", ErrorMalformedURI, id.JobWobWorkflowRef)
 	}
 
 	// Trusted workflow verification by name.
-	reusableWorkflowName := strings.Trim(workflowPath[0], "/")
-	if _, ok := trustedReusableWorkflows[reusableWorkflowName]; !ok {
-		return fmt.Errorf("%w: %s", ErrorUntrustedReusableWorkflow, reusableWorkflowName)
+	// The caller repository in the x509 extension is not fully qualified. It only contains
```
asraa, line 34: this comment does not apply. the reusable workflow path is fully qualified.

pkg/provenance.go
```diff
@@ -86,32 +169,44 @@ func VerifyProvenanceSignature(ctx context.Context, rClient *client.Rekor, prove
 	return env, cert, nil
 }
 
-func VerifyProvenance(env *dsselib.Envelope, opts *ProvenanceOpts) error {
+func VerifyProvenance(env *dsselib.Envelope, builderID string,
```
asraa, line 172: nice, this is a great improvement

asraa (commented):
thanks! ultimately this is to support other reusable workflows that the slsa repo does not distribute, right? 

we are still limited to fulcio issued signing certificates and builder identities on GitHub

will we eventually move to other identity providers? e.g. cloud build using a SA to authenticate to fulcio? is that what will allow gcb support?

main.go
laurentsimon: it is in both the certificate and the provenance for our builders. I can move it to provenanceOpts. Wdut?

laurentsimon:
> thanks! ultimately this is to support other reusable workflows that the slsa repo does not distribute, right?

the first use case it to let the user select the builder they trust or the one they expect: GCB vs GHA, for example. Later we can think about extending to builders we don't support. The problem is that they all have their own implementation, so I don't know if we'll be able to support other builders (verification of tags, branch, etc). If they are on GitHub using re-usable workflow, we can. So I'd hope we can, but it's still unclear to me. Maybe we'll just do signature verification...?

> we are still limited to fulcio issued signing certificates and builder identities on GitHub

For this PR yes. Next PR ill add support for GCB and ew'll have a different builder altogether.

> will we eventually move to other identity providers? e.g. cloud build using a SA to authenticate to fulcio? is that what will allow gcb support?

GCB just has hard coded keys for their builders. So this is what I plan to support first. Are there people using SA to connect to Fulcio at the moment?

pkg/builder.go
laurentsimon: you're right. Copy/paste abuse!

main.go
laurentsimon: Agreed. The function is doing 2 things at once (builder and source). I did not want to change it in this PR. Ill try to do tat with the GCB support PR

main.go
laurentsimon: Changed text to `the unique builder ID who created the provenance`.
Let me know if I should get rid f the builderOpt and use the provenanceOpt instead. You mentioned buildOpt offline, so I took this as a hint ^^. Can change it

asraa: Ah, after reading the code I think we should keep in both. Can we add a BuilderID in the provenance opts and use this flag to populate both the field in the builderopts and the provenanceopts? 

that way `VerifyProvenance` only takes provenance opts. Same with `VerifyWorkflowIdentity`: this should only consume from the builder options I think. do you agree?

laurentsimon: That's fine with me. The `VerifyWorkflowIdentity` still takes a source as well, but we can fix it in a follow-up PR.

asraa: awesome, plan could be that we do a ConstructBuilderOptsFor(GHA/GCB), ConstructProvenanceOptsFor(GHA/GCB). Follow-up sg.

cli/slsa-verifier/main.go
```diff
@@ -41,21 +43,25 @@ func main() {
...
 
 	if isFlagPassed("tag") {
 		ptag = &tag
 	}
 	if isFlagPassed("versioned-tag") {
 		pversiontag = &versiontag
 	}
+	if isFlagPassed("builder-id") {
+		pbuilderID = &builderID
+	}
```
ianlewis, line 56: aside: I'm curious. Remind me again why these flags are copied to local variables? Also, why we need a special `isFlagPassed` function?

experimental/rest/service.go
```diff
@@ -165,5 +171,7 @@ func (q *v1Query) validate() error {
 		return fmt.Errorf("%w: tag and versionedTag are mutually exclusive", errInvalid)
 	}
 
+	// BuilderID is optional, so not additional validation needed.
```
ianlewis, line 174:
```suggestion
	// BuilderID is optional, so no additional validation needed.
```

verification/builder.go
```diff
@@ -15,44 +15,67 @@ var (
...
+var defaultTrustedReusableWorkflows = map[string]bool{
 	trustedBuilderRepository + "/.github/workflows/generator_generic_slsa3.yml": true,
 	trustedBuilderRepository + "/.github/workflows/builder_go_slsa3.yml":        true,
 }
 
 // VerifyWorkflowIdentity verifies the signing certificate information
-func VerifyWorkflowIdentity(id *WorkflowIdentity, source string) error {
+func VerifyWorkflowIdentity(id *WorkflowIdentity,
+	builderOpts *BuilderOpts, source string,
+) (string, error) {
```
ianlewis, line 26: nit: This feels inconsistent with style elsewhere, make this one line.

verification/builder.go
```diff
@@ -15,44 +15,67 @@ var (
...
 	}
 
 	// Verify the ref.
 	if err := verifyTrustedBuilderRef(id, strings.Trim(workflowPath[1], "/")); err != nil {
-		return err
+		return "", err
 	}
 
 	// Issuer verification.
 	if !strings.EqualFold(id.Issuer, certOidcIssuer) {
```
ianlewis, line 46: aside: Also curious about this. Is unicode case folding the way we should really be comparing issuers?

verification/builder_test.go
```diff
@@ -166,19 +226,118 @@ func Test_VerifyWorkflowIdentity(t *testing.T) {
...
 	for _, tt := range tests {
 		tt := tt // Re-initializing variable so it is not changed while executing the closure below
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
-			err := VerifyWorkflowIdentity(tt.workflow, tt.source)
+			opts := tt.buildOpts
+			if opts == nil {
+				opts = &BuilderOpts{}
+			}
+			_, err := VerifyWorkflowIdentity(tt.workflow, opts, tt.source)
```
ianlewis, line 267: nit: Consider allowing `VerifyWorkflowIdentity` to take a nil `BuilderOpts`

verification/provenance.go
```diff
@@ -34,6 +34,89 @@ func provenanceFromEnv(env *dsselib.Envelope) (prov *intoto.ProvenanceStatement,
...
+		return fmt.Errorf("%w: expected source github.com repository '%s'", ErrorMalformedURI,
+			source)
+	}
+
+	// Verify source from ConfigSource field.
+	configURI, err := sourceFromURI(prov.Predicate.Invocation.ConfigSource.URI)
+	if err != nil {
+		return err
+	}
+	if !strings.EqualFold(configURI, source) {
```
ianlewis, line 78: Again, I suspect that unicode folding isn't how we likely want to do URI comparisons.

laurentsimon: Good catch. I'll change that. Thank you!

cli/slsa-verifier/main.go
laurentsimon: I don't think I found a way to have an option be a pointer to a string, and for the pointer to be nil if the user does not provide the option. If that's possible, I can update the code.

verification/builder.go
laurentsimon: great point. I'll update

verification/builder_test.go
laurentsimon: I'll keep this way for now since each field will be a pointer. I'll keep this in mind for future iteration. Let me know if that's OK.

laurentsimon:
This pR is superseded by https://github.com/slsa-framework/slsa-verifier/pull/184. I've addressed the comments in the other PR.

verification/builder_test.go
ianlewis: Yeah, it's minor but I just mean basically putting the if above inside the function and allow callers to do `VerifyWorkflowIdentity(workflow, nil, source)`

ianlewis:
I assume you mean #187

laurentsimon:
> I assume you mean #187

correct, sorry. Ive updated the comment above.

verification/builder_test.go
laurentsimon: I'll do that in the #187 PR, thanks

PR #187, opened 2022-08-04
feat: Create interface for verifiers

laurentsimon:
This PR supersedes https://github.com/slsa-framework/slsa-verifier/pull/184.
This PR attempts to create a better framework to support new builders. It starts simple:

- Create an interface that each verifier will instantiate. The interface is defined in `register/register.go`:
```go
type SLSAVerifier interface {
	// Match matches a BuilderID.
	Match(builderID string) bool

	// Verify verifies a provenance.
	Verify(ctx context.Context,
		provenance []byte, artifactHash string,
	) ([]byte, string, error)
}
```
- Each verifier registers its instantiation via `register.RegisterVerifier()`: see `verifiers/internal/gcb/verifier.go` and `verifiers/internal/gha/verifier.go`.
- If the user provides a builder in the CLI, we search for a verifier that supports the builderID using `v.Match(builderID)`. See `verifiers/verifier.go`
- If no builderID is provided by the user, we default to the GHA native builders, see `verifiers/verifier.go`. I'm doing this to avoid breaking previous version, but we can change this behavior later. We could also check whether builder supports particular repositories: GCB supports GitLab, whereas GHA builders do not.
- Add verification of material source URI and configSource URI, and that they match (hardening)

I also moved errors to their own `errors` folder.

Everything else is moving files around and renaming calls to account for files that have been moved around.


asraa:
First question: 
I can base https://github.com/slsa-framework/slsa-verifier/pull/147

on top of this, but the problem is that blob verification and container verification have different APIs. Container verification takes an image reference, whereas blobs require the provenance bytes.

Either we have some sort of struct with type information capturing the object to verify, or we have two possibilities for each builder `VerifyBlob` and `VerifyOCI`. @ianlewis

laurentsimon:
We can create a different API, as you say: VerifyImage(), VerifyArtifact() for the interface. 
Would that work?

asraa:
> We can create a different API, as you say: VerifyImage(), VerifyArtifact() for the interface. Would that work?

That SGTM

In addition, can we expose a SLSA Level `SLSALevel()` on each builder? GCB builders are v2, and GHA reusble workflows are v3, we should support exposing the information.

verifiers/internal/gcb/verifier.go
```diff
@@ -0,0 +1,37 @@
...
+func (v *GCBVerifier) Match(builderID string) bool {
+	// This verifier only supports the GCB builders.
+	return strings.HasPrefix(builderID, "https://cloudbuild.googleapis.com/GoogleHostedWorker@")
+}
+
+// Verify provenance.
+func (v *GCBVerifier) Verify(ctx context.Context,
+	provenance []byte, artifactHash string,
+	provenanceOpts *options.ProvenanceOpts,
+	builderOpts *options.BuilderOpts,
```
asraa, line 34: I believe GCB v2 provenance is signed with a configured key in the workers.

Would the option for specifying public keys exist in `BuilderOpts`?

laurentsimon: honest answer is TBD :) I agree we need to think of supporting keys. I'm always worried about that because it's dangerous. So this s TBD. I think it's do-able thru a new `BuilderOpts` I you say.

laurentsimon:
> > We can create a different API, as you say: VerifyImage(), VerifyArtifact() for the interface. Would that work?
> 
> That SGTM
> 
> In addition, can we expose a SLSA Level `SLSALevel()` on each builder? GCB builders are v2, and GHA reusable workflows are v3, we should support exposing the information.

Yes we can. Level 3 is a little subjective, so I was hesitant to do that. But maybe that's a good way to get the ball rolling.

cli/slsa-verifier/main.go
```diff
@@ -41,21 +44,25 @@ func main() {
 		os.Exit(1)
 	}
 
-	var ptag, pversiontag *string
+	var pbuilderID, ptag, pversiontag *string
 
 	if isFlagPassed("tag") {
 		ptag = &tag
 	}
 	if isFlagPassed("versioned-tag") {
```
laurentsimon, line 52: > comment from @ianlewis: 

aside: I'm curious. Remind me again why these flags are copied to local variables? Also, why we need a special isFlagPassed function?

> reply from @laurentsimon 
I don't think I found a way to have an option be a pointer to a string, and for the pointer to be nil if the user does not provide the option. If that's possible, I can update the code.

verifiers/verifier.go
```diff
@@ -0,0 +1,36 @@
...
+	"fmt"
+
+	serrors "github.com/slsa-framework/slsa-verifier/errors"
+	"github.com/slsa-framework/slsa-verifier/options"
+	"github.com/slsa-framework/slsa-verifier/register"
+	_ "github.com/slsa-framework/slsa-verifier/verifiers/internal/gcb"
+	"github.com/slsa-framework/slsa-verifier/verifiers/internal/gha"
+)
+
+func Verify(ctx context.Context,
```
laurentsimon, line 14: TODO: support VerifyWorkflowIdentity(workflow, nil, source)
/cc @ianlewis

cli/slsa-verifier/main.go
asraa: Yeah, I believe `StringVar` is as close as you can get, and the best that can be done is use a default empty string. but problem is that empty string doesn't necessarily mean no value.

In this case we can document that? empty tags passed in mean no tag verification.

register/register.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+var SLSAVerifiers = make(map[string]SLSAVerifier)
+
+type SLSAVerifier interface {
+	// Match matches a BuilderID.
+	Match(builderID string) bool
+
+	// Verify verifies a provenance.
+	Verify(ctx context.Context,
```
asraa, line 16: Should we qualify this with `VerifyArtifact` and comment `Verify verifies a provenance for a supplied artifact` 

(vs. `VerifyImage` that verifies a container image with provenance stored in the registry?)

register/register.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"context"
+
+	"github.com/slsa-framework/slsa-verifier/options"
+)
+
+var SLSAVerifiers = make(map[string]SLSAVerifier)
+
+type SLSAVerifier interface {
+	// Match matches a BuilderID.
```
asraa, line 12: against the expected builder ID for the verifier

verifiers/internal/gha/builder.go
```diff
@@ -15,44 +18,66 @@ var (
...
 
-var trustedReusableWorkflows = map[string]bool{
+var defaultTrustedReusableWorkflows = map[string]bool{
 	trustedBuilderRepository + "/.github/workflows/generator_generic_slsa3.yml": true,
 	trustedBuilderRepository + "/.github/workflows/builder_go_slsa3.yml":        true,
 }
 
 // VerifyWorkflowIdentity verifies the signing certificate information
-func VerifyWorkflowIdentity(id *WorkflowIdentity, source string) error {
+func VerifyWorkflowIdentity(id *WorkflowIdentity, builderOpts *options.BuilderOpts, source string) (string, error) {
```
asraa, line 27: TBD but we may want to move this under the Verifier type and have the signature take in the PublicKeyOrCertificate.

In the case of GCB/other keys, they will have different "identity" guarantees in the public key or cert.

Fine for now since this is just a refactor. It won't break more API to add a new method to the Verifier type

asraa (approved):

register/register.go
laurentsimon: sure. I left it as-is because we can change in your upcoming PR. But I'll do it here anyway. Thanks

register/register.go
laurentsimon: +1

verifiers/internal/gha/builder.go
laurentsimon: Let me try to make a small change move this function under builderOpt.

laurentsimon: mhh, will leave as-is for now since `VerifyWorkflowIdentity()` is called within the main entry point `VerifyArtifact()`, so there's no need to expose it directly in this PR. Can refactor more in follow-up PRs.

cli/slsa-verifier/main.go
laurentsimon: done. Added a comment.

Issue #180, opened 2022-08-02
Support for more SLSA default fields


laurentsimon:
Fields like `subject-name`, `source-sha1` may be required.
- subject-name is often used for container image
- source-sha1 is useful for commit hash on a repo, but also if source is a tarball / zip file, e.g. GCB supports building from `gs://bucket-name/file`

@asraa @ianlewis @joshuagl wdut?

ianlewis:
Do you mean as flags to the verifier? assuming that's right,

I think that #92 would need at least some way to provide the container image name so maybe `subject-name` would be more generic and could be supported with all artifact types.

`source-sha1` could probably be something like `--source=sha1:deadbeef` to allow for verifying other kinds of hash algorithms. That would make it a bit more future-proof.

laurentsimon:
That'd work yes. Does it mean we should have `--source=sha1:bla` and `--source-repository=bla` or shall we use the same argument for both?

I think we'll also need to support source being a path to a file. For example, if someone wants to verify from a local tarball..

Wdut?

ianlewis:
Maybe `--source-uri` that matches the uri listed in the `materials`?

laurentsimon:
Great idea. So `source-uri` and `source-digest` would be the 2 options we could use. That'd be a breaking change, but I think it's fine.

@asraa further thoughts?

laurentsimon:
Since we're at it, shall we also rename `-provenance` to `provenance-path`?

asraa:
Can we make a mega-issue describing a final CLI? There's this issue changing flags, comments from #147, and we will also have `--builder-id`

laurentsimon:
+1, works for me.

laurentsimon:
`-commit=sha1:abcde` is also required, unless we ask users to add the commit in `-source-uri=repo@sha1:bla`

laurentsimon:
Another flag is `-container-tag`, and we'd rename `-tag` to `-source-tag`.

PR #202, opened 2022-08-09
feat: Support for GCB verification

laurentsimon:
This PR depends on https://github.com/slsa-framework/slsa-verifier/pull/147 which should be merged.

Please verify the GCB public keys in this PR, using steps in verifiers/internal/gcb/keys/README.md.

This PR adds support for verification of GCB provenance. Some TODOS:
- [x] Unit tests. Follow-up PRs will be needed
- [x] This only works for containers today, but does not support provenance available next to the registry. (current limitation of GCB).

In a follow-up PR: https://github.com/slsa-framework/slsa-verifier/issues/221


verifiers/internal/gcb/provenance.go
```diff
@@ -0,0 +1,281 @@
...
+func (self *GCBProvenance) VerifyVersionedTag(tag string) error {
+	return fmt.Errorf("%w: GCB versioned-tag verification", serrors.ErrorNotSupported)
+}
+
+func (self *GCBProvenance) VerifySignature() error {
+	if len(self.gcloudProv.ProvenanceSummary.Provenance) == 0 {
+		return fmt.Errorf("%w: no provenance found", serrors.ErrorInvalidDssePayload)
+	}
+	// Assume a single provenance in the array.
+	prov := self.gcloudProv.ProvenanceSummary.Provenance[0]
```
laurentsimon, line 227: TODO: need to iterate. This will support multiple version specs in the future (v0.1, v0.2, v1.0, etc).

verifiers/internal/gcb/provenance.go
```diff
@@ -0,0 +1,281 @@
...
+
+	// Verify the envelope type. It shoudl be an intoto type.
```
laurentsimon, line 229: typo

verifiers/internal/gcb/keys/keys.go
```diff
@@ -0,0 +1,70 @@
...
+	// TODO: key type and size
+}
+
+func PublicKeyNew(region string) (*PublicKey, error) {
+	content, err := fs.ReadFile(publicKeys, path.Join("materials", region+".key"))
+	if err != nil {
+		return nil, fmt.Errorf("%w: cannot read key materials", err)
+	}
+
+	return &PublicKey{
```
asraa, line 30: In case someone downloads them themselves, I think the pem.Decode check for valid key should go here before sig validation

verifiers/internal/gcb/keys/keys.go
```diff
@@ -0,0 +1,70 @@
...
+		Value:  content,
+		Region: region,
+	}, nil
+}
+
+func (self *PublicKey) VerifySignature(digest [32]byte, sig []byte) error {
+	block, _ := pem.Decode(self.Value)
```
asraa, line 37: is there always exactly one PK per region?

verifiers/internal/gcb/provenance.go
```diff
@@ -0,0 +1,308 @@
...
+
+func payloadFromEnvelope(env *dsselib.Envelope) ([]byte, error) {
+	payload, err := base64.StdEncoding.DecodeString(env.Payload)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %s", serrors.ErrorInvalidDssePayload, err.Error())
+	}
+	return payload, nil
+}
+
+func (self *GCBProvenance) isVerified() error {
```
asraa, line 86: recommend changing this to a boolean rather than an error

verifiers/internal/gcb/provenance.go
```diff
@@ -0,0 +1,308 @@
...
+		return "", fmt.Errorf("%w: '%s' field is absent", serrors.ErrorInvalidDssePayload, key)
+	}
+	ts, ok := t.(string)
+	if !ok {
+		return "", fmt.Errorf("%w: '%s' is not a string", serrors.ErrorInvalidDssePayload, key)
+	}
+	return ts, nil
+}
+
+func (self *GCBProvenance) VerifySubjectDigest(expectedHash string) error {
```
asraa, line 193: The wrapped gcpProvenance headers are unverified, should we leave them that way, or verify against these fields against the imagesummary as well?

verifiers/internal/gcb/provenance.go
```diff
@@ -0,0 +1,308 @@
...
+
+	var errs []error
+	regex := regexp.MustCompile(`^projects\/verified-builder\/locations\/(.*)\/keyRings\/attestor\/cryptoKeys\/builtByGCB\/cryptoKeyVersions\/1$`)
+
+	for _, sig := range prov.Envelope.Signatures {
+		match := regex.FindStringSubmatch(sig.KeyID)
+		if len(match) == 2 {
+			// Create a public key instance for this region.
+			region := match[1]
+			pubKey, err := keys.PublicKeyNew(region)
```
asraa, line 277: maybe this should return a dsse verifier? that way you don't have to implement the payload hashing or signature decoding.

verifiers/internal/gcb/verifier.go
```diff
@@ -40,9 +41,70 @@ func (v *GCBVerifier) VerifyArtifact(ctx context.Context,
 
 // VerifyImage verifies provenance for an OCI image.
 func (v *GCBVerifier) VerifyImage(ctx context.Context,
-	artifactImage string,
+	provenance []byte, artifactImage string,
 	provenanceOpts *options.ProvenanceOpts,
 	builderOpts *options.BuilderOpts,
 ) ([]byte, string, error) {
-	return nil, "todo", serrors.ErrorNotSupported
+	prov, err := ProvenanceFromBytes(provenance)
```
asraa, line 48: will this fail if provenance is nil? would be good to test -- since generic containers don't need this

verifiers/internal/gcb/verifier.go
```diff
@@ -40,9 +41,70 @@ func (v *GCBVerifier) VerifyArtifact(ctx context.Context,
...
+	// Verify branch.
+	if provenanceOpts.ExpectedBranch != nil {
+		if err = prov.VerifyBranch(*provenanceOpts.ExpectedBranch); err != nil {
+			return nil, "", err
+		}
+	}
+
+	// Verify the tag.
+	if provenanceOpts.ExpectedTag != nil {
+		if err := prov.VerifyTag(*provenanceOpts.ExpectedTag); err != nil {
```
asraa, line 88: there's likely no way to do this right? the only way would be an online tag comparison to the commit in the URI i think

verifiers/internal/gcb/verifier.go
```diff
@@ -40,9 +41,70 @@ func (v *GCBVerifier) VerifyArtifact(ctx context.Context,
...
+	}
+
+	// Verify the versioned tag.
+	if provenanceOpts.ExpectedVersionedTag != nil {
+		if err := prov.VerifyVersionedTag(*provenanceOpts.ExpectedVersionedTag); err != nil {
+			return nil, "", err
+		}
+	}
+
+	// TODO: verify summary information:
```
asraa, line 100: ah you have imagesummary verification here :)

verifiers/internal/gcb/testdata/gcloud-container-invalid-slsaheader.json
```diff
@@ -0,0 +1,94 @@
...
+            "intotoStatement": {
+              "_type": "https://in-toto.io/Statement/v0.1",
+              "predicateType": "https://slsa.dev/provenance/v0.1",
+              "slsaProvenance": {
+                "builder": {
+                  "id": "https://cloudbuild.googleapis.com/GoogleHostedWorker@v0.2"
+                },
+                "materials": [
+                  {
+                    "uri": "https://github.com/laurentsimon/gcb-tests/commit/fbbb98765e85ad464302dc5977968104d36e455e"
```
asraa, line 21: will this always be a github source? I guess not. That is a little annoying: should we open our `--source` to handle other source material?

verifiers/internal/gcb/provenance.go
```diff
@@ -0,0 +1,308 @@
...
+			ResourceUri string           `json:"resourceUri"`
+			Envelope    dsselib.Envelope `json:"envelope"`
+		} `json:"provenance"`
+	} `json:"provenance_summary"`
+}
+
+type GCBProvenance struct {
+	gcloudProv                    *gloudProvenance
+	verifiedIntotoStatementStruct *v01IntotoStatement
+	// Note: may need to support envelope only.
```
asraa, line 51: supporting just the predicate too might make it clear that this is already verified

asraa (commented):
Thanks for starting this! Some comments. Overall, implementation of the provenance looks fine to me.

verifiers/internal/gcb/keys/keys.go
laurentsimon: I can move the Pem decoding here. What do you mean by "In case someone downloads them themselves"?

verifiers/internal/gcb/keys/keys.go
laurentsimon: yes. This may change if they rotate keys, but there is currently no plan to do that.

verifiers/internal/gcb/keys/keys.go
```diff
@@ -0,0 +1,70 @@
...
+
+	if !ecdsa.VerifyASN1(pubKey, digest[:], sig) {
+		return fmt.Errorf("%w: cannot verify with public key '%v'",
+			serrors.ErrorInvalidSignature, string(self.Region))
+	}
+
+	return nil
+}
+
+// TODO: load public keys
```
laurentsimon, line 60: TODO: remove this comment

verifiers/internal/gcb/keys/list
```diff
@@ -0,0 +1,37 @@
+NAME                     CPUS  DISKS_GB  ADDRESSES  RESERVED_ADDRESSES  STATUS  TURNDOWN_DATE
```
laurentsimon, line 1: TODO: delete this file

verifiers/internal/gcb/provenance.go
laurentsimon: you're suggesting not adding the envelope, correct?
(I think the note is outdated)

verifiers/internal/gcb/provenance.go
```diff
@@ -0,0 +1,308 @@
...
+	// TODO: verified flag?
+}
+
+// go run cli/slsa-verifier/main.go -artifact-path python/secure_package_template-0.2.0-py3-none-any.whl -provenance verifiers/internal/gcb/testdata/test.intoto.jsonl -source github.com/sethmlarson/python-slsa-release-test -builder-id=https://cloudbuild.googleapis.com/GoogleHostedWorker@0.2
```
laurentsimon, line 55: TODO: remove comment

verifiers/internal/gcb/provenance.go
laurentsimon: Can do that. Having the `serrors.ErrorNoValidSignature` in one place was easier than having the `if bool return serrors.ErrorNoValidSignature`

Does this change anything?

verifiers/internal/gcb/provenance.go
laurentsimon: I would like to verify them. Either in this PR or a follow-up PR. Thanks for the reminder.
I've been hesitant to do it till now because I don't know what the integration with gcloud will look like. As soon as I get more insight into this I'll add it.
Created https://github.com/slsa-framework/slsa-verifier/issues/221 for tracking

verifiers/internal/gcb/provenance.go
laurentsimon: I had to replace 2 characters before base64 decoding: Is this standard in dsse? It did not seem to be https://github.com/secure-systems-lab/dsse/blob/master/envelope.md

verifiers/internal/gcb/verifier.go
laurentsimon: generic containers cannot be verified via the GCB verifier. So I think this is fine until GCB supports provenance on the registry. Confirmed?

verifiers/internal/gcb/verifier.go
laurentsimon: You're right. This will just report an error "unsupported" if the user passes this option. This way it's not silently ignored.
Wdut?

verifiers/internal/gcb/verifier.go
laurentsimon: yes. Plus the human-readable part if possible, because that's the part users may see by opening the file.

verifiers/internal/gcb/testdata/gcloud-container-invalid-slsaheader.json
laurentsimon: It won't always be a GitHub source. `-source` is not restricted to github, is it?

verifiers/internal/gcb/verifier.go
laurentsimon: Actually there is a way to do that: GCB provenance contains the tag in the subject. I have not implemented it yet.

laurentsimon: I take back what I say. There is confusion between repository tag and image tag. Container tag is do-able today, but I think it should just be a `-subject` option. Maybe we should rename the option to `-source-tag`? So we'd have `-source-uri`, `-source-tag`, and `-container-tag`...?

Wdut?

verifiers/internal/gcb/provenance.go
laurentsimon: Added verification for the summary and the metadata. I will do the humand-readable part later. The reason is that it does not follow the intoto format (`slsaProvenance` instead of `predicate`), so I'd like GCB team to fix that

laurentsimon:
Added more tests, ready for review!

main_test.go
```diff
@@ -0,0 +1,427 @@
...
+			artifact:  "binary-linux-amd64-sharded",
+			source:    "github.com/slsa-framework/slsa-verifier",
+			branch:    "release/v1.0",
+			noversion: true,
+		},
+	}
+	for _, tt := range tests {
+		tt := tt // Re-initializing variable so it is not changed while executing the closure below
+		t.Run(tt.name, func(t *testing.T) {
+			// t.Parallel()
```
ianlewis, line 346: remove?

main_test.go
```diff
@@ -0,0 +1,427 @@
...
+
+func pString(s string) *string {
+	return &s
+}
+
+// Versions of the builders to test.
+var golangGeneratorVersions = []string{"v0.0.2", "v1.1.1"}
+
+func Test_runVerifyGo(t *testing.T) {
+	// t.Parallel()
```
ianlewis, line 33: remove?

main_test.go
```diff
@@ -0,0 +1,427 @@
...
+				if !errCmp(err, tt.err) {
+					t.Errorf(cmp.Diff(err, tt.err, cmpopts.EquateErrors()))
+				}
+			}
+		})
+	}
+}
+
+func TestContainerVerification(t *testing.T) {
+	runContainerVerification = func(ctx context.Context, path string) ([]oci.Signature, string, error) {
```
ianlewis, line 376: Where is this defined?

verifiers/internal/gcb/keys/keys.go
ianlewis: +1 I think returning an error here for invalid keys would make more sense than during verification. You can still keep the bytes around if you need them.

verifiers/internal/gcb/keys/keys.go
```diff
@@ -0,0 +1,58 @@
...
+
+	serrors "github.com/slsa-framework/slsa-verifier/errors"
+)
+
+//go:embed materials/*
+var publicKeys embed.FS
+
+type PublicKey struct {
+	Value  []byte
+	Region string
```
ianlewis, line 20: Any reason these need to be exported values?

verifiers/internal/gcb/provenance.go
```diff
@@ -0,0 +1,374 @@
...
+var GCBBuilderIDs = []string{"https://cloudbuild.googleapis.com/GoogleHostedWorker@v0.2"}
+
+type v01IntotoStatement struct {
+	intoto.StatementHeader
+	Predicate slsa01.ProvenancePredicate `json:"predicate"`
+}
+
+type provenance struct {
+	Build struct {
+		// TODO: compare to verified provenance.
```
ianlewis, line 30: Create an issue for this TODO?

Maybe comment out the entire `Build` field if it's empty?

verifiers/internal/gcb/provenance.go
```diff
@@ -0,0 +1,374 @@
...
+		// IntotoStatement v01IntotoStatement `json:"intotoStatement"`
+	} `json:"build"`
+	Kind        string           `json:"kind"`
+	ResourceURI string           `json:"resourceUri"`
+	Envelope    dsselib.Envelope `json:"envelope"`
+}
+
+type gloudProvenance struct {
```
ianlewis, line 38:
```suggestion
type gcloudProvenance struct {
```

verifiers/internal/gcb/provenance.go
```diff
@@ -0,0 +1,374 @@
...
+		FullyQualifiedDigest string `json:"fully_qualified_digest"`
+		Registry             string `json:"registry"`
+		Repsitory            string `json:"repository"`
+	} `json:"image_summary"`
+	ProvenanceSummary struct {
+		Provenance []provenance `json:"provenance"`
+	} `json:"provenance_summary"`
+}
+
+type GCBProvenance struct {
```
ianlewis, line 50: Just make this `Provenance` since it's in the `gcb` package?

verifiers/internal/gcb/provenance.go
```diff
@@ -0,0 +1,374 @@
...
+
+	return &GCBProvenance{
+		gcloudProv: &prov,
+	}, nil
+}
+
+func signatureAsRaw(s string) ([]byte, error) {
+	s = strings.ReplaceAll(s, "-", "+")
+	s = strings.ReplaceAll(s, "_", "/")
+	sig, err := base64.StdEncoding.DecodeString(s)
```
ianlewis, line 71: Should this be using `base64.URLEncoding` rather than trying to replace characters?

verifiers/internal/gcb/provenance.go
```diff
@@ -0,0 +1,374 @@
...
+	} `json:"image_summary"`
+	ProvenanceSummary struct {
+		Provenance []provenance `json:"provenance"`
+	} `json:"provenance_summary"`
+}
+
+type GCBProvenance struct {
+	gcloudProv                    *gloudProvenance
+	verifiedProvenance            *provenance
+	verifiedIntotoStatementStruct *v01IntotoStatement
```
ianlewis, line 53: nit: Having struct in the name is a bit redundant
```suggestion
	verifiedIntotoStatement *v01IntotoStatement
```

verifiers/internal/gcb/provenance.go
```diff
@@ -0,0 +1,374 @@
...
+		return nil, err
+	}
+	d, err := json.Marshal(self.verifiedIntotoStatementStruct)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %s", serrors.ErrorInvalidDssePayload, err.Error())
+	}
+	return d, nil
+}
+
+func (self *GCBProvenance) VerifyMetadata(provenanceOpts *options.ProvenanceOpts) error {
```
ianlewis, line 106: some doc comments on what this is doing would be helpful. Same with other Verify* methods.

verifiers/internal/gcb/testdata/gcloud-container-invalid-slsaheader.json
ianlewis: GCB definitely supports native Cloud Source Repositories

asraa: OK I think I mixed this up: for gha verifier we always expect github.com URIs, but for GCB in `VerifySourceURI`we just expected some URI

I think this is fine then?

laurentsimon: Yes that's correct. I'll double-check when I update this PR. (Leaving this conversation unresolved till then so that I don't forget)

verifiers/internal/gcb/provenance.go
laurentsimon: @ianlewis suggested using the base64 URL decoding instead. If that's enough, I'll use this. @asraa would that allow me to use the "dsse verifier"? Is it defined on their dsse repo?

verifiers/internal/gcb/keys/keys.go
laurentsimon: Done, thanks

verifiers/internal/gcb/provenance.go
laurentsimon: Added to https://github.com/slsa-framework/slsa-verifier/issues/221.

verifiers/internal/gcb/provenance.go
laurentsimon: Updated to use URL decoding.

main_test.go
laurentsimon: @asraa this may be a merge problem from me. Did you intend to comment this or shall I re-enable it?

main_test.go
laurentsimon: @asraa please advise

verifiers/internal/gcb/keys/keys.go
laurentsimon: mistake from me, thanks for catching.

main_test.go
asraa: I think your branch may be out of date: https://github.com/slsa-framework/slsa-verifier/blob/77f8bc623b96f6124ff2cc4a1daade743f01c6b3/cli/slsa-verifier/main_test.go#L66

This can be removed and is at HEAD

main_test.go
asraa: A merge problem, defined in container pkg: https://github.com/slsa-framework/slsa-verifier/blob/77f8bc623b96f6124ff2cc4a1daade743f01c6b3/cli/slsa-verifier/main_test.go#L516

FWIW there is an issue filed here to support container type interfaces, so these funcs vars will go away https://github.com/slsa-framework/slsa-verifier/issues/220

main_test.go
laurentsimon: I had a merge problem, you're correct.

verifiers/internal/gcb/provenance.go
laurentsimon: done.

verifiers/internal/gcb/provenance.go
laurentsimon: done, thanks.

main_test.go
asraa: Running the invididual tests here in parallel might result in rate-limiting, so I would avoid that

verifiers/internal/gcb/provenance.go
laurentsimon: Thank you, done.

main_test.go
laurentsimon: Closing.

main_test.go
laurentsimon: Let me add a coment.

verifiers/internal/gcb/provenance.go
asraa: Ah if it's easier to propogate the errors this way, then keep as is -- it was less readable for an `isX` to return an error

verifiers/internal/gcb/provenance.go
laurentsimon: I removed the comment. PTAL

verifiers/internal/gcb/provenance.go
asraa: Offline discussion: I can update to use dsse verifier uniformly over all the verifier types (GHA, GCB here)

verifiers/internal/gcb/provenance.go
laurentsimon: I can update the name of the funtion. What would our propose?

verifiers/internal/gcb/verifier.go
asraa: Oh, I meant `tag` from source URI, like this was saying. Not container tag. But agree that we should split the option names. This is mentioned in the issue https://github.com/slsa-framework/slsa-verifier/issues/180

Since I'm breaking the API in the next PR we can do it in the next release.

But back on the `tag`: I thought the GCB provenance didn't include source tags, just the commit. At least it seemed so in
```
	if !strings.HasPrefix(uri, expectedSourceURI+"/commit/") {
		return fmt.Errorf("%w: expected '%s', got '%s'",
			serrors.ErrorMismatchSource, expectedSourceURI, uri)
	}
```

verifiers/internal/gcb/provenance.go
laurentsimon: done in https://github.com/slsa-framework/slsa-verifier/issues/221. Thanks

verifiers/internal/gcb/provenance.go
laurentsimon: Good catch, done.

verifiers/internal/gcb/provenance.go
asraa: haha, honestly I can't think of one -- `verified()`? would not worry too much because it's internal anyway

verifiers/internal/gcb/verifier.go
laurentsimon: You're correct. The `VerifyTag` function returns "not supported", like `VerifyBranch` and `VerifyVersionedTag`

verifiers/internal/gcb/provenance.go
laurentsimon: OK, keeping as-is then. Thanks

verifiers/internal/gcb/verifier.go
asraa: Yeah, that's a pretty hard trade-off. I liked the per-verifier README.md you suggested in the documentation PR, so I think it's best to just put there. Otherwise we need to fetch the verifier before we parse args.

asraa (approved):

verifiers/internal/gcb/verifier.go
laurentsimon: So I'll leave this as-in. Confirmed?

Issue #203, opened 2022-08-10
feat: make builder-id experimental


laurentsimon:
Make this option experimental for the optional branch release

PR #207, opened 2022-08-11
feat: make builder-id flag experimental until fully tested

laurentsimon:
closes https://github.com/slsa-framework/slsa-verifier/issues/203

Last issue to address https://github.com/slsa-framework/slsa-verifier/milestone/5

### DIFF
--- a/cli/slsa-verifier/main.go
+++ b/cli/slsa-verifier/main.go
@@ -25,7 +25,8 @@ var (
 )
 
 func experimentalEnabled() bool {
-	return os.Getenv("SLSA_VERIFIER_EXPERIMENTAL") == "1"
+	_, ok := os.LookupEnv("SLSA_VERIFIER_EXPERIMENTAL")
+	return ok
 }
 
 func main() {
--- a/cli/slsa-verifier/main.go
+++ b/cli/slsa-verifier/main.go
@@ -24,8 +24,14 @@ var (
 	printProvenance bool
 )
 
+func experimentalEnabled() bool {
+	return os.Getenv("SLSA_VERIFIER_EXPERIMENTAL") == "1"
+}
+
 func main() {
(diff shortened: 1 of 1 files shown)


cli/slsa-verifier/main.go
```diff
@@ -24,8 +24,14 @@ var (
 	printProvenance bool
 )
 
+func experimentalEnabled() bool {
+	return os.Getenv("SLSA_VERIFIER_EXPERIMENTAL") == "1"
```
asraa, line 28: can we just check if the env is set at all, rather than == 1?

something like
```
_, ok := os.LookupEnv("SLSA_VERIFIER_EXPERIMENTAL")
```

asraa (commented):
other than specifically making it == "1", lgtm

cli/slsa-verifier/main.go
laurentsimon: sure

laurentsimon: done.

asraa (approved):

cli/slsa-verifier/main.go
ianlewis: Personally, I think we should look at it and verify it's not an empty string. It's pretty common practice to set env vars to empty strings to overwrite a previously set value rather than unsetting it.

laurentsimon: yeah, that's what happens when I used to do C/C++/Makefile builds. @asraa wdut?

asraa: That's fair, the convention I was used to in the Sigstore tools was set/unset, and that's fine with me. I agree users sometimes can set a value to disable rather than unset it completely.

laurentsimon: Send the PR ill'll LGTM real quick.

PR #210, opened 2022-08-12
ci: fix path to binary for release

asraa:


laurentsimon (approved):
Good catch. Maybe we need to kick ff this workflow on pre-submit too

laurentsimon:
maybe just using this config in pre-submit would be enough

asraa:
> maybe just using this config in pre-submit would be enough

good point, i'll put a presubmit slsa builder

Issue #211, opened 2022-08-12
Test release config file in pre-submit


laurentsimon:
Let's add a pre-submit that uses the release config file, to avoid https://github.com/slsa-framework/slsa-verifier/pull/210 at release time.
/cc @asraa

PR #212, opened 2022-08-12
ci: add release workflow test at presubmit

asraa:

To ensure that the release workflow will work before we push a release, because maybe the releaser config went stale.

fixes https://github.com/slsa-framework/slsa-verifier/issues/211


asraa:
@laurentsimon do you know why the go builder at v1.2.0 is failing?

laurentsimon:
not sure. We don't officially support PR, maybe there's something going on there. Maybe the OIDC token is failing (not `write` access) and so the builder cannot figure its hash to checkout itself?

laurentsimon:
```
repository:asraa/slsa-verifier
ref:add-test-presubmit
```

so it knows what to fetch. However, there's some code that validates the repo name, and only allow example-package and / or this repo.. maybe that's what's going on.

asraa:
> so it knows what to fetch. However, there's some code that validates the repo name, and only allow example-package and / or this repo.. maybe that's what's going on.

Ah I see. Maybe I should just run this on schedule instead.

laurentsimon:
Btw, how are e going to keep this pre-submit in sync with the release workflow? I mean, that the versions are the same?Shall we use the same workflow and add a trigger to it? Or use a workflow_run trigger and trigger in on PR and during release?
(We'd need to add some code in the release trigger workflow to upload the provenance manually, since the trigger to the workflow_run would be different)

Note: I think everything I said above is probably wrong.

asraa:
> Shall we use the same workflow and add a trigger to it? Or use a workflow_run trigger and trigger in on PR and during release?

Yeah, let's nuke the separate workflow, and use the same one. Instead of on PR, we can trigger once daily or something on schedule just for testing?

This way we can still keep the on push tags for automatic upload

laurentsimon:
> > so it knows what to fetch. However, there's some code that validates the repo name, and only allow example-package and / or this repo.. maybe that's what's going on.
> 
> Ah I see. Maybe I should just run this on schedule instead.

If that simplifies our life, I think that works. Let's create an issue when that happens.

laurentsimon:
Can re-use this code https://github.com/slsa-framework/example-package/blob/main/.github/workflows/scripts/e2e-report-failure.sh

laurentsimon:
> > Shall we use the same workflow and add a trigger to it? Or use a workflow_run trigger and trigger in on PR and during release?
> 
> Yeah, let's nuke the separate workflow, and use the same one. Instead of on PR, we can trigger once daily or something on schedule just for testing?
> 
> This way we can still keep the on push tags for automatic upload

+1. The code is for the generator anyway, so need to run on each PR.

asraa:
nice! done

.github/workflows/release.yml
```diff
@@ -6,9 +6,16 @@ on:
...
     tags:
       - "*" # triggers only if push new tag version, like `0.8.4`.
+  # Run daily as a dry-run/test.
+  schedule:
+    - cron: '0 1 * * *'
 
 permissions: read-all
 
+env:
+  GH_TOKEN: ${{ secrets.E2E_GO_TOKEN }}
```
laurentsimon, line 16: make sure we have this token defined in this repo.

.github/workflows/release.yml
```diff
@@ -20,3 +27,17 @@ jobs:
...
+
+  # In case this fails, e.g. build configuration changes, file an issue in slsa-verifier
+  if-failed:
+    runs-on: ubuntu-latest
+    needs: [builder]
+    if: always() && needs.builder.result == 'failure'
+    steps:
+      - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.4.0
+        with:
+          repository: slsa-framework/example-package
```
laurentsimon, line 39: re-using code like this is great!

laurentsimon (approved):
Thanks!

.github/workflows/release.yml
```diff
@@ -20,3 +27,17 @@ jobs:
...
       config-file: .github/config-release.yml
       compile-builder: true
+
+  # In case this fails, e.g. build configuration changes, file an issue in slsa-verifier
+  if-failed:
+    runs-on: ubuntu-latest
+    needs: [builder]
+    if: always() && needs.builder.result == 'failure'
+    steps:
+      - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.4.0
```
laurentsimon, line 37: nit: either remove the hash or use comment `# tag=v2.4.0` so that renovatebot can send us updates

.github/workflows/release.yml
asraa: aaah! actually i do not have access to add my PAT to the repository. maybe because i'm not an org admin

laurentsimon: Can the default GITHUB_TOKEN create issues? I think it can... Have you checked?

asraa: I think it will! it has issue read/write perms. will remove this one https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

Issue #213, opened 2022-08-12
Create an argument to verify workflow_dispatch input arguments


laurentsimon:
`--workflow_dispatch release_version=v1.2.0, whatever=value` OR
`--workflow-dispatch release_version=v1.2.0, whatever=value` OR
`--workflow-inputs release_version=v1.2.0, whatever=value` OR
`--workflow-input release_version=v1.2.0 --workflow-input whatever=value`

@asraa @ianlewis Wdut? This came up for https://github.com/GoogleContainerTools/jib/blob/master/.github/workflows/jib-cli-release.yml

laurentsimon:
repository documentation to update:
- [ ] https://github.com/GoogleContainerTools/jib/blob/master/jib-cli/README.md

laurentsimon:
Closed by https://github.com/slsa-framework/slsa-verifier/pull/216.

laurentsimon:
Re-opening due to missing e2e tests

asraa:
e2e testing added here https://github.com/slsa-framework/example-package/pull/96 for container, go, and generic

PR #216, opened 2022-08-15
feat: add a `-workflow-input` option

laurentsimon:
See https://github.com/slsa-framework/slsa-verifier/issues/213

This PR uses `-workflow-input key=value` and the option can be passed multiple times.


verifiers/internal/gha/provenance.go
```diff
@@ -224,6 +224,61 @@ func VerifyProvenance(env *dsselib.Envelope, provenanceOpts *options.ProvenanceO
...
+	}
+
+	// Verify all inputs.
+	for k, v := range inputs {
+		value, err := getAsString(pyldInputs, k)
+		if err != nil {
+			return fmt.Errorf("%w: cannot retrieve value of '%s'", serrors.ErrorMismatchWorkflowInputs, k)
+		}
+
+		if v != value {
```
asraa, line 276: do you want a case-insensitive comparison here?

cli/slsa-verifier/main.go
```diff
@@ -42,6 +66,9 @@ func main() {
 		"[optional] expected version the binary was compiled from. Uses semantic version to match the tag")
 	flag.BoolVar(&printProvenance, "print-provenance", false,
 		"print the verified provenance to std out")
+	inputs.kv = make(map[string]string)
+	flag.Var(&inputs, "workflow-input",
+		"[optional] a workflow input provided by a user at trigger time (e.g., workflow_dispatch) in the format 'key=value'.")
```
asraa, line 71: This only supports workflow_dispatch, so e.g. should be clear that this is the only supported case

cli/slsa-verifier/main.go
```diff
@@ -21,13 +27,31 @@ var (
...
+	inputs          workflowInputs
 	printProvenance bool
 )
 
 func experimentalEnabled() bool {
 	return os.Getenv("SLSA_VERIFIER_EXPERIMENTAL") == "1"
 }
 
+func (i *workflowInputs) String() string {
+	return "some representation"
```
asraa, line 39: `fmt.Sprint` or `json.Marshal` could work here

cli/slsa-verifier/main.go
laurentsimon: Done, PTAL

cli/slsa-verifier/main.go
laurentsimon: done with `Sprintf`

laurentsimon:
fyi @elefeint we're going to cut a release this week. We'll be able to update your project's documentation `-workflow-input release-version=x.y.z`. I'll send a PR.

PR #217, opened 2022-08-15
fix: fix dir in release config

asraa:


laurentsimon (approved):
Thanks.
Did you forget to update the doc? I think you said the doc have an incorrect example.

asraa:
> Thanks.
Did you forget to update the doc? I think you said the doc have an incorrect example.

That's in the generator repo!

Issue #229, opened 2022-08-23
Add a -version flag for displaying version information


sethmlarson:
Without a `-version` flag it's tough for users to figure out which version they're running locally. This came up while trying to figure out why a user wasn't able to verify a binary due to an issue that was fixed in a later version of slsa-verifier than the one they were using.

Issue #205, opened 2022-08-10
Create subcommands for artifact and image verification


asraa:
Follow up for https://github.com/slsa-framework/slsa-verifier/pull/147

Also separate the verification funcs

asraa:
And separate `verifier.VerifyArtifact`

Issue #173, opened 2022-07-26
Enhacement: Help message should exemplify arguments for run


mihaimaruseac:
Installed `slsa-verifier` and tried to run it without passing any flag

```
mihaimaruseac@ankh:/tmp$ ~/go/bin/slsa-verifier attestation.intoto.jsonl 
Usage of /usr/local/google/home/mihaimaruseac/go/bin/slsa-verifier:
  -alsologtostderr
        log to standard error as well as files
  -artifact-path string
        path to an artifact to verify
  -branch string
        expected branch the binary was compiled from (default "main")
  -log_backtrace_at value
        when logging hits line file:N, emit a stack trace
  -log_dir string
        If non-empty, write log files in this directory
  -logtostderr
        log to standard error instead of files
  -print-provenance
        print the verified provenance to std out
  -provenance string
        path to a provenance file
  -source string
        expected source repository that should have produced the binary, e.g. github.com/some/repo
  -stderrthreshold value
        logs at or above this threshold go to stderr
  -tag string
        [optional] expected tag the binary was compiled from
  -v value
        log level for V logs
  -versioned-tag string
        [optional] expected version the binary was compiled from. Uses semantic version to match the tag
  -vmodule value
        comma-separated list of pattern=N settings for file-filtered logging
```

I managed to get a run without the help screen being shown up by passing 3 separate flags:

```
mihaimaruseac@ankh:/tmp$ ~/go/bin/slsa-verifier -provenance attestation.intoto.jsonl -source mihaimaruseac/slsa-lvl3-generic-provenance-with-bazel-example -artifact-path fib
Verified signature against tlog entry index 3026659 at URL: https://rekor.sigstore.dev/api/v1/log/entries/b7d5d52dfc795dddf1c03e901e96c3cb1803c9f1ecc250a13299131b2564c422
Verified build using builder https://github.com/slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@refs/tags/v1.2.0 at commit e9b5431bcfb0df5d8b2ee622bde8ec4fad9d71d3
PASSED: Verified SLSA provenance
```

I think the help text should display that these falgs are required and what is the format expected for them (e.g., source should not be a http link but in `owner/repo` format)

(Note that passing flags with `--` also works)

I could work on this :)

laurentsimon:
> I think the help text should display that these falgs are required and what is the format expected for them (e.g., source should not be a http link but in `owner/repo` format)

[...]

Issue #93, opened 2022-06-09
Rename `--provenance` to `---provenance-path`


laurentsimon:
Just a suggestion.

Wdut?

@asraa @ianlewis

asraa:
I have no preference here, but maybe it makes sense to differentiate between inline content and files

laurentsimon:
Shall we do this before v1 announcement or too risky :-)? (update doc, fix e2e test, fix blog post)

PR #231, opened 2022-08-23
refactor: add subcommands and separate functionality from artifacts a…

asraa:

This PR does the following:
* Add sub-commands `verify-artifact` and `verify-image` to the CLI https://github.com/slsa-framework/slsa-verifier/issues/205
* Separates `verifiers` function call and the artifact image digest calculation
* Adds an explicit version command to fix https://github.com/slsa-framework/slsa-verifier/issues/229. Note that this is a flag generated at build time. 
* Help message is auto-generated. See https://github.com/slsa-framework/slsa-verifier/issues/173
* Renames `--provenance` to `--provenance-path`. See https://github.com/slsa-framework/slsa-verifier/issues/93
* Update README https://github.com/slsa-framework/slsa-verifier/issues/186 and documentation of options https://github.com/slsa-framework/slsa-verifier/issues/230

Fixes https://github.com/slsa-framework/slsa-verifier/issues/205
Fixes https://github.com/slsa-framework/slsa-verifier/issues/229
Fixes https://github.com/slsa-framework/slsa-verifier/issues/173
Fixes https://github.com/slsa-framework/slsa-verifier/issues/93

TODO:
* Update usage of verifier at HEAD in example-package


asraa:
@laurentsimon qq: what is the pupose of testing `outBuilderID` from the main CLI? it was never returned or output before

laurentsimon:
> @laurentsimon qq: what is the pupose of testing `outBuilderID` from the main CLI? it was never returned or output before

I think to display it (someday), the same way that we display information about repo / commit hash. More work is needed to decide what format we need to display it, though.

Wdut?

asraa:
>  More work is needed to decide what format we need to display it, though.

Gotcha: I can make the command return the outBuilderID as string for now to satisfy the tests (e.g. `cmd.Exec(ctx, args) returns (string, error)` Later we can formalize on that output.

laurentsimon:
> > More work is needed to decide what format we need to display it, though.
> 
> Gotcha: I can make the command return the outBuilderID as string for now to satisfy the tests (e.g. `cmd.Exec(ctx, args) returns (string, error)` Later we can formalize on that output.

SG

README.md
```diff
@@ -49,41 +49,56 @@ $ sha256sum -c --strict SHA256SUM.md
...
 The verified in-toto statement may be written to stdout with the `--print-provenance` flag to pipe into policy engines.
 
+### Options Details
+
+For provenance created using SLSA GitHub generators:
+
+| Option | Description |
+| --- | ----------- |
+| `source` | Expects a source, for e.g. `github.com/org/repo`. This option verifies that `ConfigSource` and `Materials` of the SLSA provenance contain the expected `source`. |
+| `branch` | Expects a `branch` like `main` or `dev`. Verifies this against the `github_ref` in the SLSA provenance when the `github_ref_type` is a `branch`, or the `base_ref` or `target_commitish` when a `tag`.|
```
laurentsimon, line 96: Let's add that is not supported for all triggers. It depends on GH payload. If trigger is on an annotated tag, it's not available. If a tag is not annotated, but the tag is made *not* from the HEAD of a branch (say, an older commit from main), it's not supported either.

laurentsimon, line 96: I wonder if we need this level of details here. I'm worried it will go out of sync with the implementation.
How about:
1. Keep the documentation simpler in this README. Maybe we can have a table that explains which options are supported for which builder, and that the branch option is brittle.
2. If we want more detailed doc, maybe have a README under `verifiers/internal/<builder>/README.md`. There is still a risk of getting out of sync, of course....

Anything we could do better that what I'm proposing?

laurentsimon: Also, not supported on manual workflow_dispatch (`-workflow-input` is needed instead and depends on each project).

asraa: Simplified the README.md. I mostly want to make sure that there's documentation for what the options should look like, with some examples in this PR

Eveneually we should have some documentation somewhere of how verification works for specific builders.

.github/config-release.yml
```diff
@@ -11,4 +11,7 @@ flags:
 goos: linux
 goarch: amd64
 binary: slsa-verifier-{{ .Os }}-{{ .Arch }}
-dir: ./cli/slsa-verifier
\ No newline at end of file
+dir: ./cli/slsa-verifier
+
+ldflags:
+  - "-X version.Version={{ .Version }}"
```
laurentsimon, line 17: nit: add new line.

README.md
```diff
@@ -49,41 +49,56 @@ $ sha256sum -c --strict SHA256SUM.md
...
+Verifies SLSA provenance on an artifact blob
+
+Usage:
+  slsa-verifier verify-artifact [flags]
+
+Flags:
+      --branch string            [optional] expected branch the binary was compiled from
+      --builder-id string        EXPERIMENTAL: the unique builder ID who created the provenance
+  -h, --help                     help for verify-artifact
+      --print-provenance         print the verified provenance to std out
```
laurentsimon, line 70: nit: stdout

README.md
```diff
@@ -49,41 +49,56 @@ $ sha256sum -c --strict SHA256SUM.md
...
+      --provenance-path string   path to a provenance file
+      --source string            expected source repository that should have produced the binary, e.g. github.com/some/repo
```
laurentsimon, line 72: do we need to rename to `-source-uri`? How about `-source-tag`, `-source-branch` and `source-versioned-tag`?
I am trying to think whether we may hit name collision in the future, like `-container-tag` or something else.

Also, since SLSA has 3 main requirements: `source`, `build` and `provenance`, I'm wondering if we should try to align with this as well.

README.md
```diff
@@ -49,41 +49,56 @@ $ sha256sum -c --strict SHA256SUM.md
...
+      --tag string               [optional] expected tag the binary was compiled from
+      --versioned-tag string     [optional] expected version the binary was compiled from. Uses semantic version to match the tag
+      --workflow-input map[]     [optional] a workflow input provided by a user at trigger time in the format 'key=value'. (Only for 'workflow_dispatch' events). (default map[])
```
laurentsimon, line 75: this would (maybe ?) be renamed to `build-inputs` if we follow the `build` / `source` / `provenance` requirements's naming convention. I'm not 100% sure `build-input` is better than `workflow-input`... maybe `build-workflow-input`...

README.md
```diff
@@ -49,41 +49,56 @@ $ sha256sum -c --strict SHA256SUM.md
...
 Verified signature against tlog entry index 3189970 at URL: https://rekor.sigstore.dev/api/v1/log/entries/206071d5ca7a2346e4db4dcb19a648c7f13b4957e655f4382b735894059bd199
 Verified build using builder https://github.com/slsa-framework/slsa-github-generator/.github/workflows/builder_go_slsa3.yml@refs/tags/v1.2.0 at commit 5bb13ef508b2b8ded49f9264d7712f1316830d10
 PASSED: Verified SLSA provenance
 ```
 
 The verified in-toto statement may be written to stdout with the `--print-provenance` flag to pipe into policy engines.
 
+### Options Details
+
+The following options are supported for SLSA GitHub builders and generators:
```
laurentsimon, line 91: Add a link [SLSA GitHub builders and generators](link-to-repo)

README.md
```diff
@@ -49,41 +49,56 @@ $ sha256sum -c --strict SHA256SUM.md
...
+
+| Option | Description |
+| --- | ----------- |
+| `source` | Expects a source, for e.g. `github.com/org/repo`. |
+| `branch` | Expects a `branch` like `main` or `dev`. Not supported or all GitHub Workflow triggers. |
```
laurentsimon, line 96: nit: for all GitHub

README.md
```diff
@@ -49,41 +49,56 @@ $ sha256sum -c --strict SHA256SUM.md
...
+| `tag` | Expects a  `tag` like `v0.0.1`. Verifies exact tag used to create the binary. Not supported for all triggers. |
```
laurentsimon, line 97: Maybe say it's supported for new tag and release triggers, with a link to some examples?

README.md
```diff
@@ -49,41 +49,56 @@ $ sha256sum -c --strict SHA256SUM.md
...
+| `versioned-tag` | Like `tag`, but verifies using semantic versioning. |
+| `workflow-input` | Expects key-value pairs like `key=value` |
```
laurentsimon, line 99: Maybe say it's for build triggered via `workflow_dispatch`, with a link to its doc or an example?

cli/slsa-verifier/verify/verify_image.go
```diff
@@ -0,0 +1,79 @@
...
+	if c.ProvenancePath != nil {
+		provenance, err = os.ReadFile(*c.ProvenancePath)
+		if err != nil {
+			return "", err
+		}
+	}
+
+	verifiedProvenance, outBuilderID, err := verifiers.VerifyImage(ctx, artifacts[0], provenance, provenanceOpts, builderOpts)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "FAILED: SLSA verification failed: %v\n", err)
```
laurentsimon, line 69: shall we create a string for FAILED / PASSED which is common across the verify commands?

cli/slsa-verifier/verify/verify_image.go
```diff
@@ -0,0 +1,79 @@
...
+	Source          string
+	Branch          *string
+	Tag             *string
+	VersionTag      *string
+	Inputs          map[string]string
+	PrintProvenance bool
+}
+
+func (c *VerifyImageCommand) Exec(ctx context.Context, artifacts []string) (string, error) {
+	artifactHash, err := container.GetImageDigest(artifacts[0])
```
laurentsimon, line 41: Let's return an error if more than one artifact is passed? I don't know if this is possible, but just in case, so that we don't silently ignore it.

laurentsimon (commented):
Looks really nice!
Since we're at it, I've added a bunch of comments about argument naming. Let me know what you think.

cli/slsa-verifier/verify/verify_image.go
asraa: This is already done: it's handled in the cobra code:
https://github.com/slsa-framework/slsa-verifier/blob/1f42a806ae5412bb2d00f6bc6b1e282aebeb436c/cli/slsa-verifier/verify.go#L30-L34

README.md
asraa: I think `build-inputs` is an accurate term for workflow dispatch inputs, but i wonder if that would get conflated with other types of inputs.

For GHA builders: the only inputs to workflows is through workflow dispatch inputs. But what about inputs to the trusted builders themselves?

cli/slsa-verifier/verify/verify_image.go
laurentsimon: Thanks. Maybe just add a comment here then. I don't mind having a redundant check here to make it explicit to readers.

README.md
laurentsimon: full agreed. That's why I hinted at `-build-workflow-inputs`., to make it more explicit that it's for (GitHub) workflows. Not sure this is enough. Is it possible to name the option after the type of builders that support it? `-build-gha-inputs`. I'm making the name up...

README.md
asraa: I agree that `source` should prefix the `tag` and `branch` and URI should be qualified for what `source` means. Added for `source`

README.md
asraa: I think formatting inputs is brittle and if we are considering that it is likely not good. If they are builder specific, I think some options are already interpreted slightly different -- so `build-input` for GHA could mean workflow inputs. 

OTOH I don't know how to distinguish between types of build inputs for a single builder. If someone wants to check on inputs to the trusted builders, can they access the provenance, since that is more of a "recipe"?

asraa: For now, I think it's easier to organize as `build-XYZ` or `source-XYZ` of the current options. At least it organizes the options.

laurentsimon: SGTM. 

> OTOH I don't know how to distinguish between types of build inputs for a single builder. If someone wants to check on inputs to the trusted builders, can they access the provenance, since that is more of a "recipe"?

fair enough. Let's start with that then. I am not even sure GCB supports "input", since it's just a webhook.

@ianlewis any objection the direction we're taking here?

cli/slsa-verifier/verify/options.go
```diff
@@ -0,0 +1,97 @@
...
+
+	cmd.Flags().StringVar(&o.SourceTag, "source-tag", "", "[optional] expected tag the binary was compiled from")
+
+	cmd.Flags().StringVar(&o.SourceVersionTag, "source-versioned-tag", "",
+		"[optional] expected version the binary was compiled from. Uses semantic version to match the tag")
+
+	cmd.Flags().BoolVar(&o.PrintProvenance, "print-provenance", false,
+		"print the verified provenance to stdout")
+
+	cmd.Flags().Var(&o.BuildWorkflowInputs, "build-workflow-input",
```
laurentsimon, line 67: I'm curious if the library orders option alphabetically or whether it uses the order we declare them in. Maybe let's group the option by `source` and `build`, it will be easier for readers.

laurentsimon (approved):
LGTM, thanks

cli/slsa-verifier/verify/options.go
asraa: They're ordered alphebetically in the help description, this ordering is only in the code.

asraa:
> How do we handle merging this PR? Do we merge or wait for the GCB release to be cut?

Given that GCB is breaking too, I think it would be good to merge all in one. That way GCB verification ONLY works with these new CLI arguments, and GCB folks won't have to deal with a breaking API change.

laurentsimon:
alright. Let's make it happen then.

cli/slsa-verifier/verify/options.go
laurentsimon: nit: we can still group the `build-` and `source-` together in the code.

laurentsimon, line 67: btw, did you say you wanted to make this option `-build-input`?

asraa: I thought simpler to include a migration from `workflow-input` to `build-workflow-input` like the `source` ones. i think `build-input` sounds better, but i'm hesitant it might mean builder inputs

laurentsimon: SGTM. Also, it makes it clearer it's for (GH) workflows.

asraa: done

asraa:
@laurentsimon merged your changes on compute digest FN and GCB provenance. Good to merge with https://github.com/slsa-framework/example-package/pull/100?

laurentsimon:
LGTM!

laurentsimon:
What's `FN`?

asraa:
> What's FN?

FN is the compute digest FN you had earlier in the GCB support -- I think we will remove with a better container interface, but was needed to swap in for testing

PR #233, opened 2022-08-24
Initial POC for installer action.

kpk47:


actions/installer.sh
```diff
@@ -0,0 +1,108 @@
+#!/bin/bash
```
ianlewis, line 1: nit: While not totally necessary we've tried to get away from writing a lot of bash for actions. Is there any way this could be written in TypeScript?

See the compute-sha256 actions in the slsa-github-generator repo for an example.
https://github.com/slsa-framework/slsa-github-generator/tree/main/.github/actions/compute-sha256

actions/README.md
```diff
@@ -0,0 +1,33 @@
+# slsa-verifier-installer GitHub Action
```
ianlewis, line 1: I'm not sure what the current best practices are here but most actions (but not all) take the form `setup-*` for installing tools in the GitHub Actions workspace. The [official ones](https://github.com/actions?q=setup&type=all&language=&sort=) use this convention. I'm wondering if we should follow that convention.

@laurentsimon @asraa @naveensrinivasan WDUT?

actions/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+For a full example workflow, see `.github/workflows/test.yml`.
+
+### Optional Inputs
+
+| Input | Description |
+| --- | --- |
+| `verifier-release` | `slsa-verifier` version to use. Default value is `v1.3.0`. |
+| `install-dir` | Where to install `slsa-verifier` binary. Default value is `$HOME/slsa-verifier`. |
+| `use-sudo` | Whether the install location requires `sudo`. Default value is `false`.
```
ianlewis, line 31: Do we really need this? I'm pretty user actions just run as root in their environment anyway.

actions/README.md
```diff
@@ -0,0 +1,33 @@
...
+```
+
+For a full example workflow, see `.github/workflows/test.yml`.
+
+### Optional Inputs
+
+| Input | Description |
+| --- | --- |
+| `verifier-release` | `slsa-verifier` version to use. Default value is `v1.3.0`. |
+| `install-dir` | Where to install `slsa-verifier` binary. Default value is `$HOME/slsa-verifier`. |
```
ianlewis, line 30: nit: The action could just decide the location itself and add that location to the PATH. Example in setup-go:
https://github.com/actions/setup-go/blob/268d8c0ca0432bb2cf416faae41297df9d262d7f/src/main.ts#L34-L41

Also ideally the installer action would support installing multiple versions of the verifier in a matrix. I think Go does this by installing different versions of go in different locations, but each time it's installed the PATH is updated so the latest installed takes precedence. Ex:
https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-go#using-multiple-versions-of-go

actions/installer.sh
```diff
@@ -0,0 +1,108 @@
...
+# SLSA verifier install script
+
+VERIFIER_RELEASE="$1"
```
ianlewis, line 4: nit: If we stick to bash I'd like us to try and stick to using lowercase local (unexported) variables since it's easier to parse which are exported pre-requisites and which are local. Also, linters like `shellcheck` use this convention when reporting issues.

.github/workflows/test_installer.yml
```diff
@@ -0,0 +1,138 @@
+on: [push]
+
+jobs:
```
ianlewis, line 3: nit: You could probably use a matrix strategy for the tests below.
https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

ianlewis (commented):
Thanks!

actions/installer.yml
```diff
@@ -0,0 +1,27 @@
+name: slsa-verifier-installer
+description: 'Installs SLSA verifier and adds it to your PATH'
+branding:
+  icon: 'package'
+  color: 'blue'
+inputs:
+  verifier-release:
```
laurentsimon, line 7: If we follow a model `uses: slsa-framework/slsa-verifier/actions/install@v1.2.3`, then we won't need this option, I think.
I think cosign has this for flexibility and pre-GA / testing. In our case, we don't need to give users this ability. If people really ask, we could add it later. But for now let's keep things simple

actions/installer.yml
```diff
@@ -0,0 +1,27 @@
...
+    description: 'slsa-verifier release version to install'
+    required: false
+    default: 'v1.2.0'
+  install-dir:
```
laurentsimon, line 11: Do we need this option? What's the use case?

actions/installer.yml
```diff
@@ -0,0 +1,27 @@
...
+    description: 'Absolute path to installation directory'
+    required: false
+    default: '/usr/bin/'
+  use-sudo:
```
laurentsimon, line 15: If we only support `/usr/bin`, we can remove this option. If there's a directory that does not require root (`opt/` ?), we could use it instead.

actions/installer.sh
```diff
@@ -0,0 +1,108 @@
...
+if [[ "${VERIFIER_RELEASE}" == "main" ]]; then
+    log_info "installing verifier via 'go install' from its main version"
+    GOBIN=$(go env GOPATH)/bin
+    go install github.com/slsa-framework/slsa-verifier/cli/slsa-verifier@main
+    ${SUDO} ln -s ${GOBIN}/slsa-verifier ${INSTALL_DIR}/slsa-verifier
+    echo "Installed SLSA verifier at ${INSTALL_DIR}/slsa-verifier"
+    exit 0
+fi
+
+# Download v1.2.0 for bootstrapping
```
laurentsimon, line 33: The simplest way to go about installation and reduce maintenance for us to the minimum is something like the following:
1. Download a pre-defined version of the verifier and verify its hash (may be able to call   https://github.com/slsa-framework/slsa-github-generator/tree/main/.github/actions/compute-sha256 to calculate the hash without fiddling with sha256sum yourself). This need not be updated regularly, or at all (unless a vulnerability is reported)
2. Use this version to download the version the user pinned the action to (`@v1.2.3`): we can use `GITHUB_REF_NAME`
3. Figure a way to map a hash to a release (in case of `@<hash>`), and download like in previous step. Take a look at the GitHub context to see if the tag is accessible. If not, let's look into GitHub APIs. 

@ianlewis does the above sound resoanable?

actions/README.md
laurentsimon: I'm also in favor of not exposing such option.

actions/installer.sh
laurentsimon: You can install shellcheck in VSCode to get warning about this, I think

actions/installer.yml
laurentsimon: follow-up: A folder under the user's home directory may be a better option with the need for sudo'ing, like `~/.slsa/bin/`

actions/README.md
naveensrinivasan: I agree, we should probably be implementing it is TS IMO

actions/installer.yml
ianlewis: I think these actions usually download a release rather than build from source at a particular tag but I guess we could make it download v1.2.3 if that's the referenced tag. I'm not sure if it's easy for the action to determine the tag it was referenced by though.

actions/installer.sh
ianlewis: Is `GITHUB_REF_NAME` the ref of the action or the ref from the calling workflow's context? I think it's the latter. At least for reusable workflows we had to get this ref via the OIDC token IIRC.
https://github.com/slsa-framework/slsa-github-generator/blob/c565ad10f0c2efdc6fcc8fe55b52845f85014dfa/.github/actions/detect-workflow/main.go#L104-L129

actions/README.md
laurentsimon: re: `setup-*` that sounds like a great idea. Would we call it just `setup`? The setup actions you link to are under the `actions` org, so they need a `setup-<name>`. Since this action is for the verify, do we ned to further qualify it? If so, what would it be?

re: TS. This would be ideal. Since this code is fairly small and hash verification (super critical) is implemented in TS, how about we see how much bash code is required to make this work? If it becomes too large, we update it to TS next iteration? I feel like we can get away with just a few lines of bash for this.

Feel free to disagree, though.

actions/installer.yml
ianlewis: Another issue is that we may have to fix bugs in the action after a version is released, but still allow users to download that version of the verifier. i.e. v2.0.0 has a bug in the installer but if we update to v2.0.1 then we have to release an identical version of the verifier as v2.0.1 as well which is a bit awkward.

actions/README.md
laurentsimon: I'm also in favor of not exposing such an option, unless there is a compelling use case for it.

.github/workflows/test_installer.yml
```diff
@@ -0,0 +1,138 @@
...
+    name: Install in custom directory
+    steps:
+      - name: Install verifier v1.3.0
+        uses: slsa-framework/slsa-verifier/actions/install@main
+        with:
+          install-dir: /tmp/.slsa-verifier
+      - name: Run verifier help
+        run: slsa-verifier --help
+        shell: bash
+  installer_job_sudo_enabled:
```
laurentsimon, line 128: I can't think of a way to test failure due to hash mismatch or provenance verification failure on the fetched binary. If there's a way we can do that., that'd be cool. Bu don't waste time on it :)

.github/workflows/test_installer.yml
laurentsimon: Great idea

actions/installer.sh
laurentsimon: I think you're right, it's the caller's ref. Any idea how to get our own ref without resorting to OIDC?

actions/installer.yml
laurentsimon: > I think these actions usually download a release rather than build from source at a particular tag but I guess we could make it download v1.2.3 if that's the referenced tag. I'm not sure if it's easy for the action to determine the tag it was referenced by though.

mhh, let's hope we don't need `token-id:write permissions` :/

ianlewis: Personally, I think that having the option is ok. Folks can use the latest version of the action and pass what version of the verifier they want to install and limit it to releases (rather than any git ref). The same way that other actions like `setup-go` do.

laurentsimon: Point taken. Although it may seem awkward to have `@v1.2.3` and an argument saying `version: v1.5.6`. But if we think this is OK, I'm not against it.

BTW, I found a way to get the version without using OIDC: The GitHub context now contains an `action_path` that looks like this:

`"action_path": "/home/runner/work/_actions/laurentsimon/hello-github-actions/v1.0.0/action-a"`, corresponding to the code below in the calling workflow. 
```
uses: laurentsimon/hello-github-actions/action-a@v1.0.0
```

actions/installer.sh
laurentsimon: Copying from another comment:

The GitHub context contains an `action_path` that looks like this:

`"action_path": "/home/runner/work/_actions/laurentsimon/hello-github-actions/v1.0.0/action-a"`, corresponding to the code below in the calling workflow. 
```
uses: laurentsimon/hello-github-actions/action-a@v1.0.0
```

actions/installer.yml
laurentsimon: Thinking more about this, I think @ianlewis is right about allowing such an option and decoupling the action ref from the version to install. If @asraa is fine with this, let's do that.

.github/workflows/test_installer.yml
kpk47: I added a couple of special version tags that will force checksum and provenance errors and wrote tests that use them. I don't love the solution, but it works. WDYT?

.github/workflows/test_installer.yml
```diff
@@ -0,0 +1,138 @@
...
+          go-version: 1.18
+          check-latest: true
+      - name: Install verifier at main
+        uses: slsa-framework/slsa-verifier/actions/install@main
+        with:
+          verifier-release: main
+      - name: Run verifier help
+        run: slsa-verifier --help
+        shell: bash
+  installer_job_custom_path:
```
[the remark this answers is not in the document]
kpk47: Done

actions/README.md
kpk47: I went with ```setup``` for the name. 

I tried using the SHA256 action from slsa-framework/slsa-github-generator, and I didn't like the outcome. (You can take a look at https://github.com/kpk47/slsa-verifier/tree/aborted-wf if you're interested.) I'm happy to rewrite this action in TS since it ended up being quite a lot of bash.

.github/workflows/test_installer.yml
kpk47: Done

actions/installer.sh
kpk47: Done

actions/README.md
kpk47: Done

actions/README.md
kpk47: Done

actions/installer.yml
```diff
@@ -0,0 +1,27 @@
+name: slsa-verifier-installer
```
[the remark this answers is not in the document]
kpk47: Done

actions/installer.yml
kpk47: I kept the ```verifier-release``` parameter and agree we should version the action separately to verifier releases. FWIW, GitHub recommends versioning actions and application code separately: https://docs.github.com/en/actions/creating-actions/about-custom-actions#choosing-a-location-for-your-action.

actions/installer.yml
kpk47: Done

actions/installer.yml
kpk47: Done

actions/README.md
kpk47: Done

actions/installer.sh
kpk47: Done

.github/workflows/test_install.yml
```diff
@@ -0,0 +1,51 @@
+on: [push]
+
+jobs:
+  installer_test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        verifier-release: [main, v0.0.1, v1.0.0, v1.0.1, v1.0.2, v1.1.0, v1.1.1, v1.2.0, v1.3.0]
```
laurentsimon, line 8: this is going to get out of sync quickly. Any ideas?
One idea I have is to list the versions available for the repo, and create the matrix dynamically. You can check online that this is possible - I think it makes use of script injections.. but it works :-)

.github/workflows/test_install.yml
```diff
@@ -0,0 +1,51 @@
+on: [push]
+
```
laurentsimon, line 2: always declare permissions:
```
permissions: read-all
```

.github/workflows/test_install.yml
```diff
@@ -0,0 +1,51 @@
+on: [push]
```
laurentsimon, line 1: maybe we should do this on a daily schedule; and crate an issue automatically if it fails. We have an example on the generator repo.

.github/workflows/test_install.yml
kpk47: Looks like there's an API for listing releases: https://docs.github.com/en/rest/releases/releases#list-releases

.github/workflows/test_install.yml
```diff
@@ -0,0 +1,51 @@
...
+          fi
+          exit 1
+  invalid_hash_test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v3
+      - shell: bash
+        run: |
+          if ! $GITHUB_WORKSPACE/actions/install/installer.sh test-force-checksum-error ; then
```
laurentsimon, line 37: why not use `uses: ./actions/install` as well?

actions/install/README.md
```diff
@@ -0,0 +1,29 @@
+# slsa-verifier-installer GitHub Action
+
+This action installs the SLSA verifier so you can verify SLSA provenance in your workflows.
+
+For more information about `slsa-verifier`, refer to the [its documentation](https://github.com/slsa-framework/slsa-verifier#verification-of-provenance).
```
laurentsimon, line 5: nit: the its

actions/install/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+For more information about SLSA in general, see https://slsa.dev.
```
laurentsimon, line 7: create a markdown link.

actions/install/README.md
```diff
@@ -0,0 +1,29 @@
+# slsa-verifier-installer GitHub Action
```
laurentsimon, line 1: Add a section under the main README that link to this README for installing the action

actions/install/installer.sh
```diff
@@ -0,0 +1,94 @@
+#!/bin/bash
+# SLSA verifier install script
```
laurentsimon, line 2: Always add `set -euo pipefail` otherwise errors are silently ignored

actions/install/installer.sh
```diff
@@ -0,0 +1,94 @@
...
+
+shopt -s expand_aliases
+if [ -z "$NO_COLOR" ]; then
+  alias log_info="echo -e \"\033[1;32mINFO\033[0m:\""
+  alias log_error="echo -e \"\033[1;31mERROR\033[0m:\""
+else
+  alias log_info="echo \"INFO:\""
+  alias log_error="echo \"ERROR:\""
+fi
+set -e
```
laurentsimon, line 15: put it at the top of the file. Also, `set -euo pipefail` is more comprehensive, I think

actions/install/installer.sh
```diff
@@ -0,0 +1,94 @@
...
+
+mkdir -p ${install_dir}
+cd ${install_dir}
+
+if [[ "${verifier_release}" == "main" ]]; then
+    log_info "installing verifier via 'go install' from its main version"
+    gobin=$(go env GOPATH)/bin
```
laurentsimon, line 22: always quote variables `"some_var"`

actions/install/installer.sh
```diff
@@ -0,0 +1,94 @@
...
+    go install github.com/slsa-framework/slsa-verifier/cli/slsa-verifier@main
+    ln -s ${gobin}/slsa-verifier ${install_dir}/slsa-verifier
+    echo "Installed SLSA verifier at ${install_dir}/slsa-verifier"
+    exit 0
+fi
+
+# Download v1.2.0 for bootstrapping
+binary_addr="https://github.com/slsa-framework/slsa-verifier/releases/download/v1.2.0/slsa-verifier-linux-amd64"
+sha256="37db23392c7918bb4e243cdb097ed5f9d14b9b965dc1905b25bc2d1c0c91bf3d slsa-verifier-linux-amd64"
+if [[ $verifier_release == "test-force-checksum-error" ]]; then
```
laurentsimon, line 32: quote variable

actions/install/installer.sh
```diff
@@ -0,0 +1,94 @@
...
+  binary_addr="https://github.com/slsa-framework/slsa-verifier/releases/download/v0.0.1/slsa-verifier-linux-amd64"
+fi
+curl -sL ${binary_addr} -o slsa-verifier-linux-amd64
```
laurentsimon, line 35: quote variables

actions/install/installer.sh
```diff
@@ -0,0 +1,94 @@
...
+echo ${sha256} >> SHA256SUM.md
+
+if ! sha256sum --quiet -c SHA256SUM.md ; then
```
laurentsimon, line 38: it would be nice if you could try to re-use the slsa-github-generator action (https://github.com/slsa-framework/slsa-github-generator/tree/main/.github/actions/compute-sha256) as part of the composite actions, since the sha256sum command is hard to read

laurentsimon (commented):
Thanks!
Left a few comments to try to further simplify.

actions/install/action.yml
```diff
@@ -0,0 +1,17 @@
...
+inputs:
+  verifier-release:
+    description: 'slsa-verifier release version to install'
+    required: false
+    default: 'v1.3.0'
+runs:
+  using: "composite"
+  steps:
+    - run: |
+        ${{ github.action_path }}/installer.sh ${{ inputs.verifier-release }}
```
laurentsimon, line 15: always use env variables when using variables:
```
env:
  PATH: "${{ github.action_path }}"
run: |
   bla "$PATH"
```

In this case the user would just hack themselves so it does not matter, but it's good practice to do it to learn the habit :)

.github/workflows/test_install.yml
kpk47: I want the job to succeed iff the step fails, and afaict GitHub doesn't support that. I found an open issue for a feature that would enable this use case, but it doesn't look like GH is acting on it (https://github.com/actions/toolkit/issues/399).

actions/install/action.yml
kpk47: Done

actions/install/installer.sh
kpk47: Done

actions/install/installer.sh
kpk47: Done

actions/install/installer.sh
kpk47: Done

actions/install/README.md
kpk47: Done

actions/install/README.md
kpk47: Done

actions/install/README.md
kpk47: Done

.github/workflows/test_install.yml
kpk47: Done

actions/install/installer.sh
kpk47: Done

actions/install/installer.sh
```diff
@@ -0,0 +1,89 @@
...
+binary_addr="https://github.com/slsa-framework/slsa-verifier/releases/download/v1.2.0/slsa-verifier-linux-amd64"
+sha256="37db23392c7918bb4e243cdb097ed5f9d14b9b965dc1905b25bc2d1c0c91bf3d slsa-verifier-linux-amd64"
+if [[ "$verifier_release" == "test-force-checksum-error" ]]; then
+  binary_addr="https://github.com/slsa-framework/slsa-verifier/releases/download/v0.0.1/slsa-verifier-linux-amd64"
+fi
+curl -sL "$binary_addr" -o slsa-verifier-linux-amd64
+echo "$sha256" >> SHA256SUM.md
+
+if ! sha256sum --quiet -c SHA256SUM.md ; then
+  rm slsa-verifier-linux-amd64
```
laurentsimon, line 34: Hw about creating a temporary folder with all files needed, and do a `rm *` at the end? This way we don't have to worry about which files we need to remove based on how far the script has executed. We could also have a `cleanup` function to do it. Wdut?

actions/install/installer.sh
```diff
@@ -0,0 +1,89 @@
...
+alias log_info="echo \"INFO:\""
+alias log_error="echo \"ERROR:\""
+
+verifier_release="$1"
+install_dir="$HOME/.slsa/bin/${verifier_release}"
+
+mkdir -p "$install_dir"
+cd "$install_dir"
+
+if [[ "$verifier_release" == "main" ]]; then
```
laurentsimon, line 15: @ianlewis is it standard to support "main"?

actions/install/action.yml
```diff
@@ -0,0 +1,19 @@
+name: setup
+description: 'Installs SLSA verifier and adds it to your PATH'
+branding:
+  icon: 'package'
+  color: 'blue'
+inputs:
+  verifier-release:
+    description: 'slsa-verifier release version to install'
+    required: false
+    default: 'v1.3.0'
```
laurentsimon, line 10: What do you think of making this option `required: true`?
- We don't need to maintain / update the default on every release. 
- Users will declare the version in their workflow, which is easier for readability / troubleshooting for them
- We'd need to be very careful when updating MAJOR semver, because if we forget to update users may be updating the action to a more recent MAJOR version but get a lower version installed. (Although, this is something we could have a pre-release check)

One drawback I see is that users need to update manually. A compromise would be to support versions like `v1` and we can resolve that to the most recent version, but this sort of violate the principle of pinning :/

What does everyone think?

@ianlewis @asraa

actions/install/action.yml
```diff
@@ -0,0 +1,19 @@
+name: setup
+description: 'Installs SLSA verifier and adds it to your PATH'
+branding:
+  icon: 'package'
+  color: 'blue'
+inputs:
+  verifier-release:
```
laurentsimon, line 7: shall we call this `verifier-version`? GitHub actions use `python-version` for Python and `node-version` for node, so this may be a more standard naming convention...?

.github/workflows/test_install.yml
laurentsimon: Maybe you could use a verification job even when this job fails. See some example in https://github.com/slsa-framework/example-package/blob/main/.github/workflows/e2e.generic.push.main.default.slsa3.yml#L92-L106

actions/installer.yml
laurentsimon: > Another issue is that we may have to fix bugs in the action after a version is released, but still allow users to download that version of the verifier. i.e. v2.0.0 has a bug in the installer but if we update to v2.0.1 then we have to release an identical version of the verifier as v2.0.1 as well which is a bit awkward.

I suppose in this model the verifier and actions would be considered "the same", and a bug in the action is considered a bug in the verifier and users should update. So we would release `v2.0.1` for a fix, and `v2.0.0` could be installed by a user but it would have the bug. Is there a compelling use case for allowing users to keep `v2.0.1`? If they want the v2 release, `v2.0.x` would be latest and always the one they should update to eventually... no? It would not break compatibility either since it's the same MAJOR.

actions/install/action.yml
laurentsimon: Maybe there's a solution I overlooked. If we keep `required: false` and build a pre-release step that verifies the new tag is the same as the default version. By default, the action ref and the default version would be in sync. 

If a user overwrites the version, it's clearly defines in their workflow.

The only limitation is that it's not pinned by hash, but I think this is a general limitation of allowing a version option in general.

actions/install/action.yml
```diff
@@ -0,0 +1,19 @@
...
+    default: 'v1.3.0'
+runs:
+  using: "composite"
+  steps:
+    - run: |
+        $GITHUB_ACTION_PATH/installer.sh $VERIFIER_RELEASE
+        echo "$HOME/.slsa/bin/$VERIFIER_RELEASE" >> $GITHUB_PATH
+      env:
+        VERIFIER_RELEASE: "${{ inputs.verifier-release }}"
+      shell: bash
```
ianlewis, line 19: The indentation here seems incorrect.

I would put these lines above the `run` as well and put `run` last.

actions/install/installer.sh
ianlewis: I'm not sure about all `setup-*` actions. I think the general convention is to support a "latest" special case and install the "latest" release but I don't think "main" is usually supported. From what I can tell [setup-go](https://github.com/actions/setup-go) only supports versions in a specific [manifest file](https://github.com/actions/go-versions/blob/main/versions-manifest.json) or you can omit the input and it installs the latest release. [setup-node](https://github.com/actions/setup-node) seems to support `latest`/`current`/`node` as a keyword to install the latest release. But neither support installing from `main` as far as I can tell.

In our case, I think we should not support it or, at the very least, strongly discourage it.

actions/install/installer.sh
```diff
@@ -0,0 +1,94 @@
...
+  rm slsa-verifier-linux-amd64
+  rm SHA256SUM.md
+  log_error "Failed to verify binary checksum. Did not install SLSA verifier."
+  exit 1
+fi
+mv slsa-verifier-linux-amd64 slsa-verifier-bootstrap
+chmod +x slsa-verifier-bootstrap
+
+case "${verifier_release}" in
+  "v0.0.1")
```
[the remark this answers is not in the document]
ianlewis: There seems to be a couple special cases for `v1.2.0` and `test-force-provenance-error`. `v1.2.0` because it's the same version used to bootstrap the install so it just uses the verifier binary used during bootstrap.

Though I think this could definitely be cleaned up a bit to only deal with those special cases.

actions/installer.yml
kpk47: Presumably you mean to ask if there's a compelling case for allowing clients to keep ```v2.0.0```, not ```v2.0.1```. I can't think of one. 

I don't have a strong opinion on whether we should version the installer with the binary or separately. I can see pros and cons to either approach. However, if we're going to version them together, do we want the installer to be compatible with older binary versions? If so, do we intend to patch older minor releases to add the installer? That feels like a pretty cumbersome release process.

laurentsimon: > If so, do we intend to patch older minor releases to add the installer? That feels like a pretty cumbersome release process.

Wdum? If we version them together, each GHA version would only support the same binary version. We would only support versions >= first release of the GHA.

Let me know if I missed something.

actions/install/README.md
```diff
@@ -0,0 +1,29 @@
...
+## Usage
+
+This action requires using GitHub-provided Linux runners.
+
+Add the following snippet to your Github workflow YAML file:
+
+```yaml
+uses: slsa-framework/slsa-verifier-installer@main
+with:
+    verifier-release: v1.3.0 # optional
```
asraa, line 18: we always have to expect vX.Y.Z here right? no resolution of vX to latest or vX.Y to latest patch

actions/install/installer.sh
asraa: I think we should disallow support: we often may change API/change semantics at main and need to notify users at a next major release if there's breaking changes. allowing main would mean that users would pick up on breaking changes without notification

actions/install/installer.sh
asraa: Yes! or all work in a tmp dir, with only a final success moved

actions/installer.yml
asraa: Hmmm I went back and forth on my opinions on this:

On coupling:
* Good for end user update story -- dependabot/etc will be able to update verifier versions
* I don't think we would support old verifiers if we went this route. 

On decoupling:
* I see many examples of actions that decouple: `golangci-lint-action`, `cosign-installer`, `setup-go`. So there's already a precendent.
* Clients miss out on updates because dependabot won't update args. If there's a bug fix/patch release in the `verifier-version` then how will clients receive the update? Right now the code isn't so stable and I expect this to be moving and important.
* Maybe clients are encouraged to use `latest` release unless they have a good reason not to, anyway.
* This would require a separate repo no? To manage separate versioning. Is there any blockers to doing so?

PR #245, opened 2022-09-01
feat: add CLI tests for GCB verification

laurentsimon:


cli/slsa-verifier/main.go
```diff
@@ -178,6 +179,32 @@ func getArtifactHash(artifactImage, artifactPath string) (string, error) {
...
+	// Retrieve the image digest.
+	digest, err := container.GetImageDigest(artifactImage)
+	if err != nil {
+		return "", fmt.Errorf("%w: %s", serrors.ErrorImageHash, err)
+	}
+
+	// Check if the image refers to a file.
+	// If it does, we don't expect users to provide an 'imutable'
+	// reference with `@sha256:xxx`.
+	_, err = os.Stat(artifactImage)
```
laurentsimon, line 191: local verification will fail afterwards, but this is needed for CLI tests, IIUC

cli/slsa-verifier/main_test.go
```diff
@@ -664,3 +665,192 @@ func Test_runVerifyArtifactImage(t *testing.T) {
...
+			pbuilderID: &builder,
+			err:        serrors.ErrorMismatchHash,
+		},
+	}
+	for _, tt := range tests {
+		tt := tt // Re-initializing variable so it is not changed while executing the closure below
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			checkVersions := getBuildersAndVersions(t, "", nil, GCB_ARTIFACT_IMAGE_BUILDERS)
```
laurentsimon, line 825: Note: will need some re-factoring when adding support for v0.3

cli/slsa-verifier/main.go
```diff
@@ -178,6 +179,32 @@ func getArtifactHash(artifactImage, artifactPath string) (string, error) {
...
 	}
-	// Retrieve image digest
-	return container.GetImageDigest(artifactImage)
+	// Retrieve the image digest.
+	digest, err := container.GetImageDigest(artifactImage)
+	if err != nil {
+		return "", fmt.Errorf("%w: %s", serrors.ErrorImageHash, err)
+	}
+
+	// Check if the image refers to a file.
```
asraa, line 188: Right now local files are only supported in test anyway right? Not opposed here though to including

cli/slsa-verifier/main.go
```diff
@@ -178,6 +179,32 @@ func getArtifactHash(artifactImage, artifactPath string) (string, error) {
...
+		return digest, nil
+	}
+
+	// For image in a registry, only allow immutable images.
+	ref, err := crname.ParseReference(artifactImage)
+	if err != nil {
+		return "", fmt.Errorf("crane.ParseReference(): %w", err)
+	}
+
+	if !strings.HasPrefix(ref.Identifier(), "sha256:") {
```
asraa, line 205: Could we move this into the container package to consolidate? Maybe just add a validation when artifactImage is passed in like `container.ValidateArtifactReference`?

cli/slsa-verifier/main.go
laurentsimon: correct. I need this code to allow tests to pass.

cli/slsa-verifier/main.go
laurentsimon: god call. Done, thanks

asraa (approved):

laurentsimon:
Fyi, I introduced a function parameter to change the default function that computes the digest. The global did not work with all the test that set it up concurrently.

PR #246, opened 2022-09-02
Add a GitHub Action for installing slsa-verifier.

kpk47:
This is a rewrite of https://github.com/slsa-framework/slsa-verifier/pull/233 in Typescript.

@asraa @ianlewis @laurentsimon


laurentsimon (commented):
looks good. Left a few comment. I think this is the "allow installation of any version". Heard back from Ian whether he's OK with using the `@v1.2.3` opption instead?

actions/installer/src/index.ts
```diff
@@ -0,0 +1,142 @@
+import * as core from "@actions/core";
+import * as exec from "@actions/exec";
+import * as io from "@actions/io";
+import * as tc from "@actions/tool-cache";
+import * as crypto from "crypto";
+import * as fs from "fs";
+import * as os from "os";
+
+const LATEST_VERSION = "v1.3.0";
```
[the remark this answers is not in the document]
kpk47: That part of the GH context is not available to Javascript actions. This is the subset that's available, and it does not include the action path or action ref: https://github.com/actions/toolkit/blob/main/packages/github/src/context.ts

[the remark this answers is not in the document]
laurentsimon: You can access everything. 3 Options:
- Either you read the `GITHUB_EVENT_PATH` and parse it yourself. I never remember if this is just the payload or the context
- If the context really is not available, create a default action input and populate it by default to the GH context. See 
https://github.com/ossf/scorecard-action/blob/main/action.yaml#L45-L48, you could populate it with something like https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/builder_go_slsa3.yml#L257
- Simplest option for our use case, is to read the env variable `GITHUB_ACTION_PATH` https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables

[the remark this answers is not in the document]
laurentsimon: re: breaking workflow. You can list releases and install the one that corresponds to the commit. The main question, I think, is whether we want to allow installing any version, or simply install the on the user pins the Action to.

kpk47:
@laurentsimon I spoke with Ian about pinning by release tags, and he's fine with it. Unfortunately, I've written a bunch of versions of this action and decided that I like the version with a `version` input and an explicit `LATEST_VERSION` in the javascript. Here's my reasoning. LMK if you disagree.

Pros
- Users can still pin to a release tag if they want. The `version` input is optional, and pinning to a release tag will install whichever version was latest at the time of release (i.e. `@1.2.3` will install `v1.2.3`).
- If we have an explicit `LATEST_VERSION` in the code, users can pin to a commit hash. GH's documentation on using public actions recommends pinning to a commit SHA rather than a release version or `main`, so I think it's important we support that use case.
- It's easier to write the action if it takes the version as input. `action_path` and `action_ref` are not part of the subset of the GH context that's available to Javascript actions, so getting the version that way requires an ugly hack with a second (composite) action. I wrote that version and discarded it because my intuition is that it would be brittle.
- It's easier to test the action if it takes the version as an input since we can use a matrix to cover all supported release versions. We cannot parse the version from a local action (it's always `main` at whatever commit you have checked out), so we'd need to use a second repo for running tests. I don't think it's worth the hassle.
- After looking through the GHA Marketplace, I realized that it would be very unusual if we only allowed choosing the installed version by pinning the action to a release tag. Every other `*-setup` action I found either supported taking the version as an input or only supported installing the latest release. I think that being unusual in this way would be helpful for adoption.

Cons
- We have to keep the `LATEST_VERSION` constant up to date, which most likely means adding an extra step to our release process.

laurentsimon:
> @laurentsimon I spoke with Ian about pinning by release tags, and he's fine with it. Unfortunately, I've written a bunch of versions of this action and decided that I like the version with a `version` input and an explicit `LATEST_VERSION` in the javascript. Here's my reasoning. LMK if you disagree.
> 
> Pros
> 
> * Users can still pin to a release tag if they want. The `version` input is optional, and pinning to a release tag will install whichever version was latest at the time of release (i.e. `@1.2.3` will install `v1.2.3`).
> * If we have an explicit `LATEST_VERSION` in the code, users can pin to a commit hash. GH's documentation on using public actions recommends pinning to a commit SHA rather than a release version or `main`, so I think it's important we support that use case.

We can support this too.

> * It's easier to write the action if it takes the version as input. `action_path` and `action_ref` are not part of the subset of the GH context that's available to Javascript actions, so getting the version that way requires an ugly hack with a second (composite) action. I wrote that version and discarded it because my intuition is that it would be brittle.

`action_path` is accessible, and it contains the `ref`

> * It's easier to test the action if it takes the version as an input since we can use a matrix to cover all supported release versions. We cannot parse the version from a local action (it's always `main` at whatever commit you have checked out), so we'd need to use a second repo for running tests. I don't think it's worth the hassle.

We can use an API to list the version, and dynamically generate the matrix, as in https://michaelheap.com/dynamic-matrix-generation-github-actions/.

> * After looking through the GHA Marketplace, I realized that it would be very unusual if we only allowed choosing the installed version by pinning the action to a release tag. Every other `*-setup` action I found either supported taking the version as an input or only supported installing the latest release. I think that being unusual in this way would be helpful for adoption.

That's the main argument against.

> 
> Cons
> 
> * We have to keep the `LATEST_VERSION` constant up to date, which most likely means adding an extra step to our release process.

Let's enjoy the week-end. I know re-writing so many times is frustrating. Let's sync next week

.github/workflows/test_install.yml
```diff
@@ -0,0 +1,55 @@
...
+  push:
+  schedule:
+    - cron: '14 3 * * *'
+
+jobs:
+  installer_test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        verifier-version: [v0.0.1, v1.0.0, v1.0.1, v1.0.2, v1.0.3, v1.1.0, v1.1.1, v1.2.0, v1.3.0, latest]
```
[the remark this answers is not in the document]
kpk47: Done

actions/installer/dst/index.js
```diff
@@ -0,0 +1,129 @@
+"use strict";
```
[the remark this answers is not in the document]
kpk47: Done

.github/workflows/test_install.yml
```diff
@@ -0,0 +1,20 @@
+permissions: read-all
```
laurentsimon, line 1: Can you follow the file naming convention we use on the generator https://github.com/slsa-framework/slsa-github-generator/tree/main/.github/workflows?
Basically have pre-submit.actions.yml for Action presubmit, etc.

We only had a pre-submit.yml in this repo, but looks like it's time to change.

laurentsimon (approved):
Thanks!
Just added a nit to rename files for pre-submit, and we're good to merge

actions/installer/README.md
```diff
@@ -0,0 +1,37 @@
...
+```
+
+To install a specific version, use:
+```yaml
+uses: slsa-framework/slsa-verifier-installer
+  with:
+    verifier-version: v1.3.0
+```
+
+NB: Pinning this action to a relase tag or commit SHA will pin you to the latest version of `slsa-verifier` at the time of release.
```
ianlewis, line 24:
```suggestion
NB: Pinning this action to a release tag or commit SHA will pin you to the latest version of `slsa-verifier` at the time of release.
```

BTW: what does NB mean? consider expanding it if it's an acronym.

actions/installer/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+### Optional Inputs
+
+| Input | Description |
+| --- | --- |
+| `verifier-release` | `slsa-verifier` version to use. Default value is `latest`. |
```
ianlewis, line 30: nit: run the markdown file through prettier before committing.

actions/installer/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+See https://github.com/slsa-framework/slsa-verifier/releases for the list of available `slsa-verifier` releases.
+
+For a full example workflow, see `.github/workflows/test_installer.yml`.
```
ianlewis, line 34: nit: consider making this a link? In order to be a relative URL I guess it would need to be with `../../.github/workflows/test_installer.yml`

actions/installer/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+This action requires using GitHub-provided Linux runners.
+
```
ianlewis, line 37: nit: extra line at the end.

actions/installer/spec/index.test.ts
```diff
@@ -0,0 +1,56 @@
+const exec = require("@actions/exec");
```
ianlewis, line 1: nit: we haven't been good about this but add the copyright header with 'SLSA Authors' as the author.

See the header here for an example: https://github.com/slsa-framework/slsa-github-generator/blob/main/slsa/buildtype.go

same for other code files.

actions/installer/src/index.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), "slsa-verifier_"));
+  const bootstrapDir = `${tmpDir}/bootstrap`;
+  const installDir = `${tmpDir}/${version}`;
+
+  let bootstrapVerifierPath;
+  try {
+    // Download bootstrap version and validate SHA256 checksum
+    bootstrapVerifierPath = await tc.downloadTool(
+      "https://github.com/slsa-framework/slsa-verifier/releases/download/v1.3.0/slsa-verifier-linux-amd64",
```
ianlewis, line 64: nit: use LATEST_VERSION in the url here. Or create another global that sets the bootstrap URL or bootstrap version.

actions/installer/src/index.ts
```diff
@@ -0,0 +1,141 @@
...
+import * as io from "@actions/io";
+import * as tc from "@actions/tool-cache";
+import * as crypto from "crypto";
+import * as fs from "fs";
+import * as os from "os";
+import * as path from "path";
+
+const LATEST_VERSION = "v1.3.0";
+
+const VERIFIER_V130_SHA256 =
```
ianlewis, line 12: nit: make the name indicate it's used for bootstrapping. Set another global here to specify the bootstrap version or URL (see other comment).
```suggestion
const BOOTSTRAP_VERIFIER_SHA256 =
```

actions/installer/src/index.ts
```diff
@@ -0,0 +1,141 @@
...
+    );
+    if (exitCode !== 0) {
+      throw new Error(
+        `Unable to verify binary provenance. Aborting installation. stdout: ${stdout}; stderr: ${stderr}`
+      );
+    }
+  } catch (error: unknown) {
+    const errMsg = error instanceof Error ? error.message : String(error);
+    core.setFailed(`Error executing slsa-verifier: ${errMsg}`);
+    return;
```
ianlewis, line 125: nit: is it maybe prudent to call `cleanup` when errors occur as well?

ianlewis (approved):
Looks good!

actions/installer/README.md
```diff
@@ -0,0 +1,37 @@
...
+To install the latest version, use:
+
+```yaml
+uses: slsa-framework/slsa-verifier-installer
+```
+
+To install a specific version, use:
+```yaml
+uses: slsa-framework/slsa-verifier-installer
```
laurentsimon, line 19: need to update the doc, since the verifier-version is no longer supported?

laurentsimon (commented):
I thought we'd use the `GITHUB_ACTION_PATH` to determine the version to download?

actions/installer/README.md
kpk47: It's Latin, so expanding the acronym would be kind of weird (https://en.wikipedia.org/wiki/Nota_bene). I've just replaced it with "Note: ..."

kpk47:
@laurentsimon ```GITHUB_ACTION_PATH``` is only available to composite actions (https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables).

laurentsimon:
> @laurentsimon `GITHUB_ACTION_PATH` is only available to composite actions (https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables).

ouch, all these strange corner cases. Good find!
One hack is to declare a composite action and have a single step that calls `node code.js`. Does this work?

Note: GH are working on better APIs for Actions / re-usable workflows to determine their identity / ref, so later we should be able to use it.

actions/installer/action.yml
```diff
@@ -18,5 +18,28 @@ branding:
...
+      - name: Setup Node.js 16
+        uses: actions/setup-node@2fddd8803e2f5c9604345a0b591c3020ee971a93 # tag=v3.4.1
+        with:
+          node-version: 16
+
+      - name: Install dependencies
+        working-directory: actions/installer
+        run: npm ci
+
+      - name: Run build and tests
```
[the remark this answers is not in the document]
kpk47: Done

kpk47:
Okay, I think I've address all of your comments. PTAL and then I think we're ready to merge.

laurentsimon:
Please re-base

actions/installer/README.md
```diff
@@ -0,0 +1,35 @@
...
+This action installs the SLSA verifier and adds it to your PATH.
+
+For more information about `slsa-verifier`, refer to [its documentation](https://github.com/slsa-framework/slsa-verifier#verification-of-provenance).
+
+For more information about SLSA in general, see [https://slsa.dev](https://slsa.dev).
+
+## Usage
+
+To install the latest version, use:
+
```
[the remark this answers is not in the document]
kpk47: Done

actions/installer/src/index.ts
```diff
@@ -0,0 +1,202 @@
...
+}
+
+// Resolve command line argument to a version number
+export async function getVerifierVersion(actionRef: string): Promise<string> {
+  if (validVersion(actionRef)) {
+    return actionRef;
+  }
+
+  // If actionRef is a commit SHA, then find the associated version number.
+  const shaRe = /^[a-f\d]{40}$/;
```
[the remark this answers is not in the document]
kpk47: Thanks! I copied it from you (https://github.com/slsa-framework/slsa-github-generator/pull/86). 😅

actions/installer/action.yml
```diff
@@ -0,0 +1,48 @@
...
+      uses: actions/setup-node@2fddd8803e2f5c9604345a0b591c3020ee971a93 # tag=v3.4.1
+      with:
+        node-version: 16
+
+    - name: Install dependencies
+      working-directory: actions/installer
+      shell: bash
+      run: npm ci
+
+    - name: Run build and tests
```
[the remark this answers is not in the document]
kpk47: Done.

.github/workflows/pre-submit.actions.yml
```diff
@@ -0,0 +1,43 @@
+name: Actions pre submits
+on: [pull_request, workflow_dispatch]
+
+permissions: read-all
```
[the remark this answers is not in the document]
kpk47: Ack. I'll stash that commit before I rebase so we don't have to rewrite it.

actions/installer/README.md
```diff
@@ -0,0 +1,29 @@
...
+## Usage
+
+To install a specific version of `slsa-verifier`, use:
+
+```yaml
+uses: slsa-framework/slsa-verifier-installer@v1.3.0
+```
+
+Note: Pinning this action to a release tag or commit SHA will pin you to the latest version of `slsa-verifier` at the time of release.
```
laurentsimon, line 17: this sentence is no longer true.

actions/installer/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+### Optional Inputs
```
laurentsimon, line 19: there's no option, so lets remove this as well.

ianlewis (approved):

ianlewis:
LGTM, pending doc changes that @laurentsimon commented on, and merging the conflicting `pre-submit.yml` changes.

actions/installer/README.md
kpk47: Done

actions/installer/README.md
kpk47: Done

kpk47:
I think I've fixed the merge conflicts, but I can't get GitHub to recognize it. I dropped the changes in `pre-submit.yml` in favor of a straight rename, so I'm not sure what else I can do.

laurentsimon:
yeah things get weird when re-basing and files that are deleted have some updates. Have you tried `git rm` the `pre-submit.yml` to see if it helps?

kpk47:
Okay, I think I got it. For posterity, the fix was re-adding `pre-submit.yml`, merging in changes from `main`, removing `pre-submit.cli.yml`, and then re-doing the name change from `pre-submit.yml` to `pre-submit.cli.yml`.

actions/installer/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+To install a specific version of `slsa-verifier`, use:
+
+```yaml
+uses: slsa-framework/slsa-verifier-installer@v1.3.0
+```
+
+See https://github.com/slsa-framework/slsa-verifier/releases for the list of available `slsa-verifier` releases.
+
+For a full example workflow, see [../../.github/workflows/test_installer.yml](https://github.com/slsa-framework/slsa-verifier/.github/workflows/test_installer.yml).
```
laurentsimon, line 19: I think think this file exists anymore, does it?

PR #256, opened 2022-09-12
feat: support builderID matching with or without semver for GCB

laurentsimon:


verifiers/utils/builder.go
```diff
@@ -0,0 +1,22 @@
+package utils
```
asraa, line 1: WDYT about adding a function that compares builder IDs? otherwise you have logic for comparing repeated across tests and in code. It would capture the logic that if `expected` contains the version, then it compares that too.

laurentsimon: Good idea. I've been thinking about doing it as well, but was not sure. Let me add it.

verifiers/internal/gcb/verifier.go
```diff
@@ -27,7 +27,9 @@ func GCBVerifierNew() *GCBVerifier {
 // generated by the builderID.
 func (v *GCBVerifier) IsAuthoritativeFor(builderID string) bool {
 	// This verifier only supports the GCB builders.
-	return strings.HasPrefix(builderID, "https://cloudbuild.googleapis.com/GoogleHostedWorker@")
+	// Note: `ParseBuilderID` never fails with `needVersion=false`.
```
asraa, line 30: won't it fail for something like `https://cloudbuild.googleapis.com/GoogleHostedWorker@v2@v3`?

May want to change `strings.Split` to `strings.SplitN` to avoid, but you have a test case that adequately handles this input as an error.

asraa (changes requested):

verifiers/utils/builder.go
laurentsimon: done. PTAL

verifiers/internal/gcb/verifier.go
laurentsimon: mistake from me. You're 100% right. PTAL.

verifiers/internal/gha/verifier.go
```diff
@@ -43,68 +44,79 @@ func verifyEnvAndCert(env *dsse.Envelope,
...
 	if err != nil {
-		return nil, "", err
+		return nil, nil, err
 	}
 
-	return verifyEnvAndCert(env, cert,
+	content, builderID, err := verifyEnvAndCert(env, cert,
 		provenanceOpts, builderOpts,
 		defaultArtifactTrustedReusableWorkflows)
+	if err != nil {
```
asraa, line 103: nit: why did this need to change from returning those results? (I guess we have trust in github not providing malformed job_workflow_refs so you may not even be able to get a verified but malformed builder, but the assertion maybe should be there)

verifiers/internal/gha/verifier.go
```diff
@@ -43,68 +44,79 @@ func verifyEnvAndCert(env *dsse.Envelope,
...
 		workflowInfo.CallerHash)
 	// Return verified provenance.
 	r, err := base64.StdEncoding.DecodeString(env.Payload)
-	return r, builderID, err
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// Temporary code.
+	bid, _ := utils.BuilderIDNew(builderID)
```
asraa, line 79: if someone supplies their own malformed builder ID in here, this could fail right?

verifiers/utils/builder_test.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+
+	serrors "github.com/slsa-framework/slsa-verifier/errors"
+)
+
+func Test_ParseBuilderID(t *testing.T) {
```
asraa, line 12: could you put in 1-2 tests on matches? just for match with/without version

asraa (approved):

verifiers/internal/gha/verifier.go
laurentsimon: I dont think so, because GHA only matches on the builder name, not the version. This is a temporary hack: in next PR I will add support for GHA

verifiers/internal/gha/verifier.go
laurentsimon: not actually needed. A residual change I should have removed. Reverted. Good catch.

verifiers/utils/builder_test.go
laurentsimon: yes, those are in there now, including for `BuilderIDNew()`. PTAL

verifiers/internal/gha/verifier.go
laurentsimon: actually I removed the error checking, otherwise it always fails since the `builderID` never contains the `@v1.2.3` on GHA. Adding support in the next PR.

PR #257, opened 2022-09-13
feat: support builderID matching with or without semver for GHA

laurentsimon:


cli/slsa-verifier/main_test.go
```diff
@@ -513,7 +513,7 @@ func Test_runVerifyArtifactPath(t *testing.T) {
```
laurentsimon, line 513: Note: CLI tests on the way

cli/slsa-verifier/main_test.go
```diff
@@ -491,34 +493,82 @@ func Test_runVerifyArtifactPath(t *testing.T) {
...
-					ProvenancePath:      provenancePath,
-					SourceURI:           tt.source,
-					SourceBranch:        tt.pbranch,
-					BuilderID:           tt.pbuilderID,
-					SourceTag:           tt.ptag,
-					SourceVersionTag:    tt.pversiontag,
-					BuildWorkflowInputs: tt.inputs,
+				// TODO(#258): invalid builder ref.
+				semver := path.Base(v)
+				// For each test, we run 4 sub-tests:
```
asraa, line 498: For tests whos error expects a build mismatch, this won't (correctly) fail for cases 1-3. 

Maybe you should switch on whether pbuilderID was specified? In those cases we are explicitly testing pBuilderID

laurentsimon: IIUC, only the tests for v0.0.2 use the `@main` builder, the one >= v1 seem to all use a proper `@vx.y.z`, so the build mismatch tests seem to be working. I'll double check.

verifiers/internal/gha/verifier.go
```diff
@@ -59,9 +60,11 @@ func verifyEnvAndCert(env *dsse.Envelope,
 		return nil, nil, err
 	}
 
+	fmt.Println("builderID:", builderID.String())
```
asraa, line 63: nit: remove before merging

verifiers/internal/gha/verifier.go
```diff
@@ -147,13 +144,22 @@ func (v *GHAVerifier) VerifyImage(ctx context.Context,
...
-		if verifyErr == nil {
+		if err == nil {
 			return verifiedProvenance, builderID, nil
 		}
+		errs = append(errs, err)
 	}
 
-	return nil, nil, fmt.Errorf("no valid attestations found on OCI registry: %w", verifyErr)
+	// Return the first error.
+	if len(errs) > 0 {
```
asraa, line 157: Why do we need this check? If we reach here this means we did not return any provenance, so there must be `errs`?

verifiers/utils/builder_test.go
```diff
@@ -201,6 +202,85 @@ func Test_Matches(t *testing.T) {
...
+			name:      "match long tags",
+			builderID: "some/name@refs/tags/v1.2.3",
+			match:     "some/name@refs/tags/v1.2.3",
+			allowRef:  true,
+		},
+		{
+			name:      "mismatch tag length",
+			builderID: "some/name@refs/tags/v1.2.3",
+			match:     "some/name@v1.2.3",
+			err:       serrors.ErrorMismatchBuilderID,
```
asraa, line 276: sorry if I'm a little lost here, but does a user provide `match` or `builderID`? I thought they provide `match`. Shouldn't it allow user to pass with the short name to provide?

verifiers/internal/gha/builder.go
```diff
@@ -33,63 +34,81 @@ var defaultContainerTrustedReusableWorkflows = map[string]bool{
...
+	// refs/heads/main for e2e tests. See verifyTrustedBuilderRef().
 	// No builder ID provided by user: use the default trusted workflows.
-	if builderID == nil || *builderID == "" {
-		if _, ok := defaultBuilders[path]; !ok {
-			return "", fmt.Errorf("%w: %s got %t", serrors.ErrorUntrustedReusableWorkflow, path, builderID == nil)
+	if expectedBuilderID == nil || *expectedBuilderID == "" {
+		if _, ok := defaultBuilders[certPath]; !ok {
+			return nil, fmt.Errorf("%w: %s got %t", serrors.ErrorUntrustedReusableWorkflow, certPath, expectedBuilderID == nil)
+		}
+		// Construct the builderID using the default builder's name and the certificate's tag.
```
asraa, line 90: this is not the default right? `certPath` is from the signing certiicate

verifiers/internal/gha/provenance.go
```diff
@@ -42,22 +42,14 @@ func provenanceFromEnv(env *dsselib.Envelope) (prov *intoto.ProvenanceStatement,
...
-	// Note: builderID does not contain the tag.
-	// TODO(#189): support cases where user wants to match on the full builderID, including the tag.
-	bid, err := sourceFromURI(builderID, true)
-	if err != nil {
-		return err
-	}
-	if provid != bid {
+// This function does an exact comparison, and expects certBuilderID to be the full
+// `name@refs/tags/<name>`.
+func verifyBuilderID(prov *intoto.ProvenanceStatement, certBuilderID string) error {
```
asraa, line 47: could you change the name to something like `verifyBuilderIDExactMatch`?

asraa (commented):
sorry for how long that took! it took me a while to sort things out.

verifiers/internal/gha/verifier.go
laurentsimon: good catch. Completely missed this one. Thanks!

verifiers/internal/gha/verifier.go
laurentsimon: if there is no attestation, the loop will not enter errs will be nil. I don't know the exact behavior `RunCosignImageVerification`. If it fails when no attestations are found, I agree with you. Shall I keep the code or add a comment that atts is never empty?

verifiers/utils/builder_test.go
laurentsimon: user provides `match`. BuilderID is the one in the provenance. Let me make this clearer in the doc

verifiers/internal/gha/builder.go
laurentsimon: +1, change made

verifiers/internal/gha/provenance.go
laurentsimon: +1, done.

verifiers/utils/builder_test.go
laurentsimon: > Shouldn't it allow user to pass with the short name to provide?

You're correct, but only if `allowRef: true`, which is not set here. The `allowRef` is to ensure we nly enable this feature for GHA builders. The GCB builder just use `name@version`

laurentsimon: I renamed to TrustedBuilderID, to make to clear this is the trusted one from the cert / provenance.

verifiers/internal/gha/verifier.go
asraa: It is correct that cosign will fail when no attestations are there, but I think we should not rely on that. I overlooked that missing attestations would not enter the loop. Thank you!

asraa (approved):

verifiers/utils/builder_test.go
laurentsimon: np!

PR #266, opened 2022-09-22
doc: document build id and GCB vs GHA

laurentsimon:


README.md
```diff
@@ -75,29 +85,61 @@ Flags:
...
-### Example
+### Option details
+
+The following options are supported for [SLSA GitHub builders and generators](https://github.com/slsa-framework/slsa-github-generator#generation-of-provenance):
+
+| Option | Description |
+| --- | ----------- |
+| `source-uri` | Expects a source, for e.g. `github.com/org/repo`. |
+| `source-branch` | Expects a `branch` like `main` or `dev`. Not supported for all GitHub Workflow triggers. |
+| `source-tag` | Expects a  `tag` like `v0.0.1`. Verifies exact tag used to create the binary. NSupported for new [tag](https://github.com/slsa-framework/example-package/blob/main/.github/workflows/e2e.go.tag.main.config-ldflags-assets-tag.slsa3.yml#L5) and [release](https://github.com/slsa-framework/example-package/blob/main/.github/workflows/e2e.go.release.main.config-ldflags-assets-tag.slsa3.yml) triggers. |
```
asraa, line 96: nit: /NSupported/Supported
new tag?

README.md
```diff
@@ -75,29 +85,61 @@ Flags:
...
-| `source-tag` | Expects a  `tag` like `v0.0.1`. Verifies exact tag used to create the binary. NSupported for new [tag](https://github.com/slsa-framework/example-package/blob/main/.github/workflows/e2e.go.tag.main.config-ldflags-assets-tag.slsa3.yml#L5) and [release](https://github.com/slsa-framework/example-package/blob/main/.github/workflows/e2e.go.release.main.config-ldflags-assets-tag.slsa3.yml) triggers. |
-| `source-versioned-tag` | Like `tag`, but verifies using semantic versioning. |
-| `build-workflow-input` | Expects key-value pairs like `key=value` to match against [inputs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputs) for GitHub Actions `workflow_dispatch` triggers. |
+## Verification for Google Cloud Build
+
+### Artifacts
+This is WIP and currently not supported.
 
+### Containers
+To verify a contaimer image, run the following command:
```
asraa, line 128: maybe mention very briefly that slsa-verifier requires input by digest because of TOCTOU? (to justify the second line)

README.md
```diff
@@ -75,29 +85,61 @@ Flags:
...
+
+```bash
+$ IMAGE=us-west2-docker.pkg.dev/gosst-scare-sandbox/quickstart-docker-repo/quickstart-image:v39
```
asraa, line 131: for the sake of an example, can we use one that has public perms?
i tried it without any auth, and right now I have permission denied

README.md
```diff
@@ -75,29 +85,61 @@ Flags:
...
+$ IMAGE=$(docker inspect --format='{{.RepoDigests}}' ${IMAGE} | cut -f1 -d ' ' | cut -d "[" -f2 | cut -d "]" -f1)
```
asraa, line 132: i'll try this when the image is public but to avoid docker pull before docker inspect, maybe we can use `crane digest`?

README.md
laurentsimon: Let me actually add a column about which builders support what.

README.md
laurentsimon: done. PTAL

README.md
laurentsimon: great point. Done.

README.md
laurentsimon: I'll provide the 2 options?
What you're saying is that we should not be pulling an image unless it's trusted, correct?

asraa: I don't think it's *needed*, although i think it is better to verify before pulling, but only because i think this second command won't work unless there's a docker pull before it

laurentsimon: done.

asraa (approved):

README.md
laurentsimon: agreed. I only added the crane command. I wanted to avoid asking users to install another tool. But I think it also avoids asking users to rm the image from their local cache if verification fails.

PR #276, opened 2022-09-27
Update README.md

laurentsimon:
Signed-off-by: laurentsimon <64505099+laurentsimon@users.noreply.github.com>

gcloud artifact registry does not support `image:tag@sha256:xxx`, so ask users to fetch the provenance with the image tag only


asraa (approved):
Thanks! I will also check this in the draft blog

PR #277, opened 2022-09-29
fix: fix rekor verification

asraa:

This fixes rekor verification by:

1. Removes support for v0.0.2 builders and v1.0.0 builders. At v1.1.1 and all other major versions, we support embedding the cert in the DSSE envelope, which allows us to make verifiable Rekor lookups rather than rely on a key-value index store.
2. Consolidate logic in `VerifyProvenanceSignature`: verifies the provenance in a step-by-step manner: covering signature validation, rekor lookup, and signature time validation (against entry upload time)
3. Removes dead code that was needed for the key-value index store lookups.
4. Adds testing for retrieving a rekor entry.
5. Uses Rekor's libraries for verifying a TLOG entry to avoid issues with Rekor API changes.


asraa:
This change will need backporting to all major releases! TODO tomorrow.

laurentsimon (approved):
Thanks!

asraa:
Thank you! I realized in doing the backported fixes that I can make a much less invasive change because that helped me pinpoint the exact issue; I will make the full fixes here (relying on rekor libraries) after the backports just FYI

PR #278, opened 2022-09-30
fix(backport/v1.1): make client shard aware when verifying entries on inactive shards

asraa:

Backport shard-aware fix to v1.1: allows searching through inactive shard's STHs. when verifying entries consistency with an STH.


asraa:
this successfully verifies v1.2.0 builder!

laurentsimon (approved):

PR #286, opened 2022-10-02
fix(backport/1.0): make client shard aware when verifying entries on inactive shards

asraa:


pkg/provenance.go
```diff
@@ -129,6 +132,13 @@ func verifyRootHash(ctx context.Context, rekorClient *client.Rekor, proof *model
 	if err := sth.UnmarshalText([]byte(*logInfo.SignedTreeHead)); err != nil {
 		return err
 	}
+	for _, inactiveShard := range logInfo.InactiveShards {
+		if *inactiveShard.TreeID == treeIDString {
+			if err := sth.UnmarshalText([]byte(*inactiveShard.SignedTreeHead)); err != nil {
+				return err
+			}
+		}
+	}
```
ianlewis, line 141: nit: I don't think we can fix this but this seems odd that we would have to build the signed checkpoint ourselves; including iterating over inactive shards. It feels like this should be a method on the `logInfo`.

ianlewis (approved):
I realize it's not completely new, but I think it's a bit unfortunate that rekor clients like slsa-verifier need to have logic for sharding. It feels like clients are being exposed to too many implementation details and that's making things brittle.

pkg/provenance.go
asraa: Big +1, I think it should be part of params. I'll open a follow-up issue

asraa:
> I realize it's not completely new, but I think it's a bit unfortunate that rekor clients like slsa-verifier need to have logic for sharding. It feels like clients are being exposed to too many implementation details and that's making things brittle.

I completely agree: sharding was meant to be internal client, but it wasn't: things are getting better (the return UUID sharding comparisons eventually can be removed) but seem like they won't be stable until after GA. It's been a pain to deal with this logic and debug.

I should be able to clean up this logic when GA does give us a stable client though. For e.g. we will no longer need to fetch the root hash and search shards for the correct one: a root hash will be returned with the entry.

I'm hesitant to make some of these changes though, because rekor keeps rolling back versions :|

pkg/provenance.go
asraa: https://github.com/sigstore/rekor/issues/1091

main_test.go
```diff
@@ -308,7 +308,7 @@ func Test_runVerify(t *testing.T) {
 			name:      "rekor upload bypassed",
 			artifact:  "binary-linux-amd64-no-tlog-upload",
 			source:    "github.com/slsa-framework/example-package",
-			err:       pkg.ErrorRekorSearch,
+			err:       pkg.ErrorNoValidRekorEntries,
```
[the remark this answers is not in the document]
asraa: Figured it out! It turns out it was a problem in the UUID lengths returned that was hairy:
old entries (like the ones in these tests) were added when the UUID lengths of rekor were 64 chars. that was causing verification errors now when trying to extract tree IDs from full UUIDs (now 80 chars) in order to perform the `verifyRootHash` that broke when rotating shards.

Now I use the full 80 char UUID that was returned from `GetLogEntryByUUID`

ianlewis:
> I should be able to clean up this logic when GA does give us a stable client though. For e.g. we will no longer need to fetch the root hash and search shards for the correct one: a root hash will be returned with the entry.

It's a bit unclear from the outside what is GA and what isn't with regard to Rekor and that translates to slsa-github-generator since we rely on the public Rekor.

> I'm hesitant to make some of these changes prematurely though, because rekor keeps rolling back versions :|

Completely understandable. We need to work with what we have currently.

PR #292, opened 2022-10-07
gcb: add gcb compatibility for provenance formats and buckets

asraa:

This PR does two things in order to achieve GCB provenance compatibility:
1. Adds support for incorrect string/int JSON for intoto. I temporarily vendor the GCB provenance and omit the `definedInMaterial` field. Until GCB updates their provenance generation, this will have to do.
2. Also adds support for GCS versioned buckets. I spoke with the GCB team: `v0.3` does not necessarily produce level 3 provenance, and we can only guarantee the provenance requirements (up to non-falsifiability for v0.3). We should definitely prioritize a versioning option in the future that would disallow generating provenance in a lower level mannger.

Adds testcase for the JSON marshalling issue and the GCS Buckets.

### DIFF
--- a/verifiers/internal/gcb/intoto.go
+++ b/verifiers/internal/gcb/intoto.go
@@ -1,0 +1,68 @@
+package gcb
+
+// Copy of github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v0.1
+// This holds an internal copy of in-toto-golang's structs for
+// SLSA predicates to handle GCB's incompatibility with the
+// published specification. Specifically, GCB provenance currently
+// produces a string for ProvenancePredicate.Recipe.DefinedInMaterial
+// rather than the compliant signed integer.
+
+import "time"
+
+const (
+	// PredicateSLSAProvenance represents a build provenance for an artifact.
+	PredicateSLSAProvenance = "https://slsa.dev/provenance/v0.1"
+)
+
+// ProvenancePredicate is the provenance predicate definition.
+type ProvenancePredicate struct {
+	Builder   ProvenanceBuilder    `json:"builder"`
+	Recipe    ProvenanceRecipe     `json:"recipe"`
+	Metadata  *ProvenanceMetadata  `json:"metadata,omitempty"`
+	Materials []ProvenanceMaterial `json:"materials,omitempty"`
+}
+
+// ProvenanceBuilder idenfifies the entity that executed the build steps.
+type ProvenanceBuilder struct {
+	ID string `json:"id"`
+}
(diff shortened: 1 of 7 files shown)


inferno-chromium (approved):

verifiers/internal/gcb/intoto.go
```diff
@@ -0,0 +1,68 @@
...
+type ProvenanceBuilder struct {
+	ID string `json:"id"`
+}
+
+// ProvenanceRecipe describes the actions performed by the builder.
+type ProvenanceRecipe struct {
+	Type string `json:"type"`
+	// DefinedInMaterial can be sent as the null pointer to indicate that
+	// the value is not present.
+	// DefinedInMaterial *int        `json:"definedInMaterial,omitempty"`
```
ianlewis, line 35: nit: add a short local comment as to why this is commented out. Maybe point to comment at top of file.

verifiers/internal/gcb/intoto.go
```diff
@@ -0,0 +1,68 @@
+package gcb
+
+// Copy of github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v0.1
```
ianlewis, line 3: nit: add a "NOTE: " prefix to the comment.

verifiers/internal/gcb/provenance.go
```diff
@@ -357,7 +356,9 @@ func (self *Provenance) VerifySourceURI(expectedSourceURI string, builderID util
 		return fmt.Errorf("%w: no materials", serrors.ErrorInvalidDssePayload)
 	}
 	uri := materials[0].URI
-	if !strings.HasPrefix(expectedSourceURI, "https://") {
+
+	// It is possible that GCS builds at level 2 use GCS sources, prefixed by gs://.
+	if strings.HasPrefix(uri, "https://") && !strings.HasPrefix(expectedSourceURI, "https://") {
```
ianlewis, line 361: nit: You might also leave a note as to why this is necessary in the first place. It's not necessarily obvious why "https://" might need to be prefixed at the beginning.

ianlewis: Also, consider checking for any URI scheme rather than "https://" specifically. So you don't end up with strange URIs like "https://http://example.com/..." if the user enters "--source http://example.com"

```suggestion
	if strings.HasPrefix(uri, "https://") && !strings.HasPrefix(expectedSourceURI, "://") {
```

ianlewis (commented):
A bit late but for posterity.

PR #300, opened 2022-10-12
fix: address gcb verifier comments and add gcb documentation

asraa:

Addressing comments from https://github.com/slsa-framework/slsa-verifier/pull/292/files/dcb49d5bc9d6bc0119b3a41d64c529eda79a112c

Note: I reverted the behavior and clarified expectations in the README.md:
* If the GCB provenance used a GCS bucket, we log a warning to Stderr to the user that the build was not configured with a GH trigger and therefore they cannot expect a version controlled source URI.
* We still match on the previous behavior of the `gs://[PROEJCT_ID]_cloudbuild/source/[BUCKET].tar.gz` to preserve behavior.


verifiers/internal/gcb/provenance.go
```diff
@@ -355,32 +356,56 @@ func (self *Provenance) VerifySourceURI(expectedSourceURI string, builderID util
...
+	if err != nil {
+		return fmt.Errorf("parsing materials URI: %w", err)
+	}
+
+	expectedURI, err := url.Parse(expectedSourceURI)
+	if err != nil {
+		return fmt.Errorf("parsing materials URI: %w", err)
+	}
+
+	// If users do not specify a scheme, use HTTPS by default.
```
asraa, line 370: this preserves the previous behavior

verifiers/internal/gcb/provenance.go
```diff
@@ -355,32 +356,56 @@ func (self *Provenance) VerifySourceURI(expectedSourceURI string, builderID util
...
-		// In v0.2, it uses format
-		// `https://github.com/laurentsimon/gcb-tests/commit/01ce393d04eb6df2a7b2b3e95d4126e687afb7ae`.
-		if !strings.HasPrefix(uri, expectedSourceURI+"/commit/") &&
-			!strings.HasPrefix(uri, expectedSourceURI+"#") {
-			return fmt.Errorf("%w: expected '%s', got '%s'",
-				serrors.ErrorMismatchSource, expectedSourceURI, uri)
+		// In v0.2, it uses the formats
+		// `https://github.com/laurentsimon/gcb-tests/commit/01ce393d04eb6df2a7b2b3e95d4126e687afb7ae`
+		// `gs://slsa-tooling_cloudbuild/source/1663616632.078353-fc7db143dcc64b5f9fe71d0497125ca1.tgz#1663616635134845`
+		if uri.Fragment == "" && !strings.HasPrefix(uri.Path, expectedURI.Path+"/commit/") {
```
asraa, line 395: it may be better to switch on scheme here.

verifiers/internal/gcb/provenance.go
```diff
@@ -355,32 +356,56 @@ func (self *Provenance) VerifySourceURI(expectedSourceURI string, builderID util
...
+			return fmt.Errorf("%w: expected matching paths '%s', got '%s'",
+				serrors.ErrorMismatchSource, expectedSourceURI, uri.String())
+		}
+		if uri.Fragment != "" && !strings.HasPrefix(uri.Path, expectedURI.Path) {
```
ianlewis, line 399: When comparing `gs://` URLs this check is maybe missing comparing the host/bucket name?

https://go.dev/play/p/hgDTpa5CX--

This applies to the other URL comparisons you are doing above and below as well.

ianlewis:
> > TBC, I'm not sure we need to support source checks for builds submitted as gcloud builds submit. I also wonder how v0.3 handles builds that are submitted this way.
> 
> The v0.3 builds use the same `gs://PROJECT_cloudbuild/source` style bucket, with a separate digest field with the sha256.
> 
> v0.3 doesn't mean SLSA 3 unfortunately, also.

Yeah. The reason I asked was because you don't have any special logic for `gs://` URLs in the `v0.3` part of the switch statement. You only have that logic under `v0.2`

laurentsimon:
> We still match on the previous behavior of the gs://[PROEJCT_ID]_cloudbuild/source/[BUCKET].tar.gz to preserve behavior.

will this include the trailing `#<id>`?

README.md
```diff
@@ -116,6 +116,8 @@ PASSED: Verified SLSA provenance
 
 The verified in-toto statement may be written to stdout with the `--print-provenance` flag to pipe into policy engines.
 
+Only GitHub URIs are supported with the `--source-uri` flag. A tag should not be specified, even if the provenance was built at some tag. If you intend to do source versioning validation, use `--print-provenance` and inspect the commit SHA of the config source or materials.
```
laurentsimon, line 119: shall we create an issue for `--source-sha1`? Probably nor urgent...

asraa:
> will this include the trailing #<id>?

The current behavior when I made it did not, so I kept as is (I had thought it was "parity" with the `/commit` matching). Since we're v2-releasing this I can change that too

laurentsimon (approved):
LGTM. I left a question i a comment about the trailing `#<id>` in the matching. I recall you said there was one sometimes. Will it be required for the match? I think yes, but wanted confirmaiton.

verifiers/internal/gcb/provenance.go
asraa: closing because this PR wont change logic

laurentsimon:
> > will this include the trailing #?
> 
> The current behavior when I made it did not: we match on everything up to `.tar.gz`, so I kept as is (I had thought it was "parity" with the `/commit` matching). Since we're v2-releasing this I can change that too

I see. The commit is omitted for the source-uri matching because it's not supposed to be there - it should be in a dedicated field `sha` according to SLSA specs. I don't know what the `<id>` means for bucket source. Does it make a difference? You may be right that it does not. Having sone clarification from GCB team may help validate we're not omitting an important parameter.

asraa:
> Does it make a difference? You may be right that it does not. Having sone clarification from GCB team may help validate we're not omitting an important parameter.

To some degree it may not matter at all until they expose source provenance FOR the bucket. That bucket just contains anything in the context of the `gcloud builds submit`. I know it is uploaded to the same bucket each time, with new versions which change the `#blah` fragment. I'll add some details to https://github.com/slsa-framework/slsa-verifier/issues/309

PR #302, opened 2022-10-13
rekor(backport/v1.3): use rekor client with retries

asraa:
* Uses default retry count 3 for rekor client

Part of https://github.com/slsa-framework/slsa-github-generator/issues/959


asraa:
lol, why didn't presubmits trigger?

ianlewis:
> lol, why didn't presubmits trigger?

It looks like we just don't have that many pre-submit workflows in the `release/v1.3` branch. We would need to backport the pre-submits as well if we want them to run.

ianlewis:
Close in favor of #329

PR #303, opened 2022-10-13
rekor(backport/v1.2): use rekor client with retries

asraa:


ianlewis:
Close in favor of #330

PR #310, opened 2022-10-19
test: add v1.2.1 builder tests

asraa:

cc @ianlewis @laurentsimon 

I have no idea why the test timing is growing so exponentially. It should be an extra 6 entries, which is annoying, and we should make some backported offline verification in tests to cache the rekor entry fetching.


laurentsimon (approved):

laurentsimon:
> I have no idea why the test timing is growing so exponentially. 

what do you mean? Any number you can share about timing?

PR #313, opened 2022-10-21
Fix installer: Add arguments to actions/checkout so that it checks ou…

kpk47:
…t slsa-framework/slsa-verifier instead of the repo using the Action.

Signed-off-by: kpk47 <kkris@google.com>


kpk47:
Sorry, wrong button.

asraa:
oooh! thanks -- could you add a quick pre-submit check? maybe use the action to verify a release?

kpk47:
Will do, but this PR isn't ready for review. I clicked the wrong button and opened it prematurely.

PR #316, opened 2022-10-21
Add optional

wietse-gmail:
This adds missing [optional] annotations to flags in the verify-artifact --help message.

Alternatively, we could just indicate what flags are [required] - there are only two of them - and remove all the [optional] noise from all other flags.


ianlewis (approved):
LGTM

PR #319, opened 2022-10-22
Fix installer: Add arguments to actions/checkout so that it checks ou…

kpk47:
Fix installer: Check out the correct repo and fix support for pinning by commit SHA.

I didn't add any tests in this PR because trying to test in this repo covered up these bugs in the first place. Instead, I've opened a corresponding PR against slsa-framework/example-package (https://github.com/slsa-framework/example-package/pull/116). The test fails now, but I'm confident it will succeed based on a separate branch that's targeting the branch this PR is based on (https://github.com/kpk47/example-package/pull/1/files -- ignore the workflow failures. They're due to provenance issues with the slsa-verifier versions built in my fork. The installer logic before the provenance verification works.)


actions/installer/action.yml
```diff
@@ -20,9 +20,16 @@ branding:
 runs:
   using: 'composite'
   steps:
+    - name: Get Action Ref
```
laurentsimon, line 23: we don't need this step if all it does it use the ACTION_REF. You can reference it directly using `${{ github.action_ref }}`

actions/installer/action.yml
```diff
@@ -45,4 +52,5 @@ runs:
         REPOSITORY: "${{ github.repository }}"
       working-directory: actions/installer/dist
       shell: bash
-      run: nodejs index.js 
+      run: npm start
+
```
laurentsimon, line 56: note: I don't think we need:
- repository option for the Action input. It's not used by the code, is it?
- composite action. ACTION_REF is accessible by TS actions, so we should be able to simplify action.yaml like in https://docs.github.com/en/actions/creating-actions/creating-a-javascript-action#creating-an-action-metadata-file (calling dist/index.js)?

kpk47: Done

actions/installer/action.yml
kpk47: Done

laurentsimon (approved):
Looking good, thanks!

PR #324, opened 2022-10-25
Make GitHub token optional

laurentsimon:
Signed-off-by: laurentsimon <64505099+laurentsimon@users.noreply.github.com>

### DIFF
--- a/actions/installer/action.yml
+++ b/actions/installer/action.yml
@@ -20,7 +20,8 @@ branding:
 inputs:
   github-token:
     description: 'GitHub token'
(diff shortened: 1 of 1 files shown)


asraa (approved):

laurentsimon:
@kpk47 I think it's fine to make the token optional, correct?

kpk47:
The Action makes a GitHub API call, so I think you'll get a 403 if you don't pass the token.

kpk47:
Oh, right. My mistake.

PR #327, opened 2022-10-27
feat: run CLI tests daily

laurentsimon:
As part of improving unit testing https://github.com/slsa-framework/slsa-github-generator/issues/1172

### DIFF
--- a/.github/workflows/pre-submit.cli.yml
+++ b/.github/workflows/pre-submit.cli.yml
@@ -1,5 +1,12 @@
-name: Pre submits
-on: [pull_request, workflow_dispatch]
+name: Pre submits cli
+on: 
+  pull_request:
+    branches: [ "main" ]
+  workflow_dispatch:
+  # Used in combination with schedule.cli.yml
+  # to avoid duplicating the test code.
+  schedule:
+    - cron: '25 6 * * 5'
 
 permissions: read-all
 
@@ -14,6 +21,17 @@ jobs:
         uses: actions/setup-go@268d8c0ca0432bb2cf416faae41297df9d262d7f # tag=v3.3.0
         with:
           go-version: '1.18'
+
+      - name: Save event name
+        env:
+          EVENT_NAME: ${{ github.event_name }}
+        run: |
+          echo $EVENT_NAME > ./event_name.txt
+
+      - uses: actions/upload-artifact@83fd05a356d7e2593de66fc9913b3002723633cb # tag=v3.1.1
+        with:
+          name: event_name
+          path: ./event_name.txt
 
       - name: Run tests for verifier
         run: |
--- a/.github/workflows/schedule.cli.yml
+++ b/.github/workflows/schedule.cli.yml
@@ -1,0 +1,38 @@
+name: Schedule cli
+on:
+  workflow_run:
+    workflows: ["Pre submits cli"]
+    types: [completed]
+    branches: [main]
+
+permissions: read-all
+
+env:
+  GH_TOKEN: ${{ secrets.CREATE_ISSUES }}
+  ISSUE_REPOSITORY: ${{ github.repository }}
+
+jobs:
+  if-failed:
+    runs-on: ubuntu-latest
+    # See https://github.com/orgs/community/discussions/26238.
+    if: github.event.workflow_run.conclusion != 'success'
+    steps:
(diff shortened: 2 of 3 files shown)


.github/workflows/pre-submit.cli.yml
```diff
@@ -1,5 +1,12 @@
-name: Pre submits
-on: [pull_request, workflow_dispatch]
+name: Pre submits cli
+on: 
+  pull_request:
+    branches: [ "main" ]
```
ianlewis, line 4: I think we want these to run on `release/*` branches as well when we backport. I'm not sure about the `test/e2e` branch.

.github/workflows/schedule.cli.yml
```diff
@@ -0,0 +1,38 @@
+name: Schedule cli
+on:
+  workflow_run:
+    workflows: ["Pre submits cli"]
+    types: [completed]
+    branches: [main]
```
ianlewis, line 6: again, we probably want this to run on `release/*` branches.

.github/workflows/pre-submit.cli.yml
```diff
@@ -1,5 +1,12 @@
...
+  workflow_dispatch:
+  # Used in combination with schedule.cli.yml
+  # to avoid duplicating the test code.
```
ianlewis, line 7: I'm not really understanding why the two files are necessary? It looks like you should be able to just include the `schedule.cli.yml` code here since it's only run when this worflow completes. Am I missing something?

.github/workflows/schedule.cli.yml
```diff
@@ -0,0 +1,38 @@
...
+        id: name
+        run: |
+          name=$(cat ./event_name.txt)
+          ctned="false"
+          if [[ "$name" == "schedule" ]] || [[ "$name" == "workflow_dispatch" ]]; then
+            ctned="true"
+          fi
+          echo "continue=$ctned" >> $GITHUB_OUTPUT
+      - uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # tag=v2.4.0
+        if: steps.name.outputs.continue == 'true'
```
ianlewis, line 33: nit: this is ok, but at some point we may want to just have steps like `name` fail and continue on failure instead of setting a `continue` output flag. That might be cleaner. I know we already use this pattern a lot in shims for our e2e tests.

.github/workflows/pre-submit.cli.yml
laurentsimon: done, thanks. I did not add the e2e test branch. Is it used?

.github/workflows/schedule.cli.yml
laurentsimon: I've removed it since it will be triggered by the other workflow.

.github/workflows/pre-submit.cli.yml
ianlewis: I'm not sure. We can leave it off for now.

.github/workflows/pre-submit.cli.yml
laurentsimon: yes I can. I separated them to have the workflow filenames be "consistent" with when they run. Instead I could rename the workflow `pre-submit-schedule.cli.yml` and have the code all in one place. Let me know.

.github/workflows/schedule.cli.yml
laurentsimon: I can change it if you think it's cleaner? Not clear to me

ianlewis: Yeah, this is fine for now. I'm mostly just musing about making this cleaner in the future.

.github/workflows/pre-submit.cli.yml
ianlewis: `schedule.cli.yml` would run on pull_requests wouldn't it? I think it runs whenever the "Pre submits cli" workflow is completed regardless of the trigger that started it doesn't it?

laurentsimon: @ianlewis shall I change this or keep as-is?

ianlewis: I'll go ahead and approve and we can adjust later if necessary.

laurentsimon: yes you're right it runs on PRs but does nothing except for schedule and workflows_dispatch events. (see `if`)

PR #328, opened 2022-10-27
fix: daily cli tests

laurentsimon:


asraa (commented):
Could you use https://github.com/marketplace/actions/download-workflow-artifact for simplicity?

.github/workflows/schedule.cli.yml
```diff
@@ -10,30 +10,57 @@ permissions: read-all
...
   if-failed:
     runs-on: ubuntu-latest
     # See https://github.com/orgs/community/discussions/26238.
     steps:
-      - uses: actions/download-artifact@9782bd6a9848b53b110e712e20e42d89988822b7 # tag=v1.10.0
+      - name: Download artifact
+        uses: actions/github-script@v6
         with:
-          name: event_name
+          script: |
```
ianlewis, line 25: This seems like a lot of complexity just so we can run these workflows in two files.

maybe you can consider abstracting out the code you want to avoid duplicating into another reusable workflow or action?

laurentsimon: the code to re-use is the one in pre-submit.cli.yml

laurentsimon:
> Could you use https://github.com/marketplace/actions/download-workflow-artifact for simplicity?

I can do that yes. I'll pin it by version since all permissions a read-only in this workflow. I don't want to add burden on PR because of this Action. Let me know if that's OK.

.github/workflows/schedule.cli.yml
laurentsimon: I could put the download script in an Action instead. Shall I do that now or wait to see if we have other use cases for it?

ianlewis: We can wait for now.

ianlewis:
@laurentsimon Is this still needed or can we close?

laurentsimon:
would be nice to have but not on my priority list atm

ianlewis:
Ok, let's go ahead and close since this is getting pretty old. We can revive it at some point if needed.

PR #348, opened 2022-11-02
refactor: consolidate verification funcs for GHA

asraa:

This consolidates the verification funcs for GHA provenance:
* Now we use rekor exported verification functions and replace our own verifyRootHash, inclusion proof, and SET verification (all contained in the `rverify.VerifyTlogEntry`)
* Factor out the DSSE signature verification


verifiers/internal/gha/rekor.go
```diff
@@ -334,25 +205,36 @@ func GetRekorEntriesWithCert(rClient *client.Rekor, provenance []byte) (*dsselib
 	}
 
 	logEntry := resp.Payload[0]
+	var ts time.Time
```
kpk47, line 208: What does this timestamp signify? From context, I can see that it's associated with the last log entry and is being used to verify the attestation signature, but it isn't clear to me how they're connected.

I'd suggest clarifying the relationship either through a less generic variable name or comments.

verifiers/internal/gha/provenance.go
```diff
@@ -150,27 +151,32 @@ func verifySha256Digest(prov *intoto.ProvenanceStatement, expectedHash string) e
...
 // VerifyProvenanceSignature returns the verified DSSE envelope containing the provenance
 // and the signing certificate given the provenance and artifact hash.
-func VerifyProvenanceSignature(ctx context.Context, rClient *client.Rekor, provenance []byte, artifactHash string) (*dsselib.Envelope, *x509.Certificate, error) {
-	// Get Rekor entries corresponding to provenance
-	env, cert, err := GetRekorEntriesWithCert(rClient, provenance)
-	if err == nil {
-		return env, cert, nil
+func VerifyProvenanceSignature(ctx context.Context, rClient *client.Rekor,
+	provenance []byte, artifactHash string) (
+	*dsselib.Envelope, *x509.Certificate, error) {
```
kpk47, line 156: I notice that you (almost?) always pass around the first two of these return values together, so would it make sense to wrap them in a struct that maps to a higher-level construct?

The triple return value looks odd to me and makes me wonder if the API has room for improvement. (This reaction could be a bias coming from C++, so feel free to ignore!)

asraa: Nope: I honestly thing we should do that! I think structs are better, and this triple return is pretty annoying.

verifiers/internal/gha/rekor.go
asraa: Got it: I should clarify maybe with the naming and the docs around the verification func. Because the attestation is signed with a short-lived cert, we need a timestamp.

verifiers/internal/gha/provenance.go
asraa: Done! I think that worked out nice: I can re-use it for verifying the signature using the cert and timestamp as well.

kpk47 (approved):

PR #356, opened 2022-11-11
fix: fix error logs

suzuki-shunsuke:
Fix error logs.

I guess these codes are copied from the just before error handling mistakenly.

https://github.com/slsa-framework/slsa-verifier/blob/137528a257f4533d7445ec6ae1bc57011784dab0/verifiers/internal/gha/rekor.go#L81


suzuki-shunsuke:
I think this is a trivial bug fix so an issue is unneeded, but if an issue is needed please let me know.

ianlewis (approved):

asraa (approved):
Thanks for this!

PR #361, opened 2022-11-25
Update README

ianlewis:
1. Adds some more basic explanation about SLSA to make it more accessible to folks less familiar with supply chain security.
1. Updates the TOC 
1. Adds badges and initial OpenSSF best practices badge. 
1. Formatted the README with prettier

Signed-off-by: Ian Lewis <ianmlewis@gmail.com>


asraa (approved):
Thank you, this looks great!

As an aside, we should mention something about leveling. I'll add in something in a follow up

ianlewis:
> As an aside, we should mention something about leveling. I'll add in something in a follow up

Awesome! If do you add something consider adding it to the slsa-github-generator repo's README as well.

Issue #371, opened 2022-11-29
Question/suggestion for v2.0.0-rc.0 release candidate


behnazh-w:
I was looking at the changelog for [`v2.0.0-rc.0` release candidate](https://github.com/slsa-framework/slsa-verifier/releases/tag/v2.0.0-rc.0) and couldn't figure out the breaking change that is resulting in a major version bump. I think using the `Breaking Change` notations in [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) can improve the changelogs. But because the commit history can't change now, it would help to add the breaking change to the release description.

Also, to make the changelog more readable, it would be good to run [commitizen](https://github.com/commitizen-tools/commitizen) (also available as a [GitHub Actions workflow](https://github.com/commitizen-tools/commitizen-action)) to filter out `chore` and other insignificant commits that don't need to be communicated to users of `slsa-verifier`.

ianlewis:
Hi, Thanks for the suggestions.

The reason for the bump is that the `slsa-verifier` CLI has changed in a backwards incompatible way. It now uses different flags (e.g. `--source-tag` vs. the old `--tag`) and has different commands (e.g. `slsa-verifier verify-artifact ...`, `slsa-verifier verify-image ...`).

I'll ask @asraa to add a human readable overview of the changes to the release.

jenstroeger:
> The reason for the bump is that the `slsa-verifier` CLI has changed in a backwards incompatible way. It now uses different flags (e.g. `--source-tag` vs. the old `--tag`) and has different commands (e.g. `slsa-verifier verify-artifact ...`, `slsa-verifier verify-image ...`).

@ianlewis in that case there should be a commit whose message contains that “breaking change” indicator — either in the message title or footer of the commit message, see [here](https://www.conventionalcommits.org/en/v1.0.0/#examples) for proper examples.

Also note that multiple commits in the [release candidate](https://github.com/slsa-framework/slsa-verifier/releases/tag/v2.0.0-rc.0) do not comply with the conventional commit spec. As @behnazh-w suggests, please consider using automation tools like commitizen to manage your commits and releases.

Otherwise, release version numbers and changelogs of this project are unreliable and more of a [sentimental version](http://sentimentalversioning.org/) than machine-usable for checking supply-chain integrity.

ianlewis:
[...]

PR #372, opened 2022-11-30
ci: add pr workflow to check pr title format

asraa:
* ci: adds a ci workflow on PRs to check the PR title format against regexes that follow conventional commits

Part of https://github.com/slsa-framework/slsa-verifier/issues/371

### DIFF
--- a/.github/pr-title-checker-config.json
+++ b/.github/pr-title-checker-config.json
@@ -1,17 +1,14 @@
-
 {
-    "LABEL": {
-      "name": "title needs adjustment",
-      "color": "EEEEEE"
-    },
-    "CHECKS": {
-      "prefixes": [
-        "chore: "      ],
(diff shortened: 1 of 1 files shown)


.github/pr-title-checker-config.json
```diff
@@ -0,0 +1,17 @@
+
+{
+    "LABEL": {
+      "name": "title needs adjustment",
+      "color": "EEEEEE"
+    },
+    "CHECKS": {
+      "prefixes": [
+        "chore: "      ],
```
ianlewis, line 9: nit this file is formatted a bit strangely.

ianlewis, line 9: nit: 'chore' seems to be handled by the regex?

ianlewis (approved):
We could use a similar config in slsa-github-generator's repo too.

.github/pr-title-checker-config.json
asraa: removing that cleans it up a bit.

Issue #377, opened 2022-12-02
Possible code duplication


lukehinds:
Was reading over the code and noticed the following:

https://github.com/slsa-framework/slsa-verifier/blob/798db79f54ede97ddd9a925fe7a7df39b496b7b7/verifiers/internal/gcb/provenance.go#L413-L430

There maybe some logic in the duplication, if so this is invalid, but I thought it worth raising in case it's a mistake and causes some edge case bug.

ianlewis:
@lukehinds Did you mean to close this? Even if the code is valid for some reason it's a bit unintuitive so perhaps could use a comment.

ianlewis:
It looks like where `decodeSignature` is used is also strange as it returns an `[]error` but that slice isn't checked.

https://github.com/slsa-framework/slsa-verifier/blob/798db79f54ede97ddd9a925fe7a7df39b496b7b7/verifiers/internal/gcb/provenance.go#L463-L467

ianlewis:
I'll reopen since this looks strange to me.

/cc @laurentsimon since it looks like this was added in #251

lukehinds:
> @lukehinds Did you mean to close this? Even if the code is valid for some reason it's a bit unintuitive so perhaps could use a comment.

I did, but perhaps it does merit some attention. My eyes did not make out URLEncoding and StdEncoding as different.

laurentsimon:
Let me explain. We try both encoding because, when verifying GCB provenance, it sometimes failed due to an encoding issue. So I added this "redundant-looking" base64 decoding to avoid the problem. I agree it's worth adding a comment, which I should have done :/

`decodeSIgnature()` is called https://github.com/slsa-framework/slsa-verifier/blob/798db79f54ede97ddd9a925fe7a7df39b496b7b7/verifiers/internal/gcb/provenance.go#L463
and appended to a list of errors https://github.com/slsa-framework/slsa-verifier/blob/798db79f54ede97ddd9a925fe7a7df39b496b7b7/verifiers/internal/gcb/provenance.go#L465

[...]

PR #380, opened 2022-12-02
docs: Add comment for signature decoding

laurentsimon:
closes https://github.com/slsa-framework/slsa-verifier/issues/377

We perform both std and URL base64 decoding, so adding a comment why.

### DIFF
--- a/verifiers/internal/gcb/provenance.go
+++ b/verifiers/internal/gcb/provenance.go
@@ -412,6 +412,7 @@ func (self *Provenance) VerifyVersionedTag(tag string) error {
 
 func decodeSignature(s string) ([]byte, []error) {
 	var errs []error
+	// First try the std decoding.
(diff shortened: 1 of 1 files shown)


verifiers/internal/gcb/provenance.go
```diff
@@ -412,13 +412,19 @@ func (self *Provenance) VerifyVersionedTag(tag string) error {
...
 	rsig, err := base64.StdEncoding.DecodeString(s)
 	if err == nil {
 		// No error, return the value.
 		return rsig, nil
 	}
 	errs = append(errs, err)
 
+	// If std decoding failed, try URL decoding.
+	// We try both because we encountered decoding failures
+	// during our tests. The DSSE documentation does not prescribes
```
asraa, line 425:
```suggestion
	// during our tests. The DSSE documentation does not prescribe
```

asraa (approved):

PR #385, opened 2022-12-03
fix: Fix error check for decodeSignature

ianlewis:
Updates #377 

Signed-off-by: Ian Lewis <ianmlewis@gmail.com>

### DIFF
--- a/verifiers/internal/gcb/provenance.go
+++ b/verifiers/internal/gcb/provenance.go
@@ -467,7 +467,7 @@ func (self *Provenance) verifySignatures(prov *provenance) error {
 
 			// Decode the signature.
 			rsig, es := decodeSignature(sig.Sig)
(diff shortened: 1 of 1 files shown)


laurentsimon (approved):
Great catch, thanks!

If decodeSignature failed, we'd  still be catching an error in VerifySignature(), so this should not be a vulnerability. We'd be missing some errors when report to the user.

Please confirm we have the same understanding.

ianlewis:
> Great catch, thanks!
> 
> If decodeSignature failed, we'd still be catching an error in VerifySignature(), so this should not be a vulnerability. We'd be missing some errors when report to the user.
> 
> Please confirm we have the same understanding.

Yes I think that's right. We would have just been discarding the actual errors returned by decodeSignature

PR #389, opened 2022-12-05
feat: add more tests for GCB verification

laurentsimon:
@lukehinds thanks for the https://github.com/slsa-framework/slsa-verifier/issues/377 you filed last week. Even though it was not a bypass (`verifySignature()` would fail), I have made some changes:
1. Removed the use of slices and replaced the return of the function with a more traditional `error`
2. Added unit tests for decoding function
3. Added more unit tests to VerifySignature()
4. Added more CLI tests


cli/slsa-verifier/main_test.go
```diff
@@ -838,6 +839,7 @@ func Test_runVerifyGCBArtifactImage(t *testing.T) {
 		source         string
 		pBuilderID     *string
 		outBuilderID   string
+		outProvenance  string
```
asraa, line 842: this is unused until the TODO?

laurentsimon: oops, I forgot to remove this. You're correct. Let me delete this field.

asraa (approved):

PR #413, opened 2022-12-15
ci: Add javascript to CodeQL analysis

ianlewis:
Signed-off-by: Ian Lewis <ianlewis@google.com>


laurentsimon (approved):
Any thoughts on making codeQL just a daily run, and create an issue if it finds something?

ianlewis:
> Any thoughts on making codeQL just a daily run, and create an issue if it finds something?

I'm ok with it running on PRs, esp since scorecard checks for it.

I'm really not sure why it's complaining about `sourcemap-register.js` since I think it's always the same but it doesn't complain on the slsa-github-generator repo... I'll check later.

laurentsimon:
> > Any thoughts on making codeQL just a daily run, and create an issue if it finds something?
> 
> I'm ok with it running on PRs, esp since scorecard checks for it.

That can be changed in scorecard too. There are other projects (systemd) maintainers who complain about the unfeasibility to run SAST on large code base. We're not large code base, but SAST on memory-safe languages seem to find very little in practice.

Issue #174, opened 2022-07-26
Attestation of 2 artifacts is considered valid even if only one is passed


mihaimaruseac:
```
mihaimaruseac@ankh:/tmp$ ~/go/bin/slsa-verifier -provenance attestation.intoto.jsonl -source mihaimaruseac/slsa-lvl3-generic-provenance-with-bazel-example -artifact-path fib -artifact-path hello
Verified signature against tlog entry index 3026659 at URL: https://rekor.sigstore.dev/api/v1/log/entries/b7d5d52dfc795dddf1c03e901e96c3cb1803c9f1ecc250a13299131b2564c422
Verified build using builder https://github.com/slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@refs/tags/v1.2.0 at commit e9b5431bcfb0df5d8b2ee622bde8ec4fad9d71d3
PASSED: Verified SLSA provenance
```

Getting the same result if passing only `fib` or `hello` arguments.

I could look into this.

laurentsimon:
> ```
> mihaimaruseac@ankh:/tmp$ ~/go/bin/slsa-verifier -provenance attestation.intoto.jsonl -source mihaimaruseac/slsa-lvl3-generic-provenance-with-bazel-example -artifact-path fib -artifact-path hello
> Verified signature against tlog entry index 3026659 at URL: https://rekor.sigstore.dev/api/v1/log/entries/b7d5d52dfc795dddf1c03e901e96c3cb1803c9f1ecc250a13299131b2564c422
> Verified build using builder https://github.com/slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@refs/tags/v1.2.0 at commit e9b5431bcfb0df5d8b2ee622bde8ec4fad9d71d3
> PASSED: Verified SLSA provenance
> ```
> 
> Getting the same result if passing only `fib` or `hello` arguments.

we did not intend to support multiple `-artifact-path`, but it could be a new feature . https://github.com/slsa-framework/slsa-verifier/issues/91 is related.

> I could look into this.

Awesome, assigned to you

mihaimaruseac:
Moving this to #91 as it contains all possible scenarios and relevant discussion

PR #418, opened 2022-12-28
feat: Enhance help message

mihaimaruseac:
One part that is missing in the output if invocation is wrong is that there is no indication that user has to pass path to a file (artifact/image).

The remaining potential errors from #173 are handled via #231.

Found while looking at #174.

Signed-off-by: Mihai Maruseac <mihaimaruseac@google.com>

### DIFF
--- a/cli/slsa-verifier/verify.go
+++ b/cli/slsa-verifier/verify.go
@@ -32,7 +32,7 @@ func verifyArtifactCmd() *cobra.Command {
 	o := &verify.VerifyOptions{}
 
 	cmd := &cobra.Command{
-		Use: "verify-artifact",
+		Use: "verify-artifact [flags] artifact",
 		Args: func(cmd *cobra.Command, args []string) error {
 			if len(args) != 1 {
 				return errors.New("expects a single path to an artifact")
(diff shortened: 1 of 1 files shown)


cli/slsa-verifier/verify.go
```diff
@@ -32,7 +32,7 @@ func verifyArtifactCmd() *cobra.Command {
 	o := &verify.VerifyOptions{}
 
 	cmd := &cobra.Command{
-		Use: "verify-artifact",
+		Use: "verify-artifact [flags] artifact",
```
laurentsimon, line 35: Do `[flags]` have to come after `artifact`, as `verify-artifact artifact [flags]`. That's what the doc says https://github.com/slsa-framework/slsa-verifier#artifacts, but maybe what you wrote works too?

mihaimaruseac: I usually put them before, they can come after too. In general, it seems Linux prefers to put the options (`[flags]`) before the arguments in help messages, even if they could be used after:

```
[~] λ ls --help
Usage: ls [OPTION]... [FILE]...
...
```

```
[~] λ gcc --help
Usage: gcc [options] file...
Options:
...
```

So I thought I'd follow the same approach

laurentsimon: /cc @asraa

laurentsimon (approved):

cli/slsa-verifier/verify.go
asraa: with Cobra it can be either or! I prefer putting arguments before as well +1

asraa (approved):
thank you!

PR #424, opened 2022-12-29
fix: Use `Run` instead of `RunE` to handle usage/errors

mihaimaruseac:
We want to only display the usage message if there is an error in the parsing of arguments. Once the arguments are parsed and we invoke the internal implementation the usage message should not be displayed anymore, regardless of what internal error gets returned. Fortunately, the fix is easy.

Tested:

* passing scenario:
```
[...]$ slsa-verifier verify-artifact \
       --provenance-path multiple.intoto.jsonl \
       --source-uri github.com/example/repo fib
Verified signature against tlog entry index 9712459 at URL: ...
Verified build using builder ...
Verifying artifact fib: PASSED

PASSED: Verified SLSA provenance
```

* failing scenario with missing artifact file:
```
[...]$ slsa-verifier verify-artifact \
       --provenance-path multiple.intoto.jsonl \
       --source-uri github.com/example/repo fibs
Verifying artifact fibs: FAILED: open fibs: no such file or directory

FAILED: SLSA verification failed: open fibs: no such file or directory
```

* failing scenario with invalid artifact:
```
[...]$ slsa-verifier verify-artifact \
       --provenance-path multiple.intoto.jsonl \
       --source-uri github.com/example/repo multiple.intoto.jsonl
Verified signature against tlog entry index 9712459 at URL: ...
Verifying artifact multiple.intoto.jsonl: FAILED: expected hash ...

FAILED: SLSA verification failed: expected hash ...
```

* failing scenario due to invalid usage (missing required arguments):
```
[...]$ slsa-verifier verify-artifact \
       --provenance-path multiple.intoto.jsonl
Usage:
  slsa-verifier verify-artifact [flags] artifact [artifact..]

Flags:
      --build-workflow-input map[]    [optional] a workflow input provided by a user at trigger time in the format 'key=value'. (Only for 'workflow_dispatch' events on GitHub Actions). (default map[])
      --builder-id string             [optional] the unique builder ID who created the provenance
  -h, --help                          help for verify-artifact
      --print-provenance              [optional] print the verified provenance to stdout
      --provenance-path string        path to a provenance file
      --source-branch string          [optional] expected branch the binary was compiled from
      --source-tag string             [optional] expected tag the binary was compiled from
      --source-uri string             expected source repository that should have produced the binary, e.g. github.com/some/repo
      --source-versioned-tag string   [optional] expected version the binary was compiled from. Uses semantic version to match the tag

required flag(s) "source-uri" not set
exit status 1
```

Fixes #307

[...]

Issue #382, opened 2022-12-02
feaet: migrate main_test.go to execute the CLI directly


asraa:
Currently, we populate the Verify*Command's properties directly, instead of using CLI flags. I added support for additionally testing CLI flags, but this would only allow us to test the output and success conditions. The Verify*Commands also return the `outBuilderId`, which we use to validate matches against the input builder id.

Was this check necessary @laurentsimon ? Or should it be implicit by success/failure of the CLI command?

laurentsimon:
I think you're right that if the CLI succeeds, it means the builderId verification passed. I think these tests were created before we had the CLI options...?

asraa:
Oh, it's possible! In that case, I'll put up a cleanup fix after this patch release.

laurentsimon:
Before we do this migration, shall we re-factor https://github.com/slsa-framework/slsa-verifier/blob/main/options/options.go to be the public-facing command which mirrors the CLI?

Right now we have to construct manually https://github.com/slsa-framework/slsa-verifier/blob/main/cli/slsa-verifier/main_test.go#L582-L608. The Go Api should not ask users to perform these steps, right? I'm thinking more something like https://github.com/ossf/scorecard/blob/main/options/options.go

The existing options.go can be probably still be used, but made internal.

I have not thought all this through yet...

PR #429, opened 2023-01-04
fix: fix exit status on command execution errors

asraa:

Fixes e2e tests, for e.g. https://github.com/slsa-framework/slsa-github-generator/issues/1445

The error showed that the exit code for failing verification was 0. This was a small regression from PR https://github.com/slsa-framework/slsa-verifier/pull/424 that changed RunE to Run, and so the exit status was removed.

### DIFF
--- a/cli/slsa-verifier/verify.go
+++ b/cli/slsa-verifier/verify.go
@@ -62,6 +62,7 @@ func verifyArtifactCmd() *cobra.Command {
 
 			if _, err := v.Exec(cmd.Context(), args); err != nil {
 				fmt.Fprintf(os.Stderr, "%s: %v\n", FAILURE, err)
+				os.Exit(1)
 			} else {
 				fmt.Fprintf(os.Stderr, "%s\n", SUCCESS)
 			}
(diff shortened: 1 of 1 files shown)


cli/slsa-verifier/verify.go
```diff
@@ -62,6 +62,7 @@ func verifyArtifactCmd() *cobra.Command {
 
 			if _, err := v.Exec(cmd.Context(), args); err != nil {
 				fmt.Fprintf(os.Stderr, "%s: %v\n", FAILURE, err)
+				os.Exit(1)
```
laurentsimon, line 65: any chance we can have a pre-submit for it, or not worth it?
I was thinking of building the verifier and running it in a pre-submit. I'm not sure if it's worth it, since the code rarely changes and would be caught via the daily e2e runs.

Wdut?

laurentsimon (approved):

cli/slsa-verifier/verify.go
asraa: > any chance we can have a pre-submit for it, or not worth it?

Yeah, I think this is worth it, and I can write a follow-up using the example-package scripts. What would be nice is if we can re-use artifact/provenance from example-package though. So something like:
1. presubmit builds verifier at HEAD
2. fetches some latest artifacts from e2e package 
3. runs the same e2e.common.default.verify.sh script 

Or should we just check in some artifacts/use the testdata? Basically we just need to make sure the extensive e2e verifications are being run with the CLI itself.

laurentsimon: we can re-use all the test in main_test.go, on the CLI itself. I think that's what you're suggesting. Works for me!
https://github.com/slsa-framework/slsa-verifier/issues/382 is related, but I don't think we would catch the exit status in the code itself, ie we need to run the CLI in a script

PR #430, opened 2023-01-04
ci: add verifier e2e presubmit that runs CLI at main

asraa:

Adds a verifier e2e presubmit that mimics the example-package test with go/workflow dispatch. This should catch regressions early that show up in the e2e tests.

Related https://github.com/slsa-framework/slsa-verifier/issues/382


laurentsimon (approved):

ianlewis:
@asraa I think you checked in a slsa-verifier binary into the repo on this PR. Could you remove it?

PR #435, opened 2023-01-06
ci: fix a deprecation warning

suzuki-shunsuke:
> args
> The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


asraa (approved):
nice, thank you!

PR #439, opened 2023-01-11
test: add v1.4.0 build tests for gha_go gha_generic and gha_generic_container

asraa:
Adds v1.4.0 builder tests for github builders.

@ianlewis there aren't currently any tag matches for gha_generic_container. Want me to add in a follow-up?

@laurentsimon should this line be commented out? https://github.com/slsa-framework/slsa-verifier/blob/main/cli/slsa-verifier/main_regression_test.go#L587. Testing the outBuilderID against `@${TAG}` doesn't work for GHA, but it does for GCB. If so, should I remove these comments for GHA artifact/image?


laurentsimon (approved):
Thanks. Will let @ianlewis provide the final LGTM since he knows that part of the code better than I do

asraa:
> Is that a bug I should fix then?

I think it's just a issue in the test code. Providing the `--builder-id` for GHA seems to work with `@${TAG}`, `@refs/head/${TAG}` and no tag.

I'll file an issue for that to track.

asraa:
can you please take a look when you have a chance (esp at the comment)? @ianlewis

ianlewis:
> can you please take a look when you have a chance (esp at the comment)? @ianlewis

Sorry, I did see this but I didn't get to it yesterday.

> Thanks. Will let @ianlewis provide the final LGTM since he knows that part of the code better than I do

I do?!?! 😵‍💫 

> @ianlewis there aren't currently any tag matches for gha_generic_container. Want me to add in a follow-up?

I'm not sure I know what you mean.  Are you saying that we have a "tag no match empty tag workflow_dispatch" test but not one that matches the tag successfully?

If so, then yes please do add in follow up.

> I think it's just a issue in the test code. Providing the `--builder-id` for GHA seems to work with `@${TAG}`, `@refs/head/${TAG}` and no tag.
> 
> I'll file an issue for that to track.

Please add a TODO(issue url) comment after the issue's been created.

ianlewis:
BTW, related to https://github.com/slsa-framework/slsa-github-generator/issues/1110

asraa:
> Are you saying that we have a "tag no match empty tag workflow_dispatch" test but not one that matches the tag successfully?

yes exactly! they are generated in `example-package`, but the tests just aren't written. Will update in a min.

Issue #441, opened 2023-01-11
[feature] Support v1.0 SLSA provenance format


asraa:
We need to be able to extract builder ID and context info from the new format

PR #442, opened 2023-01-11
feat: support sigstore bundle format for offline verification

asraa:

Blocked on https://github.com/slsa-framework/slsa-verifier/issues/441 (builder implementations use 1.0 format)


asraa (commented):
`certificate does not include required embedded SCT`: weird that these errors are JUST popping up without any sigstore dep bumps.

asraa:
Splitting into 2

Issue #440, opened 2023-01-11
[test] Cleanup builder-id in main_regression_test


asraa:
Testing against the `builder@TAG` is commented out for the GHA artifacts and images here: https://github.com/slsa-framework/slsa-verifier/blob/main/cli/slsa-verifier/main_regression_test.go#L587

It seems to pass on the CLI when running tests, but https://github.com/slsa-framework/slsa-verifier/blob/main/cli/slsa-verifier/main_regression_test.go#L629 fails when running the test.

I'm not sure why the `outBuilderID.Matches` is needed to test (redundancy?).

cc @laurentsimon

asraa:
Laurent also mentioned

> I think it's just a issue in the test code. Providing the --builder-id for GHA seems to work with @${TAG}, @refs/head/${TAG} and no tag.

PR #455, opened 2023-01-25
test: add builder id tests for short form

asraa:

Fixes https://github.com/slsa-framework/slsa-verifier/issues/440

This adds back in the commented out tests for testing the short form of the builder ID for GHA artifacts and images `builder@TAG`. The tests were failing because they set `allowRef = false`, but for GHA, the builder identity returned is always in long form, meaning that we have to allow the short form matches in the test (which is also done in the code path: https://github.com/slsa-framework/slsa-verifier/blob/5eea7c55379c777df48a3e5a95cedea44169f6c1/verifiers/internal/gha/builder.go#L105)

cc @laurentsimon


laurentsimon (approved):

laurentsimon:
Good catch!

Issue #363, opened 2022-11-25
Enable Go linters


ianlewis:

asraa:
deadcode is deprecated and replaced by unused. just FYI

ianlewis:
> deadcode is deprecated and replaced by unused. just FYI

Yeah, we can just remove it and `varcheck` (though I see you removed it already) since we already have `unused` enabled.

asraa:
> Yeah, we can just remove it and varcheck (though I see you removed it already) since we already have unused enabled.

Weirdly enough, it still seemed to be working. I removed some deadcode because of it's alert, but when I remove the deadcode check and introduce it again, it doesn't complain. It seems to be triggering either some form of unused w/ deadcode, OR it is still working but is staged to be deprecated. I'll investigate if I can remove it in a further PR

ianlewis:
Removed deadcode here: https://github.com/slsa-framework/slsa-verifier/pull/572/files#diff-6179837f7df53a6f05c522b6b7bb566d484d5465d9894fb04910dd08bb40dcc9L24

PR #456, opened 2023-01-25
chore: enable some Go linters

asraa:
Part of https://github.com/slsa-framework/slsa-verifier/issues/363

### DIFF
--- a/cli/slsa-verifier/main.go
+++ b/cli/slsa-verifier/main.go
@@ -16,10 +16,6 @@ func check(err error) {
 	}
 }
 
-func ExperimentalEnabled() bool {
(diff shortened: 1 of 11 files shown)


cli/slsa-verifier/main.go
```diff
@@ -16,10 +16,6 @@ func check(err error) {
 	}
 }
 
-func ExperimentalEnabled() bool {
```
laurentsimon, line 19: Maybe we can keep this and disable the linter `//nolint: something`, since I suspect we'll use it again

laurentsimon (approved):

cli/slsa-verifier/main.go
asraa: sgtm, agree

PR #457, opened 2023-01-27
feat: add offline bundle signature verification

asraa:

* Adds cryptographic bundle verification following the Sigstore client spec. I expect this will be replaced by a direct call to another library in the long term
* Removed a test case that was no longer relevant. That test was checking our code against the Rekor server update that returned a new return response. Now all code tests the new server response anyway.
* Pipes out trusted root info outside the verification calls
* Updates cosign to v2


verifiers/internal/gha/provenance_test.go
```diff
@@ -969,3 +970,35 @@ func Test_VerifyVersionedTag(t *testing.T) {
...
+func Test_verifyBundle(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name     string
+		path     string
+		expected error
+	}{
+		{
+			name: "valid",
+			path: "./testdata/bundle/attestation.intoto.sigstore",
```
laurentsimon, line 983: we'll add invalid / adversarial ones inn the future?

verifiers/internal/gha/rekor.go
```diff
@@ -67,13 +67,17 @@ func verifyTlogEntryByUUID(ctx context.Context, rekorClient *client.Rekor, entry
 			return nil, errors.New("expected matching UUID")
 		}
 		// Validate the entry response.
-		return verifyTlogEntry(ctx, rekorClient, entry)
+		return verifyTlogEntry(ctx, entry, true)
```
laurentsimon, line 70: iiuc, we'll perform an online verification by default to be consistent with our existing code. And we'll provide an option to turn it off in the futuer, e.g. `rekor-log-offline`?

laurentsimon (approved):
Looks great!

laurentsimon:
qq: does the verification also verifies the SCT (ie the fulcio CA cert was recorded in the TLS CT log)?
something we need to do is unable this for the TLS connection as well, correct?

verifiers/internal/gha/provenance_test.go
asraa: Just added one or two that were easily to manipulate without creating a test Rekor signer. Filed a follow-up issue 
https://github.com/slsa-framework/slsa-verifier/issues/461

We'll be generating a suite from sigstore-go pretty soon, so we can close this out either by borrowing cases there or moving the code to call a verification func there that's tested.

verifiers/internal/gha/rekor.go
asraa: Right now the existing code still validates the inclusion proof (which requires online).

I haven't configured augmenting the offline verification with an online proof yet, this code just checks verification of the offline material. I'll file an issue though to augment with an option to check inclusion online as well.

asraa:
> qq: does the verification also verifies the SCT (ie the fulcio CA cert was recorded in the TLS CT log)?

Yes! By default now the check is on

> Are the leaf cert also appended to the PKI CT logs? Or only CA certs are?

Hmm not sure I totally understand this. The leaf cert is in the CT log.

laurentsimon:
> > qq: does the verification also verifies the SCT (ie the fulcio CA cert was recorded in the TLS CT log)?
> 
> Yes! By default now the check is on

+1

> > Are the leaf cert also appended to the PKI CT logs? Or only CA certs are?
> 
> Hmm not sure I totally understand this. The leaf cert is in the CT log.

Are CA certs (ie Fulcio cert) in CT log as well? 

How about rekor cert (which is a leaf cert)?

Which are verified?

asraa:
> Are CA certs (ie Fulcio cert) in CT log as well? Do we care?

Oh, yes the CA cert is in the CT log. SCT verification being enforced now (regenerating test cases, this is why they are failing) checks the offline promise of that. We don't check them online, but neither does normal CT. So this part is now enforced/verified.

> How about rekor cert (which is a leaf cert)? Do we care?

This I do not understand. Rekor signs with a public key that is retrieved from the TUF root and does not have a leaf cert.

laurentsimon:
> > Are CA certs (ie Fulcio cert) in CT log as well? Do we care?
> 
> Oh, yes the CA cert is in the CT log. SCT verification being enforced now (regenerating test cases, this is why they are failing) checks the offline promise of that. We don't check them online, but neither does normal CT. So this part is now enforced/verified.
> 
> > How about rekor cert (which is a leaf cert)? Do we care?
> 
> This I do not understand. Rekor signs with a public key that is retrieved from the TUF root and does not have a leaf cert.

OK. I thought maybe there was a cert issued as well. Ignore my comment then.

verifiers/internal/gha/builder.go
```diff
@@ -17,6 +17,9 @@ var (
 	trustedBuilderRepository = "slsa-framework/slsa-github-generator"
 	e2eTestRepository        = "slsa-framework/example-package"
 	certOidcIssuer           = "https://token.actions.githubusercontent.com"
+	// This is used in cosign's CheckOpts for validating the certificate. We
+	// do specific builder verification after this.
+	certSubjectRegexp = "https://github.com/*"
```
laurentsimon, line 22: that will also be necessary for npm provenance verification (the one without the trusted builder)

asraa: do we intend on verifying non-L3 provenance though?

but yes - it's just a nit that cosign v2 requires some validation on it so I had to add it even though we do a more detail check on it later.

laurentsimon:
LGTM to merge.

PR #463, opened 2023-02-03
refactor: generalize provenance out of predicate type info

asraa:

This generalized the provenance we verify in GHA to work against a generic Provenance interface, which may be provenance v0.2 or soon, v1.0

We can eventually move this package into the common code, when GCB also makes a migration.


verifiers/internal/gha/slsaprovenance/v0.2/provenance.go
```diff
@@ -0,0 +1,98 @@
...
+type ProvenanceV02 struct {
+	*intoto.ProvenanceStatement
+}
+
+func New() slsaprovenance.Provenance {
+	return &ProvenanceV02{}
+}
+
+func (prov *ProvenanceV02) BuilderID() string {
+	return prov.Predicate.Builder.ID
```
laurentsimon, line 29: can the value be nil and panic? Should every function return a `(type, error)`?
Same question for other functions.

verifiers/internal/gha/slsaprovenance/v0.2/provenance.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+type ProvenanceV02 struct {
+	*intoto.ProvenanceStatement
+}
+
+func New() slsaprovenance.Provenance {
+	return &ProvenanceV02{}
+}
+
```
laurentsimon, line 27: nit: add some description or the linter will complain at some point

verifiers/internal/gha/slsaprovenance/slsaprovenance.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// ConfigURI is the full URI (including tag) of the configuration material.
+	ConfigURI() string
+
+	// Subject is the list of intoto subjects in the provenance.
+	Subjects() []intoto.Subject
+
+	// GetStringFromEnvironment retrieves a string parameter from the environment
+	// attested to in the provenance.
+	GetStringFromEnvironment(name string) (string, error)
```
laurentsimon, line 29: do you want to go further and have the high-level functions `Branch()`, `Tag()`?

laurentsimon (approved):

verifiers/internal/gha/provenance_test.go
```diff
@@ -564,7 +574,7 @@ func Test_VerifyWorkflowInputs(t *testing.T) {
 				"some_bool":       "true",
 				"some_integer":    "123",
 			},
-			expected: serrors.ErrorInvalidDssePayload,
+			expected: serrors.ErrorMismatchWorkflowInputs,
```
asraa, line 577: Reviewers: I had to update this testcase and the error it generated because it was actually a weirdly formed DSSE envelope input (doubly encoded), and my code made it fail when it was detecting predicateType. Meanwhile, the error had to change from InvalidDssePayload to the actual error check, which is mismatch workflow inputs.

verifiers/internal/gha/slsaprovenance/v0.2/provenance.go
asraa: This one I can add a panic condition, although we do require a builder.id no matter what predicate type. So this one is particular unlikely to crash, because unmarshalling will always catch if the field is missing.

But for safety, yes -- I'll add errors proactively

verifiers/internal/gha/slsaprovenance/slsaprovenance.go
asraa: Yeah, I was a little bit unsure. On one hand it makes it very clear what info we require from the provenance. On the other hand I wasn't too sure to optimize away all of this too early. 

I will follow-up w/ an issue and mark as fixed when I add the v1.0 impl

verifiers/internal/gha/slsaprovenance/v0.2/provenance.go
laurentsimon: Let's avoid panic, especially if it's run as a service in the long-term, we want to always return cleanly

verifiers/internal/gha/slsaprovenance/slsaprovenance.go
asraa: https://github.com/slsa-framework/slsa-verifier/issues/464

laurentsimon: SG!

asraa: FWIW i'm leaning towards exposing the high-level functions. Addressing in my next follow-up!
Thanks!

laurentsimon: sure, no urgency on this

Issue #464, opened 2023-02-03
[v1.0] Use high-level accessors for Branch(), Tag() on provenance


asraa:
We currently define a Provenance interface that is implemented by v0.2 and v1.0 predicate types. They expose out environment vars.

Consider exposing the Branch, Tag, and other important info directly.

Issue #462, opened 2023-02-01
Add an option to additionally verify the offline Rekor entry in the bundle with an online inclusion proof


asraa:

laurentsimon:
Fine to close.

PR #470, opened 2023-02-08
feat: add slsa v1?draft provenance experimental support

asraa:

Description:
* feat: Adds slsa v1?draft provenance support with unit SLSA v1 provenance tests
* feat: [EXPERIMENTAL] Adds  `--bundle-path` option to specify a Sigstore bundle.

* test: I've added unit tests with the SLSA v1 provenance.

Follow-up issue:
 * e2e test the experimental gated --bundle-path: https://github.com/slsa-framework/slsa-verifier/issues/471. But first, I require generator PRs to be merged so my e2e tests can be completed.
 * Complete Tag() and Branch() checks for SLSA v1 provenance: https://github.com/slsa-framework/slsa-verifier/issues/472

Fixes https://github.com/slsa-framework/slsa-verifier/issues/464
Fixes https://github.com/slsa-framework/slsa-verifier/issues/441


asraa:
cc @laurentsimon this is ready for review.

verifiers/internal/gha/slsaprovenance/v0.2/provenance.go
```diff
@@ -101,3 +127,94 @@ func (prov *ProvenanceV02) GetInputs() (map[string]interface{}, error) {
...
+func getEventPayload(environment map[string]interface{}) (map[string]interface{}, error) {
+	eventPayload, ok := environment["github_event_payload"]
+	if !ok {
+		return nil, fmt.Errorf("%w: %s", serrors.ErrorInvalidDssePayload, "parameters type event payload")
+	}
+
+	payload, ok := eventPayload.(map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf("%w: %s", serrors.ErrorInvalidDssePayload, "parameters type payload")
+	}
```
ianlewis, line 217: nit: Might as well use `any`
```suggestion
func getEventPayload(environment map[string]any) (map[string]any, error) {
	eventPayload, ok := environment["github_event_payload"]
	if !ok {
		return nil, fmt.Errorf("%w: %s", serrors.ErrorInvalidDssePayload, "parameters type event payload")
	}

	payload, ok := eventPayload.(map[string]any)
	if !ok {
		return nil, fmt.Errorf("%w: %s", serrors.ErrorInvalidDssePayload, "parameters type payload")
	}
```

verifiers/internal/gha/slsaprovenance/v1.0/provenance.go
```diff
@@ -0,0 +1,75 @@
...
+func (prov *ProvenanceV1) Subjects() ([]intoto.Subject, error) {
+	subj := prov.Subject
+	if len(subj) == 0 {
+		return nil, fmt.Errorf("%w: %s", serrors.ErrorInvalidDssePayload, "no subjects")
+	}
+	return subj, nil
+}
+
+func (prov *ProvenanceV1) GetBranch() (string, error) {
+	return "", errors.New("unimplemented")
```
ianlewis, line 66: nit: should we have TODOs/issues for the inimplemented methods?

verifiers/internal/gha/slsaprovenance/v1.0/provenance.go
```diff
@@ -0,0 +1,75 @@
...
+	"errors"
+	"fmt"
+
+	intoto "github.com/in-toto/in-toto-golang/in_toto"
+	slsa1 "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v1.0"
+	serrors "github.com/slsa-framework/slsa-verifier/v2/errors"
+	"github.com/slsa-framework/slsa-verifier/v2/verifiers/internal/gha/slsaprovenance"
+)
+
+// TODO(asraa): Use a static mapping.
```
ianlewis, line 13: nit: let's point to issues rather than usernames for TODOs

verifiers/utils/provenance.go
```diff
@@ -0,0 +1,21 @@
+package utils
+
+import (
+	"fmt"
+
+	serrors "github.com/slsa-framework/slsa-verifier/v2/errors"
+)
+
+func GetAsString(environment map[string]interface{}, field string) (string, error) {
```
ianlewis, line 9: nit: use `any`
```suggestion
func GetAsString(environment map[string]any, field string) (string, error) {
```

ianlewis (approved):

cli/slsa-verifier/verify.go
```diff
@@ -60,6 +61,21 @@ func verifyArtifactCmd() *cobra.Command {
 				v.BuilderID = &o.BuilderID
 			}
 
+			// In experimental mode, we allow either provenance or bundle path, but exactly
+			// one must be set. We already check to ensure that they are mutually exclusive.
+			if ExperimentalEnabled() {
+				if !(cmd.Flags().Changed("provenance-path") ||
+					cmd.Flags().Changed("bundle-path")) {
```
laurentsimon, line 68: I'm curious if we want another options for this or simply re-use `provenance-path`?
Does a new option require users to be "aware" of the "type" of provenance file they want to verify? Is this OK?

asraa: Hmm good point, from our perspective we will be having generators that may output DSSE or sigstore bundle, and clients may still want a uniform interface rather than switching over the extension type. 

I can update to allow that -- we'll have to test unmarshal both to figure out the right type, but it will actually clean up the interface.

ianlewis: I thought about this too but didn't comment since I figured that having separate options has the benefit of avoiding format type detection logic. If all types have the same security implications then probably it's fine, but if one format could be more secure than others then we might have to worry about downgrades.

asraa: https://github.com/slsa-framework/slsa-verifier/issues/462

So long as we turn this option on my default the same content is present and same validations will occur. So let's definitely resolve this one before releasing.

Issue #485, opened 2023-02-15
[tests] Add all docker-based testing


asraa:
They can be merged with the other GHA artifact tests

PR #486, opened 2023-02-15
test: add docker based spport and start adding tests

asraa:
This PR adds the docker-based workflow to the supported trusted builders.

We currently only have `main` released for it, so I've began by only adding main tests. Follow-up here: https://github.com/slsa-framework/slsa-verifier/issues/485

This is also gated behind `SLSA_VERIFIER_EXPERIMENTAL` due to the bundle format, anyway.


laurentsimon (approved):
want to create an issue tracking that the verification will need to be updated to verify `@<tag>` and not `@main`?

laurentsimon:
> want to create an issue tracking that the verification will need to be updated to verify `@<tag>` and not `@main`?

Ignore the comment, just saw the existing issue in the description.

laurentsimon:
> We currently only have main released for it

these test will need to be updated in the follow-up PR, correct?

asraa:
> these test will need to be updated in the follow-up PR, correct?

Yes! Right now this unblocks us from using `verify` in the e2e package repo though, so that's the idea. Without that we can't test @main and actually GA a release.

PR #509, opened 2023-03-01
feat: Update SLSA verifier to support a global signing key for GCB V1 which…

khalkie:
… creates the signature in a DSSE-conformant PAE format

- new public key for "global PAE signing key"
- test data and unit tests


asraa (commented):
Thank you so much!

Instead of special casing this, could you make https://github.com/slsa-framework/slsa-verifier/blob/12910ea59692bc0e18229167ef9aac190d937679/verifiers/internal/gcb/keys/keys.go#L25 return a PublicKey interface and if `PublicKeyNew` sees a `global` region, then it creates a `GlobalPublicKey` that has the `VerifySignature` method as you have it using dsselib?

That way there's no repeated verify call or repeated opening for the PEM files

khalkie:
> Thank you so much!
> 
> Instead of special casing this, could you make
> 
> https://github.com/slsa-framework/slsa-verifier/blob/12910ea59692bc0e18229167ef9aac190d937679/verifiers/internal/gcb/keys/keys.go#L25
> 
> return a PublicKey interface and if `PublicKeyNew` sees a `global` region, then it creates a `GlobalPublicKey` that has the `VerifySignature` method as you have it using dsselib?
> That way there's no repeated verify call or repeated opening for the PEM files

Ah I see, that's a good suggestion! 

Being global is not actually the only thing special about this case as well - there is already a "global" regional key in the existing verification. This key is global also, but the main difference is that the signature was generated using PAE encoding of the envelope, I wasn't quite sure how to describe that difference in a variable/struct name.

verifiers/internal/gcb/keys/keys.go
```diff
@@ -1,20 +1,26 @@
...
 	serrors "github.com/slsa-framework/slsa-verifier/v2/errors"
 )
 
 //go:embed materials/*
 var publicKeys embed.FS
 
+const GLOBAL_PAE_KEY_ID = "projects/verified-builder/locations/global/keyRings/attestor/cryptoKeys/provenanceSigner/cryptoKeyVersions/1"
+const GLOBAL_PAE_PUBLIC_KEY_NAME = "global-pae"
+
 type PublicKey struct {
```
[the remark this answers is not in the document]
khalkie: I do still use one to verify the PAE signature as well

verifiers/internal/gcb/provenance.go
```diff
@@ -448,16 +463,18 @@ func (p *Provenance) verifySignatures(prov *provenance) error {
 				errs = append(errs, err)
 				continue
 			}
+		} else {
+			continue
```
asraa, line 467: is this else continue necessary?

asraa (commented):
Thank you! Great changes, one very very small comment, but this looks great!

verifiers/internal/gcb/provenance.go
khalkie: If we reach this 'else', it will mean this signature did not match the global PAE key, or the regional key regex. If we don't continue here, the 'Verification succeeded' code below will be executed

asraa: AH I mixed up which lines were in that scope, thanks!

asraa (approved):

verifiers/internal/gcb/provenance.go
```diff
@@ -416,20 +418,33 @@ func (p *Provenance) verifySignatures(prov *provenance) error {
...
+				errs = append(errs, err)
+				continue
+			}
+
+			err = globalPaeKey.VerifyEnvelope(&prov.Envelope)
+			if err != nil {
+				errs = append(errs, err)
+				continue
+			}
+		} else if match := regionalKeyRegex.FindStringSubmatch(sig.KeyID); len(match) == 2 {
```
laurentsimon, line 445: nit: please add a comment that this is region keys without PAE, which was the original implementation. PAE is the recent update.

laurentsimon (approved):

PR #519, opened 2023-03-08
fix: fix GCB verification with git material source prefix

asraa:
The most recent GCB build I've pushed uses a `git+` prefix on the source. Updates slsa-verifier to allow these for matching against expected provenance.

See this failing run: https://github.com/slsa-framework/example-package/actions/runs/4368807186/jobs/7641862328#step:5:205

cc @laurentsimon @ianlewis


laurentsimon:
@khalkie not sure if this was flagged as a potential breaking change. Shall we work on better integration between GCB and this repo?

ianlewis (approved):

khalkie:
Sorry, my fault - the materials without "git+" were failing the BCID verifier, so I assumed they were required for SLSA as well. I did have it in my test provenance files, and added it to several unit tests - I thought I was testing the end-to-end validation in the VerifyTestProvenance test but I missed this case. It's a good idea to incorporate the SLSA verifier into our release tests

laurentsimon:
np, it happens. During the review, we missed that the tests were missing too!
Let's improve our collaboration to have slsa-verifier run on GCB code changes, and it should catch regression problems in the future. 

Thanks @khalkie again for the contribution, much appreciated!

Issue #492, opened 2023-02-18
[feature] Re-enable material URI verification


laurentsimon:
Temporarily disabled for https://github.com/gh-community/npm-provenance-private-beta-community/issues/8

laurentsimon:
Below is the content of the issue description:

Seems like materials and configSource don't use the same URI:
```json
{
  "_type": "https://in-toto.io/Statement/v0.1",
  ...
    "invocation": {
      "configSource": {
        "uri": "git+https://github.com/laurentsimon/provenance-npm-test@refs/heads/main",
       ...
        "entryPoint": "laurentsimon/provenance-npm-test/.github/workflows/release.yml@refs/heads/main"
      },
    ...
    "materials": [
      {
        "uri": "git+https://github.com/laurentsimon/provenance-npm-test",
       ...
      }
    ]
```

In all the fields, the URI = `git+https://github.com/repo/name@<ref>`. In the material, the `<ref>` is not present.
Is this intentional?

In the https://github.com/slsa-framework/slsa-github-generator repo, we always report the `ref`.

PR #521, opened 2023-03-09
feat: verify sourceURI for npm packages

laurentsimon:
closes https://github.com/slsa-framework/slsa-verifier/issues/492

### DIFF
--- a/verifiers/internal/gcb/provenance.go
+++ b/verifiers/internal/gcb/provenance.go
@@ -350,7 +350,7 @@ func (p *Provenance) VerifySourceURI(expectedSourceURI string, builderID utils.T
 	}
 	uri := materials[0].URI
 	// NOTE: the material URI did not contain 'git+' for GCB versions <= v0.3.
-	// A change occured sometimes in v0.3 witout version bump.
+	// A change occurred sometimes in v0.3 witout version bump.
 	// Versions >= 0.3 contain the prefix (https://github.com/slsa-framework/slsa-verifier/pull/519).
 	uri = strings.TrimPrefix(uri, "git+")
 
--- a/verifiers/internal/gcb/provenance.go
+++ b/verifiers/internal/gcb/provenance.go
@@ -349,6 +349,8 @@ func (p *Provenance) VerifySourceURI(expectedSourceURI string, builderID utils.T
 		return fmt.Errorf("%w: no materials", serrors.ErrorInvalidDssePayload)
 	}
 	uri := materials[0].URI
+	// NOTE: the material URI did not contain 'git+' for GCB versions <= 0.3.
+	// Versions >= 0.3 contain the prefix (https://github.com/slsa-framework/slsa-verifier/pull/519).
 	uri = strings.TrimPrefix(uri, "git+")
 
 	// It is possible that GCS builds at level 2 use GCS sources, prefixed by gs://.
--- a/verifiers/internal/gcb/provenance.go
+++ b/verifiers/internal/gcb/provenance.go
@@ -349,7 +349,8 @@ func (p *Provenance) VerifySourceURI(expectedSourceURI string, builderID utils.T
 		return fmt.Errorf("%w: no materials", serrors.ErrorInvalidDssePayload)
 	}
 	uri := materials[0].URI
-	// NOTE: the material URI did not contain 'git+' for GCB versions <= 0.3.
+	// NOTE: the material URI did not contain 'git+' for GCB versions <= v0.3.
+	// A change occured sometimes in v0.3 witout version bump.
 	// Versions >= 0.3 contain the prefix (https://github.com/slsa-framework/slsa-verifier/pull/519).
 	uri = strings.TrimPrefix(uri, "git+")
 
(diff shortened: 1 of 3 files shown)


verifiers/internal/gha/provenance.go
```diff
@@ -123,13 +125,17 @@ func verifySourceURI(prov slsaprovenance.Provenance, expectedSourceURI string, v
 	return nil
 }
 
-func sourceFromURI(uri string) (string, error) {
+// NOTE: `allowNoRef` is to allow for verification of npm packages
```
asraa, line 128: should we set a tracking issue to remove this after GA?

asraa (approved):

verifiers/internal/gcb/provenance.go
```diff
@@ -349,6 +349,8 @@ func (p *Provenance) VerifySourceURI(expectedSourceURI string, builderID utils.T
 		return fmt.Errorf("%w: no materials", serrors.ErrorInvalidDssePayload)
 	}
 	uri := materials[0].URI
+	// NOTE: the material URI did not contain 'git+' for GCB versions <= 0.3.
```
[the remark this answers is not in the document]
laurentsimon: I meant the builder GoogleHostedWorker@v0.3 started adding the `git+`, agreed it's not versioned.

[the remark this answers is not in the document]
asraa: I don't think this is true either - we have some existing 0.3 cases without it, the git+ fix went out just last week

verifiers/internal/gha/provenance.go
laurentsimon: good idea. I was hesitant because I don't know that we'll be able to remove it right away, since some provenance will have been generated with this code. Unless all the provenance generated before GA is considered "no prod ready" and the npm teams removes it. Cerated https://github.com/slsa-framework/slsa-verifier/issues/524

verifiers/internal/gcb/provenance.go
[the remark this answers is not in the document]
laurentsimon: I meant the changes has happened in the middle of v0.3, I wanted to have a comment for it. Let me update to make it clearer.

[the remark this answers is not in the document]
laurentsimon: ptal at the updated comment

verifiers/internal/gha/provenance.go
```diff
@@ -114,6 +110,12 @@ func verifySourceURI(prov slsaprovenance.Provenance, expectedSourceURI string, v
 
 	// Last, verify that both fields match.
 	// We use the full URI to match on the tag as well.
+	if allowNoMaterialRef && len(strings.Split(materialSourceURI, "@")) == 1 {
+		// NOTE: this is an exception for npm packages built before GA,
+		// see https://github.com/gh-community/npm-provenance-private-beta-community/issues/8.
```
ianlewis, line 115: It's worth noting these links aren't accessible to the general public. We probably want to create our own issue if we want to have some public info we can link to.

verifiers/internal/gha/provenance.go
ianlewis: We can maybe only allow it for provenance created before a certain date? or something like that?

ianlewis, line 128:
```suggestion
// sourceFromURI retrieves the source repository given a repository URI with ref.
//
// NOTE: `allowNoRef` is to allow for verification of npm packages
```

verifiers/internal/gha/provenance.go
```diff
@@ -123,13 +125,17 @@ func verifySourceURI(prov slsaprovenance.Provenance, expectedSourceURI string, v
...
 	if uri == "" {
 		return "", fmt.Errorf("%w: empty uri", serrors.ErrorMalformedURI)
 	}
 
 	r := strings.SplitN(uri, "@", 2)
-	if len(r) < 2 {
+	if len(r) < 2 && !allowNoRef {
 		return "", fmt.Errorf("%w: %s", serrors.ErrorMalformedURI,
 			uri)
 	}
```
ianlewis, line 141: BTW, I don't think that `len(r) < 1` (below) is possible with `strings.SplitN`.
https://go.dev/play/p/jnxHlLHbsS0

ianlewis (approved):

verifiers/internal/gha/provenance.go
asraa: Yeah, I wonder about the #524, and we can ask them when they go GA - i expect that is when they will stabilize

verifiers/internal/gcb/provenance.go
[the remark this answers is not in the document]
asraa: sg! sorry for the back and forth

asraa (approved):

verifiers/internal/gha/provenance.go
laurentsimon: Good catch. I changed the link to https://github.com/slsa-framework/slsa-verifier/issues/492

Issue #526, opened 2023-03-10
Add a file extension `.exe` to Windows artifacts


suzuki-shunsuke:
- https://github.com/slsa-framework/slsa-verifier/blob/5a77b25fb4e7e0490847dbbd3ce6cddf479b15da/.slsa-goreleaser/windows-amd64.yml#L13
- https://github.com/slsa-framework/slsa-verifier/blob/5a77b25fb4e7e0490847dbbd3ce6cddf479b15da/.slsa-goreleaser/windows-arm64.yml#L13

Currently, Windows artifacts has no file extension.
In Windows, executable files should have file extensions.
They should have file extensions `.exe`.

ref. https://superuser.com/questions/228680/on-windows-what-filename-extensions-denote-an-executable

ianlewis:
#572 addressed this so let's close.

PR #527, opened 2023-03-10
chore: add a file extension ".exe" to Windows artifacts

suzuki-shunsuke:
- https://github.com/slsa-framework/slsa-verifier/issues/526


asraa (approved):
Thank you so much for both creating the issue and the fix @suzuki-shunsuke !

PR #530, opened 2023-03-14
chore: update docs for release v2.1.0

asraa:
#label:release v2.1.0

PR #532, opened 2023-03-17
fix: Add checksum for v2.1.0

ianlewis:


ianlewis:
THis is solved in #530

PR #541, opened 2023-03-23
docs: Update README.md

drewroengoogle:
Adding an alternative option for installing slsa-verifier if you do not rely on additional tooling. The benefit of this option is improved readability.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,11 @@ $ go mod tidy
 $ cd tooling
 $ grep _ tooling_test.go | cut -f2 -d '"' | xargs -n1 -t go install
 ``` 
+Alternatively, if your project does not rely on additional tools and only uses slsa-verifier, you can instead run the following commands:
+```bash
+$ cd tooling
+$ go install github.com/slsa-framework/slsa-verifier/v2/cli/slsa-verifier
+``` 
 
 #### Option 2: Compile manually
 


README.md
```diff
@@ -154,6 +154,11 @@ $ go mod tidy
 $ cd tooling
 $ grep _ tooling_test.go | cut -f2 -d '"' | xargs -n1 -t go install
 ``` 
+Alternatively, if your project does not rely on additional tools and only uses slsa-verifier, you can instead run the following commands:
+```bash
+$ cd tooling
+$ go install github.com/slsa-framework/slsa-verifier/v2/cli/slsa-verifier
```
ianlewis, line 160: Thanks! We do already have a note on how to install at the beginning of this section, around line 126. Maybe  it's hard to pick out and should get it's own section?

We'd also like folks to install a release rather than from HEAD.

laurentsimon: I think he means that you can install this way *with* the go.mod, instead of using `grep _ ...` command.
The install will ick up the version defined in the go.mod. This is more readable if there's a single tool, ie just the slsa-verifier to install

drewroengoogle: Thanks for taking a look! I was trying to convey that you can replace the `grep _ tooling_test.go ...` line with `go install github.com/slsa-framework/slsa-verifier ...` in your CI and still get the release version that is set in `go.mod`. I figured this might be a nice alternative for increased readability of what is happening in your CI, but let me know if there's a better way to phrase it. If you think it is too much of a duplicate of the installation option above, no worries :)

laurentsimon: Maybe we should add a section:
```
Option 1: Compile locally
Option 2: Commpile in CI
...
```

@ianlewis wdut?

laurentsimon: The example we give separates the tooling.go. I think it's a feature so that we don't add dependencies to the main project. At least that's how I view it

drewroengoogle: > Maybe we should add a section:
> 
> ```
> Option 1: Compile locally
> Option 2: Compile in CI
> ...
> ```
> 
> wdut?

I can add it in this PR if it makes sense, let me know what you all think!

laurentsimon: @ianlewis any more though on this? Or shall we merge as-in?

laurentsimon (approved):

README.md
ianlewis: I guess I'm ok with it.

ianlewis (approved):

PR #543, opened 2023-03-24
chore: report scheduled release workflow failures

asraa:
While I get email notifications for failures, the failures aren't reported for the daily runs of the verifier release workflow.

* Adds issue reporting for release workflow failures
* Fixes an issue where only non-scheduled runs should be verifying the release assets


laurentsimon (approved):
Thanks!

Issue #538, opened 2023-03-20
doc: ToC appears twice in README


laurentsimon:
Also, the Action should not be listed under "Compile from source" installation

PR #557, opened 2023-04-13
docs: remove duplicated table of contents

asraa:
Fixes https://github.com/slsa-framework/slsa-verifier/issues/538


laurentsimon (approved):
Good catch!

PR #562, opened 2023-04-18
feat: support for BYOB verification

laurentsimon:
CLI tests will need to be added once we have a non RC release.
I have made the builder-id a mandatory option, to ensure there is no bias towards GitHub as default trusted builders.

Will need a follow-up PR to add exception to allow certain workflows to be referenced at main for testing.


verifiers/internal/gha/provenance.go
```diff
@@ -41,18 +41,38 @@ func EnvelopeFromBytes(payload []byte) (env *dsselib.Envelope, err error) {
...
+	}
+	if err := provBuilderID.MatchesFull(expectedBuilderID, true); err != nil {
+		return fmt.Errorf("%w", err)
 	}
+	return nil
+}
 
+// Verify Builder ID in provenance statement.
+// This function verifies the names match. If the expected builder ID contains a version,
+// it also evrifies the verisons match.
```
ianlewis, line 63:
```suggestion
// it also verifies the verisons match.
```

verifiers/internal/gha/npm.go
```diff
@@ -21,6 +21,8 @@ import (
 
 const (
 	publishAttestationV01 = "https://github.com/npm/attestation/tree/main/specs/publish/"
+	builderGitHubRunnerID = "https://github.com/actions/runner"
+	ossfNpmBuilderID      = "https://github.com/slsa-framework/slsa-github-generator/.github/workflows/builder_nodejs_slsa3.yml"
```
ianlewis, line 25: I guess you don't actually need this anymore since the builder-id option is required?

ianlewis (approved):

verifiers/internal/gha/npm.go
laurentsimon: Good catch. You're correct, I'll remove it.

PR #587, opened 2023-05-09
docs: Add docs for npm package verification

### LINKED ISSUE #574: Docs for npm package verification

ianlewis:
Fixes #574


laurentsimon (approved):

README.md
```diff
@@ -320,6 +321,31 @@ The verified in-toto statement may be written to stdout with the `--print-proven
...
+curl -Sso attestations.json $(npm view @ianlewis/actions-test --json | jq -r '.dist.attestations.url') && \
+curl -SsO "$(npm view @ianlewis/actions-test --json | jq -r '.dist.tarball')"
+```
+
+You can then verify the package by running the following command:
+
+```shell
+$ SLSA_VERIFIER_EXPERIMENTAL=1 slsa-verifier verify-npm-package actions-test-0.1.126.tgz \
+  --attestations-path attestations.json \
+  --builder-id "https://github.com/slsa-framework/slsa-github-generator/.github/workflows/builder_nodejs_slsa3.yml@refs/tags/v1.6.0" \
```
laurentsimon, line 339: I think here we can omit the `@<ref>`. It scales better if we ask users to verify with the builder ID *without* its version. Btw `@v1.6.0` should work too and is simpler for users, I think.

README.md
```diff
@@ -320,6 +321,31 @@ The verified in-toto statement may be written to stdout with the `--print-proven
...
+  --package-name "@ianlewis/actions-test" \
+  --package-version 0.1.126 \
+  --source-uri github.com/ianlewis/actions-test
+```
+
+The verified in-toto statement may be written to stdout with the `--print-provenance` flag to pipe into policy engines.
+
+Only GitHub URIs are supported with the `--source-uri` flag. A tag should not be specified, even if the provenance was built at some tag. If you intend to do source versioning validation, use `--print-provenance` and inspect the commit SHA of the config source or materials.
```
laurentsimon, line 347: re: source versioning. They don't need to use `--print-provenance` for that. They can use `--source-tag`

README.md
```diff
@@ -320,6 +321,31 @@ The verified in-toto statement may be written to stdout with the `--print-proven
 
 Note that `--source-uri` supports GitHub repository URIs like `github.com/$OWNER/$REPO` when the build was enabled with a Cloud Build [GitHub trigger](https://cloud.google.com/build/docs/automating-builds/github/build-repos-from-github). Otherwise, the build provenance will contain the name of the Cloud Storage bucket used to host the source files, usually of the form `gs://[PROJECT_ID]_cloudbuild/source` (see [Running build](https://cloud.google.com/build/docs/running-builds/submit-build-via-cli-api#running_builds)). We recommend using GitHub triggers in order to preserve the source provenance and valiate that the source came from an expected, version-controlled repository. You _may_ match on the fully-qualified tar like `gs://[PROJECT_ID]_cloudbuild/source/1665165360.279777-955d1904741e4bbeb3461080299e929a.tgz`.
 
+## Verification for npm packages
+
+Verification of npm packages is currently an experimental feature. To verify a
```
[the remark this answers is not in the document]
laurentsimon: I took a stab at it.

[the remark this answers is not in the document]
ianlewis: I'll add a note about the supported builders.

README.md
ianlewis: Yeah, I wasn't sure about how the builder-id matched. I just used verbatim the value that was set in provenance.

README.md
ianlewis: This was a copy and paste from earlier in the document. I think maybe the other sections are also out of date?

laurentsimon: ah yeah, let's update them as well.

ianlewis: Since git tags are mutable, how good is `source-tag` for validation anyway?

laurentsimon: `source-tag` is optional, but it's verified thru the GitHub event data, so it's accurate. Someone can re-tag the source later, but that won't change the original tag it was built from on the GitHu event.

ianlewis: Right. It's basically just verifying the tag that it was built against at the time. I'm wondering though if that's what's meant by "source versioning validation" here. I think tags are more of a "release versioning" type of thing.

Issue #605, opened 2023-05-19
[feature][byob] Verification support v1.0


laurentsimon:
Need to add support for verify-artifact command

PR #609, opened 2023-05-22
feat: Support for v1.0 verification in BYOB

laurentsimon:
closes https://github.com/slsa-framework/slsa-verifier/issues/605

This builds upon https://github.com/slsa-framework/slsa-verifier/pull/604 which needs to be merged first.


laurentsimon:
@asraa have a look at the regression test failure. I think I'm going to comment out the test and you'll need to update the docker-based builder to output the right v1.0 format. Does that sounds right? (I'll create a tracking issue if so)

laurentsimon:
> @asraa have a look at the regression test failure. I think I'm going to comment out the test and you'll need to update the docker-based builder to output the right v1.0 format. Does that sounds right? (I'll create a tracking issue if so).
> 
> I know that the "workflow" is not part of the builder interface (see https://github.com/slsa-framework/github-actions-buildtypes/tree/main/workflow/v1), but it seems nice to have for troubleshooting purposes: it allows someone to determine where the build started.

Created https://github.com/slsa-framework/slsa-verifier/issues/610 for tracking. /cc @rbehjati

ianlewis (approved):

Issue #472, opened 2023-02-09
[v1] Implement SLSA v1.0 checks against tag and branch options


asraa:
Follow up from https://github.com/slsa-framework/slsa-verifier/pull/470

I am not entirely sure of the SLSA v1.0 format for GH environment, so waiting w/ unimplemented methods

PR #618, opened 2023-05-25
feat: Support different source branch/tag from trigger branch/tag.

ianlewis:
- Updates handling of github environment variables such that they can be uppercase or lowercase.
- Pulls source branch/tag from the dependencies URI (`materials` or `resolvedDependencies`)

This means that `--source-branch` and `--source-tag` will not work if provided because the ref is not included in the source dependency URI if a different sha is provided from the triggering sha.

### DIFF
--- a/verifiers/internal/gha/slsaprovenance/v1.0/provenance.go
+++ b/verifiers/internal/gha/slsaprovenance/v1.0/provenance.go
@@ -2,6 +2,7 @@ package v1
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	intoto "github.com/in-toto/in-toto-golang/in_toto"
@@ -112,21 +113,33 @@ func (prov *ProvenanceV1) Subjects() ([]intoto.Subject, error) {
 }
 
 func (prov *ProvenanceV1) GetBranch() (string, error) {
-	// TODO(https://github.com/slsa-framework/slsa-verifier/issues/472): Add GetBranch() support.
-	sysParams, ok := prov.Predicate.BuildDefinition.InternalParameters.(map[string]interface{})
-	if !ok {
-		return "", fmt.Errorf("%w: %s", serrors.ErrorInvalidDssePayload, "system parameters type")
+	// Returns the branch from the source URI.
+	sourceURI, err := prov.SourceURI()
+	if err != nil {
+		return "", err
 	}
 
-	return slsaprovenance.GetBranch(sysParams, prov.predicateType)
+	parts := strings.Split(sourceURI, "@")
+	if len(parts) > 1 && strings.HasPrefix(parts[1], "refs/heads") {
+		return parts[1], nil
+	}
+
+	return "", nil
 }
 
 func (prov *ProvenanceV1) GetTag() (string, error) {
-	sysParams, ok := prov.Predicate.BuildDefinition.InternalParameters.(map[string]interface{})
-	if !ok {
-		return "", fmt.Errorf("%w: %s", serrors.ErrorInvalidDssePayload, "system parameters type")
+	// Returns the tag from the source materials.
+	sourceURI, err := prov.SourceURI()
+	if err != nil {
+		return "", err
 	}
-	return slsaprovenance.GetTag(sysParams, prov.predicateType)
+
+	parts := strings.Split(sourceURI, "@")
+	if len(parts) > 1 && strings.HasPrefix(parts[1], "refs/tags") {
+		return parts[1], nil
+	}
+
+	return "", nil
 }
 
 func (prov *ProvenanceV1) GetWorkflowInputs() (map[string]interface{}, error) {
@@ -134,7 +147,7 @@ func (prov *ProvenanceV1) GetWorkflowInputs() (map[string]interface{}, error) {
 	if !ok {
 		return nil, fmt.Errorf("%w: %s", serrors.ErrorInvalidDssePayload, "system parameters type")
 	}
-	return slsaprovenance.GetWorkflowInputs(sysParams, prov.predicateType)
+	return slsaprovenance.GetWorkflowInputs(sysParams)
 }
 
 // TODO(https://github.com/slsa-framework/slsa-verifier/issues/566):
--- a/verifiers/internal/gha/provenance.go
+++ b/verifiers/internal/gha/provenance.go
@@ -160,12 +160,12 @@ func sourceFromURI(uri string, allowNoRef bool) (string, error) {
 
 	r := strings.Split(uri, "@")
 	if len(r) < 2 && !allowNoRef {
-		return "", fmt.Errorf("%w: %s", serrors.ErrorMalformedURI,
+		return "", fmt.Errorf("%w: ref required: %s", serrors.ErrorMalformedURI,
 			uri)
 	}
 	if len(r) < 1 {
-		return "", fmt.Errorf("%w: %s", serrors.ErrorMalformedURI,
-			uri)
+		// Split should always return a slice with length > 0
+		panic(fmt.Errorf("%w: %s", serrors.ErrorMalformedURI, uri))
 	}
 
 	return r[0], nil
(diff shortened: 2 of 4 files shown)


verifiers/internal/gha/slsaprovenance/v1.0/provenance.go
```diff
@@ -112,29 +113,41 @@ func (prov *ProvenanceV1) Subjects() ([]intoto.Subject, error) {
...
+	if len(parts) > 1 && strings.HasPrefix(parts[1], "refs/heads") {
+		return parts[1], nil
 	}
 
-	return slsaprovenance.GetBranch(sysParams, prov.predicateType)
+	return "", nil
 }
 
 func (prov *ProvenanceV1) GetTag() (string, error) {
-	sysParams, ok := prov.Predicate.BuildDefinition.InternalParameters.(map[string]interface{})
```
laurentsimon, line 125: do we need a TODO here for https://github.com/slsa-framework/slsa-verifier/issues/472?

laurentsimon (approved):
Can you add some unit tests?

verifiers/internal/gha/slsaprovenance/v1.0/provenance.go
```diff
@@ -112,29 +113,41 @@ func (prov *ProvenanceV1) Subjects() ([]intoto.Subject, error) {
 }
 
 func (prov *ProvenanceV1) GetBranch() (string, error) {
-	// TODO(https://github.com/slsa-framework/slsa-verifier/issues/472): Add GetBranch() support.
```
laurentsimon, line 115: so this will remove support for all cases, not just the cases where checkout sha != github event sha. Correct? Would this not be a breaking change? (same question for getTag()).

ianlewis: I think in all cases the "most correct" source branch and source tag will come from the materials/resolvedDependencies. My understanding is that up until we added the `sha1` input all of our builders have provided the same value in the parameters as in the dependencies. So the value returned here shouldn't change in that case.

So I believe that it shouldn't break older provenance, but I suppose that technically we didn't actually check that the branch and sha were the same between the source and trigger.

I suppose the alternative is to get from `InternalParameters` if the shas are the same and the source uri if they are different but I'm not sure if we gain anything by doing that. I'll look a bit into what the older versions of the builders did.

verifiers/internal/gha/slsaprovenance/v1.0/provenance.go
ianlewis: #472 is closed and I wasn't sure what it was referring to in this case. I removed the TODO above because it was closed.

Do you think we should re-open it?

verifiers/internal/gha/slsaprovenance/v1.0/provenance.go
laurentsimon: branch and tag info is not present in materials / resolvedDeps. I think it's only present via the GH context?

verifiers/internal/gha/slsaprovenance/v1.0/provenance.go
laurentsimon: iiuc, this PR removes support for branch / tag for v1?

verifiers/internal/gha/slsaprovenance/v1.0/provenance.go
ianlewis: There is a branch/tag ref on the URI in the materials/resolvedDeps. We explicitly remove it when sha1 is specified.

verifiers/internal/gha/slsaprovenance/v1.0/provenance.go
ianlewis: branch / tag verification won't work for v1 if the trw specifies the a different sha than the trigger sha but it should still work otherwise. At least that's my intention.

verifiers/internal/gha/slsaprovenance/v1.0/byob.go
```diff
@@ -0,0 +1,265 @@
...
+	// Returns the branch from the source URI.
+	sourceURI, err := p.SourceURI()
+	if err != nil {
+		// Get the value from the internalParameters if there is no source URI.
+		sysParams, ok := p.prov.Predicate.BuildDefinition.InternalParameters.(map[string]interface{})
+		if !ok {
+			return "", fmt.Errorf("%w: %s", serrors.ErrorInvalidDssePayload, "internal parameters type")
+		}
+		return common.GetBranch(sysParams, slsa1.PredicateSLSAProvenance)
+	}
```
ianlewis, line 174: NOTE that I fall back to the previous behavior of getting the branch from the `internalParameters`.

Some tests seem to have test data that doesn't include a `resolvedDependencies` so this is here to make those work but I'm not sure that there are real cases where provenance wouldn't have it.

verifiers/internal/gha/slsaprovenance/v1.0/byob.go
```diff
@@ -0,0 +1,265 @@
...
+var (
+	// byobBuildType is the base build type for BYOB delegated builders.
+	byobBuildType = "https://github.com/slsa-framework/slsa-github-generator/delegator-generic@v0"
+
+	// containerBasedBuildType is the build type for the container-based builder and is based on BYOB.
+	containerBasedBuildType = "https://slsa.dev/container-based-build/v0.1?draft"
+
+	// genericGHABuildType is used by some tests.
+	// TODO: Update tests to use a real buildType
+	genericGHABuildType = "https://github.com/Attestations/GitHubActionsWorkflow@v1"
```
ianlewis, line 24: I wonder if we need to support this buildType or if this was just used for tests.

laurentsimon: i think it's an old type we dont use, and probably can remove it

verifiers/internal/gha/slsaprovenance/v1.0/byob.go
```diff
@@ -0,0 +1,268 @@
...
+		return parts[1], nil
+	}
+
+	return "", nil
+}
+
+// GetTag implements Provenance.GetTag.
+func (p *BYOBProvenanceV1) GetTag() (string, error) {
+	// Returns the tag from the source materials.
+	sourceURI, err := p.SourceURI()
```
laurentsimon, line 187: same comment / question as for GetBranch

verifiers/internal/gha/slsaprovenance/v0.2/byob.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// BuilderID implements Provenance.BuilderID.
+func (p *BYOBProvenanceV02) BuilderID() (string, error) {
+	return p.prov.Predicate.Builder.ID, nil
+}
+
+// SourceURI implements Provenance.SourceURI.
+func (p *BYOBProvenanceV02) SourceURI() (string, error) {
+	if len(p.prov.Predicate.Materials) == 0 {
```
laurentsimon, line 35: I'm wondering whether our existing code was wrong and the intention was to verify against `sourceConfig`...?
Not very important at this point...

verifiers/internal/gha/slsaprovenance/v0.2/generic.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// BuilderID implements Provenance.BuilderID.
+func (p *GenericProvenanceV02) BuilderID() (string, error) {
+	return p.prov.Predicate.Builder.ID, nil
+}
+
+// SourceURI implements Provenance.SourceURI.
+func (p *GenericProvenanceV02) SourceURI() (string, error) {
+	if len(p.prov.Predicate.Materials) == 0 {
```
laurentsimon, line 48: is there a way we could share all these function with a "base structure", and only re-wrrite those that need changes?

laurentsimon (commented):
maybe it helps to split this PR in into 2:
- re-factoring of code
- changes for sha1 support wrt tag /branch

verifiers/internal/gha/slsaprovenance/v1.0/byob.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+// BuilderID implements Provenance.BuilderID.
+func (p *BYOBProvenanceV1) BuilderID() (string, error) {
+	return p.prov.Predicate.RunDetails.Builder.ID, nil
+}
+
+// SourceURI implements Provenance.SourceURI.
+func (p *BYOBProvenanceV1) SourceURI() (string, error) {
+	// Use resolvedDependencies.
```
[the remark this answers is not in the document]
ianlewis: Right, maybe I can just update this to include the refactor and hold off on the changes for the sha1 support since the uri & branch/tag will likely need to come from that rather than from `resolvedDependencies`.

ianlewis:
I think I'll focus on some refactors for a bit which I can break into separate PRs. Will keep this open until I am able to break it up.

PR #621, opened 2023-05-26
fix: use ExternalParameters["source"] for the Source URI for SLSA v1.0 provenance

asraa:
See https://github.com/slsa-framework/slsa-github-generator/issues/2182

* Use externalParameters["source"] for detecting source URI in slsa v1.0 provenance

This effectively fixes the container-based builder e2e tests in conjunction with the generator update. (tested)

closes https://github.com/slsa-framework/slsa-verifier/issues/623


laurentsimon (commented):
same question as for the generator PR. The GH buildType explicitly says that the first resolvedDependencies is the one for the source. So I wonder if we need this annotation in practice.

laurentsimon (approved):
Thanks.

PR #628, opened 2023-06-01
refactor: Provenance tests

ianlewis:
Refactors GHA provenance tests to use `testProvenance` which makes it clearer what is actually being tested. This will also make it easier to support `buildType` as a way to have different verification logic as the tests no longer rely on testdata with the `"https://github.com/Attestations/GitHubActionsWorkflow@v1"` build type, which isn't used by any supported builders.

A couple of updates to utilities:
- `VerifyTag` will now validate the ref returned by the `Provenance` instance.
- `VerifyBranch` will now validate the ref returned by the `Provenance` instance.
- `VerifyDigest` now supports the 160 bit `"sha1"` algo (FWIW) and will now search all subject entries even if one subject entry's algorithm does not match the expected algorithm.


verifiers/internal/gha/provenance.go
```diff
@@ -179,12 +179,18 @@ func verifyDigest(prov slsaprovenance.Provenance, expectedHash string) error {
 	}
 
 	// 8 bit represented in hex, so 8/2=4.
-	l := len(expectedHash) * 4
+	bitLength := len(expectedHash) * 4
+	expectedAlgo := fmt.Sprintf("sha%v", bitLength)
+	// sha1 is 160 bit (FWIW).
```
laurentsimon, line 184: curious: why do we want to support this algo (which is not considered secure)?

ianlewis: Well, the alternative is to support "sha160" which isn't a name anyone uses. Should we support a minimum bit length?

ianlewis: Added an issue and TODO comment to track: https://github.com/slsa-framework/slsa-verifier/issues/630

PR #629, opened 2023-06-01
fix: revert to using resolvedDepdendencies for source verification

asraa:
Reverts the logic in https://github.com/slsa-framework/slsa-verifier/commit/db0560e3285e082f58a6d718a720047bb302fdd9.

For now, we assume a GHA builder verifier contract to detect the source in the first resolvedDependency. Per SLSA discussions, we will likely resolve this with some metadata in the name or annotations to distinguish the dependencies.

* fix: Use the first resolvedDependencies for SLSA v1.0 source verification


laurentsimon (approved):
/cc @ianlewis

PR #635, opened 2023-06-06
fix: builder ID verification for testing

ianlewis:
Related: https://github.com/slsa-framework/slsa-github-generator/issues/2224


verifiers/utils/builder.go
```diff
@@ -138,6 +138,7 @@ func IsValidBuilderTag(ref string, testing bool) error {
 			semver.Build(pin) != "" {
 			return fmt.Errorf("%w: %s: version tag not valid", serrors.ErrorInvalidRef, pin)
 		}
+		return nil
```
asraa, line 141: NICE CATCH! Took me a second myself to figure out why it was going to the non-testing block.

asraa (approved):

PR #638, opened 2023-06-08
test: add tests for v1.7.0 builders

asraa:


laurentsimon (approved):

cli/slsa-verifier/main_regression_test.go
```diff
@@ -74,6 +74,10 @@ func getBuildersAndVersions(t *testing.T,
 }
 
 func Test_runVerifyGHAArtifactPath(t *testing.T) {
+	// We cannot use t.Setenv due to parallelized tests.
```
ianlewis, line 77: Could we just not run the test in parallel?

asraa: This one in particular takes a very long time since it scales across the builders and vers and makes online requests to Rekor - I'll try it out, but I think the best solution is just of get rid of the env var by regenerating the test.

ianlewis: I'm not sure how to specify it, but if we could run these tests (the test cases run by `t.Run`) in parallel but not concurrently with other top-level tests then it should work.

PR #641, opened 2023-06-16
feat: Verification for when sha1 is specified in BYOB TRW

### LINKED ISSUE #600: [feature][byob] Verification for sha1 provided by TRW for v1.0
verification counterpart for https://github.com/slsa-framework/slsa-github-generator/issues/2137

Make sure to follow the comments in https://github.com/slsa-framework/slsa-verifier/issues/599 and differentiate between generators and builders

ianlewis:
Fixes #600


verifiers/utils/git.go
```diff
@@ -7,27 +7,66 @@ import (
 	serrors "github.com/slsa-framework/slsa-verifier/v2/errors"
 )
 
-// ParseGitRef validates that the given git ref is a valid ref of the given type and returns its name.
-func ParseGitRef(refType, ref string) (string, error) {
-	refPrefix := fmt.Sprintf("refs/%s/", refType)
-	if !strings.HasPrefix(ref, refPrefix) {
-		return "", fmt.Errorf("%w: %s: not of the form '%s<name>'", serrors.ErrorInvalidRef, ref, refPrefix)
+// NormalizeGitURI normalizes a git URI to include a git+https:// prefix.
+func NormalizeGitURI(s string) string {
```
ianlewis, line 11: Add some tests for these functions.

verifiers/internal/gha/provenance.go
```diff
@@ -76,100 +76,84 @@ func verifyBuilderIDLooseMatch(prov iface.Provenance, expectedBuilderID string)
...
-		source = "git+" + source
-	}
-
-	return source
-}
-
 // Verify source URI in provenance statement.
-func verifySourceURI(prov iface.Provenance, expectedSourceURI string, allowNoMaterialRef bool) error {
-	source := asURI(expectedSourceURI)
+func verifySourceURI(prov iface.Provenance, expectedSourceURI string) error {
```
ianlewis, line 80: `--source-branch main` still passes when it should fail. Source branch (and similarly source tag) should be retrieved from the `materials` or `resolvedDependencies`. This will be empty if a sha1 was provided was provided.

ianlewis: Still need support this for SLSA v0.2

verifiers/internal/gha/slsaprovenance/v1.0/container_based.go
```diff
@@ -1,11 +1,188 @@
...
-// ContainerBasedProvenance is provenance generated by the container-based builder.
+	intoto "github.com/in-toto/in-toto-golang/in_toto"
+	slsa1 "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v1"
+	serrors "github.com/slsa-framework/slsa-verifier/v2/errors"
+
+	"github.com/slsa-framework/slsa-verifier/v2/verifiers/internal/gha/slsaprovenance/common"
+)
+
+// ContainerBasedProvenance is SLSA v1.0 provenance for the slsa-github-generator BYOB build type.
 type ContainerBasedProvenance struct {
```
ianlewis, line 16: Note to reviewers: ContainerBasedProvenance doesn't support sha1 and only supports source == trigger

verifiers/internal/gha/slsaprovenance/v0.2/provenance.go
```diff
@@ -52,22 +38,19 @@ func New(payload []byte) (iface.Provenance, error) {
...
-			upperEnv: true,
-			prov:     a,
-		}, nil
-	case a.Predicate.BuildType == goBuilderBuildType ||
-		a.Predicate.BuildType == genericGeneratorBuildType ||
-		a.Predicate.BuildType == containerGeneratorBuildType ||
-		a.Predicate.BuildType == npmCLIBuildType ||
-		a.Predicate.BuildType == legacyBuilderBuildType ||
-		a.Predicate.BuildType == legacyGoBuilderBuildType:
+	case a.Predicate.BuildType == common.BYOBBuildTypeV0:
```
laurentsimon, line 41: Same comment as above, ie the caller should determine what build IDs are acceptable.

verifiers/internal/gha/slsaprovenance/v1.0/byob.go
```diff
@@ -2,239 +2,92 @@ package v1
...
-	sysParams, ok := p.prov.Predicate.BuildDefinition.InternalParameters.(map[string]interface{})
-	if !ok {
-		return "", fmt.Errorf("%w: %s", serrors.ErrorInvalidDssePayload, "internal parameters type")
-	}
-
-	return common.GetBranch(sysParams, true)
 }
 
 // GetTag implements Provenance.GetTag.
 func (p *BYOBProvenance) GetTag() (string, error) {
```
laurentsimon, line 59: where do we check whether sha1 != GH event sha1? Maybe in a follow-up PR?

verifiers/internal/gha/slsaprovenance/v1.0/byob_test.go
```diff
@@ -0,0 +1,259 @@
...
+			name: "internalParameters uri @ ref/tags/v1.0.0",
+			prov: BYOBProvenance{
+				provenanceV1: &provenanceV1{
+					prov: &intotoAttestation{
+						StatementHeader: intoto.StatementHeader{},
+						Predicate: slsa1.ProvenancePredicate{
+							BuildDefinition: slsa1.ProvenanceBuildDefinition{
+								InternalParameters: map[string]interface{}{
+									"GITHUB_REF_TYPE": "tag",
+									"GITHUB_REF":      "refs/tags/v1.0.0",
```
laurentsimon, line 234: i dont remember why we need to fall back to `GITHUB_REF`. Why is reliance on resolvedDep only not correct?

verifiers/utils/git_test.go
```diff
@@ -0,0 +1,244 @@
...
+func Test_NormalizeGitURI(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		name     string
+		uri      string
+		expected string
+	}{
+		{
+			name:     "empty uri",
```
laurentsimon, line 20: do we need a test with a `@ref` as well? Or it's only relevant for `ParseGitURIAndRef`?

verifiers/internal/gha/slsaprovenance/v1.0/byob.go
ianlewis: We don't actually do any sha checks since we don't verify the specific sha against a user provided value yet. The only change to make really was to fail if the user expects a specific branch because when sha1 != GH sha we can't actually determine that.

verifiers/internal/gha/slsaprovenance/v1.0/byob_test.go
ianlewis: This was mostly to ensure backwards compatibility with the previous code. I kept it mostly to be safe, but I agree that we shouldn't need to keep it.

verifiers/internal/gha/provenance.go
```diff
@@ -76,100 +76,84 @@ func verifyBuilderIDLooseMatch(prov iface.Provenance, expectedBuilderID string)
...
-		return nil
+		if buildType == common.NpmCLIBuildTypeV1 {
+			return nil
+		}
+		// NOTE: BYOB builders can build from a different ref than the triggering ref.
+		// This most often happens when a TRW makes a commit as part of the release process.
+		// NOTE: Currently only building from a different git sha is supported
+		// which means the sourceRef is empty. Building from an arbitrary ref
+		// is currently not supported.
+		if buildType == common.BYOBBuildTypeV0 {
```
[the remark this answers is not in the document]
ianlewis: The `buildType` isn't really the thing that we trust. The thing that the user puts their trust in is the builder ID no?

I think I may agree that we can be strict and expect only that certain `buildTypes` will be generated by certain builders.

verifiers/internal/gha/provenance_test.go
```diff
@@ -322,14 +326,37 @@ func Test_verifySourceURI(t *testing.T) {
...
+			provMaterialsURI:  "git+https://github.com/some/repo",
+			expectedSourceURI: "https://github.com/some/repo",
+		},
+		{
+			name:              "mismatch source material ref (byob)",
+			provBuildType:     common.BYOBBuildTypeV0,
+			provTriggerURI:    "git+https://github.com/some/repo@v1.2.3",
+			provMaterialsURI:  "git+https://github.com/some/repo@v1.2.4",
+			expectedSourceURI: "https://github.com/some/repo",
+			err:               serrors.ErrorInvalidDssePayload,
```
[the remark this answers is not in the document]
ianlewis: We already have a test that verifies that `VerifyBranch` fails when the branch or ref is empty (i.e. when the provenance implementation returns an empty string).

I added a test to `BYOBProvenance` to test when the resolved dependency has no ref.

I guess the question is whether we want to allow something like `--source-branch ""` to pass in this case where `sha1` is different. Right now we don't allow it because the `Provenance` implementations (BYOB, etc.) return an error if the resolved dependencies don't have a ref, but we could just as easily allow it and return an empty string instead.

verifiers/utils/git_test.go
ianlewis: It's not really relevant in this case since the it doesn't look at the ref, but I added the tests just in case.

verifiers/internal/gha/provenance_test.go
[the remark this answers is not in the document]
ianlewis: I think I'll leave it as it is now since that's the closest to the current behavior. We can revisit it in the future if needed.

ianlewis:
Still a few things not working...

This PR is getting too big so I'll see if I can't break it up into smaller PRs.

verifiers/internal/gha/slsaprovenance/v1.0/byob.go
laurentsimon: gotcha. Thanks!

verifiers/internal/gha/slsaprovenance/v1.0/byob_test.go
laurentsimon: since we have not released any v1.0 spec support, do we need backward compatibility? Or can we *not* fall back to using the GITHUB_REF`?

verifiers/internal/gha/provenance.go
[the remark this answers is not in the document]
laurentsimon: In the particular case of the ref verification, it's not too serious. But It'd be better for the caller to indicate whether it's acceptable to ignore the ref verification

verifiers/internal/gha/slsaprovenance/v0.2/provenance.go
laurentsimon: In the part of the code I think the problem could be more serious. If say, another builder is compromised, it could replace the builder ID in the provenance to be of BYOB type, and replace the builder ID with a builder of its choice.

Wdut?

verifiers/internal/gha/slsaprovenance/v1.0/byob_test.go
ianlewis: Well we might still need to do it for v0.2. But you're strictly right that for v1.0 we don't need the backward's compatibility.

laurentsimon: and for v0.2, where do we need it? npm?

Do our regression tests fail if we don't use `GITHUB_REF`?

ianlewis: For npm no, but maybe(?) for others.

laurentsimon: all our builders generate the resolvedDeps, from the first release, afaik

ianlewis: SG, I think we can just remove that code then.

laurentsimon: SG. Trying to not take on technical debt.

laurentsimon (approved):

verifiers/internal/gha/provenance.go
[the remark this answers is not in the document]
ianlewis: Updated to include builder id -> buildType mappings to validate this.

verifiers/internal/gha/slsaprovenance/v1.0/byob_test.go
ianlewis: done.

verifiers/internal/gha/slsaprovenance/v0.2/provenance.go
ianlewis: Added a map of builder ID -> buildType to validate buildTypes.

verifiers/internal/gha/slsaprovenance/v0.2/byob_test.go
```diff
@@ -0,0 +1,236 @@
...
+								"GITHUB_EVENT_NAME": "push",
+								"GITHUB_EVENT_PAYLOAD": map[string]any{
+									"base_ref": nil,
+								},
+							},
+						},
+					},
+				},
+			),
+			err: serrors.ErrorInvalidDssePayload,
```
laurentsimon, line 115: add one tests with `URI: "git+https://github.com/kubernetes/kubernetes"` (ie, no branch in resolvedDeps) and 
```
Environment: map[string]interface{}{
								"GITHUB_REF_TYPE": "branch",
								"GITHUB_REF":      "refs/heads/main",
							},
```

verifiers/internal/gha/slsaprovenance/v0.2/byob_test.go
```diff
@@ -0,0 +1,236 @@
...
+							Environment: map[string]interface{}{
+								"GITHUB_REF_TYPE": "tag",
+								"GITHUB_REF":      "refs/tags/v1.0.0",
+							},
+						},
+					},
+				},
+			),
+			err: serrors.ErrorInvalidDssePayload,
+		},
```
laurentsimon, line 219: add one tests with `URI: "git+https://github.com/kubernetes/kubernetes"` (ie, no branch in resolvedDeps) and 
```json
Environment: map[string]interface{}{
	"GITHUB_REF_TYPE": "tag",
	"GITHUB_REF":      "refs/tags/v1.0.0",
},
```

That's the main common case

verifiers/internal/gha/slsaprovenance/v0.2/byob_test.go
ianlewis: Done.

verifiers/internal/gha/slsaprovenance/v0.2/byob_test.go
ianlewis: Done.

verifiers/internal/gha/slsaprovenance/v0.2/byob_test.go
```diff
@@ -0,0 +1,280 @@
...
+						},
+						Materials: []slsacommon.ProvenanceMaterial{
+							{
+								URI: "git+https://github.com/kubernetes/kubernetes",
+							},
+						},
+					},
+				},
+			),
+			err: serrors.ErrorInvalidDssePayload,
```
laurentsimon, line 137: should return ""?

verifiers/internal/gha/slsaprovenance/v0.2/byob_test.go
```diff
@@ -0,0 +1,280 @@
...
```
laurentsimon, line 262: should return ""?

PR #643, opened 2023-06-19
refactor: Use Go 1.20

### LINKED ISSUE #589: Update to use Go 1.20

ianlewis:
Fixes #589


laurentsimon (approved):

asraa (commented):
Hey @ianlewis - it seems like this may have caused a regression in the workflow - I haven't been following this much, but it seems like it's probably because the go-version-file needs to be prepended with the project checkout dir?

https://github.com/slsa-framework/slsa-verifier/actions/workflows/release.yml

PR #644, opened 2023-06-26
feat: Use tags `vX.Y.Z-<language>` for JReleaser builders

laurentsimon:


verifiers/utils/builder_test.go
```diff
@@ -770,3 +770,68 @@ func Test_IsValidBuilderTag(t *testing.T) {
...
+		{
+			name: "valid semver: no minor",
+			ref:  "refs/tags/v1-java",
+			err:  serrors.ErrorInvalidRef,
+		},
+		{
+			name: "valid semver: no minor",
+			ref:  "refs/tags/v1-java",
+			err:  serrors.ErrorInvalidRef,
+		},
```
ianlewis, line 799: This is repeated?

verifiers/internal/gha/provenance.go
```diff
@@ -292,6 +292,12 @@ func isValidDelegatorBuilderID(prov iface.Provenance) error {
 	if len(parts) != 2 {
 		return fmt.Errorf("%w: %s", serrors.ErrorInvalidBuilderID, id)
 	}
+
+	// Exception for JReleaser builders.
+	// See https://github.com/slsa-framework/slsa-github-generator/issues/2035#issuecomment-1579963802.
+	if strings.HasPrefix(parts[0], JReleaserRepository) {
```
ianlewis, line 298: I'm not sure how I feel about having one-off exceptions in verification. Especially since IIUC this is specific to JReleaser in that the single JReleaser action can work on a Java or Rust repo and we aren't especially likely to see this use case on other builders.

verifiers/utils/builder.go
```diff
@@ -150,3 +150,26 @@ func IsValidBuilderTag(ref string, testing bool) error {
...
+		return err
+	}
+
+	// Valid semver of the form vX.Y.Z-<language> with no metadata.
+	// NOTE: When adding a language, update the corresponding
+	// unit test.
+	languages := map[string]bool{
+		"-java": true,
+	}
+	_, correctLanguage := languages[semver.Prerelease(pin)]
```
ianlewis, line 167: nit: s/correct/valid/

ianlewis (approved):

verifiers/utils/builder_test.go
laurentsimon: thanks. done.

verifiers/internal/gha/provenance.go
laurentsimon: I could not think of another way to handle it. Let me know if you have other ideas.

laurentsimon:
@aalmiray you should be able to verify BYOB provenance with slsa-verifier at main.
You must:
1. reference the delegator at a tag
2. reference the TRW (from PW) at a tag

One more PR is needed to support user-provided sha1.

PR #648, opened 2023-06-29
refactor: Use full builder id

ianlewis:
Internally use full builder IDs including server url rather than worflow ref as a path. This should hopefully avoid confusion between dealing with builder IDs and `GITHUB_WORKFLOW_REF` which only contains the path portion. `GITHUB_WORKFLOW_REF` is the only thing that doesn't include the domain/server url part of the workflow/builder ID. The Fulcio OID claims include the full url.

Code extracted from #641


verifiers/internal/gha/builder.go
```diff
@@ -235,6 +233,32 @@ type WorkflowIdentity struct {
 	Issuer string
 }
 
+// SubjectWorkflowName returns the subject workflow without the git ref.
+func (id *WorkflowIdentity) SubjectWorkflowName() string {
+	withoutRef, err := url.Parse(id.SubjectWorkflow.String())
+	if err != nil {
+		// This should never happen.
+		panic(err)
```
laurentsimon, line 241: avoid panic'ing? Especially when if folks run our code as a service in the future. Wdut? Return "" or an error?

laurentsimon (approved):
Thanks! I agree the builder code was a little hard to read

verifiers/internal/gha/builder.go
ianlewis: I think panicking is often a good idea for things that should never ever happen. In this case an already parsed URL shouldn't ever be unparseable.

It seems that I should also be able to copy the URL safely so I will go ahead and do that instead.
https://github.com/golang/go/issues/38351

```golang
withoutRef := *id.SubjectWorkflow
```

ianlewis:
> Thanks! I agree the builder code was a little hard to read

Thanks. It also raised the issue of the server url. We validate github.com urls in a couple places but we often compared just the path portions of buliderIDs which I think could be problematic security-wise.

verifiers/internal/gha/builder.go
ianlewis: Done.

PR #657, opened 2023-07-11
[docs][byob] Include information about needed BYOB builder-id input in README

### LINKED ISSUE #656: [docs][byob] Update docs to reflect that --builder-id is not optional for byob
If the `--builder-id` flag is not set when using slsa-verifier on a BYOB builder generated provenance and artifact, then the verifier will pick up the delegator as the builder-id.

Documentation should be updated to highlight this caveat for BYOB builders.

Failed SLSA verification when running on BYOB Bazel Builder generated provenance and artifact:

```
FAILED: SLSA verification failed: builderID does not match provenance: expected name 'https://github.com/enteraga6/slsa-github-generator/.github/workflows/builder_bazel_slsa3.yml', 
got 'https://github.com/slsa-framework/slsa-github-generator/.github/workflows/delegator_lowperms-generic_slsa3.yml'
exit status 1
```

After including `--builder-id https://github.com/enteraga6/slsa-github-generator/.github/workflows/builder_bazel_slsa3.yml`:

```
Verified build using builder "https://github.com/slsa-framework/slsa-github-generator/.github/workflows/delegator_lowperms-generic_slsa3.yml@refs/tags/v1.7.0" at commit 4684e76a43a096e161631f689ebbe654379e78a5
Verifying artifact ../slsa-github-generator/internal/builders/bazel/test-bin/fib2/fib2: PASSED

```

enteraga6:
closes #656

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,6 +208,12 @@ The following options are available:
 
 ## Verification for GitHub builders
 
+### BYOB BuilderID Caveat
+
+If the Github Builder used to generate the artifacts and provenance is a [BYOB Builder](https://github.com/slsa-framework/slsa-github-generator/blob/main/BYOB.md),
+then the `--builder-id` flag must be set to the BYOB Builder that was used. If not, the verification will fail as the --builder-id will automatically be set to
+BYOB's delegator.
+
 ### Artifacts
 
 To verify an artifact, run the following command:


README.md
```diff
@@ -208,6 +208,12 @@ The following options are available:
 
 ## Verification for GitHub builders
 
+### BYOB BuilderID Caveat
+
+If the Github Builder used to generate the artifacts and provenance is a [BYOB Builder](https://github.com/slsa-framework/slsa-github-generator/blob/main/BYOB.md),
```
laurentsimon, line 213: most folks that validate provenance won't know about BYOB. Maybe this should be part of the BYOB.md doc in the generator repo instead? But I see that's it's strange that some builders hosted on our repo still require it and some don't. Maybe we need to fix that in the code, ie if the builder is hosted on the generator repo, then it remains optional. Wdut?

enteraga6: > Maybe this should be part of the BYOB.md doc in the generator repo instead?

I think we still fall into the same problem then. If most that validate do not know about BYOB then including it in BYOB.md won't help most. I think adding the caveat in BYOB.md helps a bit at least.

> Maybe we need to fix that in the code, ie if the builder is hosted on the generator repo, then it remains optional.

I agree. I think this is the best course of action. When it fails on BYOB generated artifacts, slsa-verifier is able to locate what the correct builderID is:

`FAILED: SLSA verification failed: builderID does not match provenance: expected name 'https://github.com/enteraga6/slsa-github-generator/.github/workflows/builder_bazel_slsa3.yml', got 'https://github.com/slsa-framework/slsa-github-generator/.github/workflows/delegator_lowperms-generic_slsa3.yml'`

Making it remain optional would work by implicitly loading the builderID flag if it can auto-detect that the builder is  from slsa-github-generator. This would solve pain from users who just want to verify the provenance but do not care to learn about BYOB vs native, and make the process smoother.

enteraga6:
pull request closed - check out comments on #656

PR #660, opened 2023-07-18
feat: Update error message

laurentsimon:


verifiers/utils/builder.go
```diff
@@ -132,7 +132,11 @@ func IsValidBuilderTag(ref string, testing bool) error {
 	if testing {
 		// Tags on trusted repositories should be a valid semver with version
 		// core including all three parts and no build identifier.
-		versionCore := strings.Split(pin, "-")[0]
+		parts := strings.Split(pin, "-")
+		if len(parts) == 0 {
```
ianlewis, line 136: nit: I'm pretty sure `strings.Split` will never return a slice with a length of zero (if `sep` is non-empty, which it is in this case). Was this an issue you actually encountered?

verifiers/internal/gha/provenance.go
```diff
@@ -419,6 +423,10 @@ func VerifyVersionedTag(prov iface.Provenance, expectedTag string) error {
 		return err
 	}
 
+	if ref == "" {
```
ianlewis, line 426: Same as above.

verifiers/internal/gha/provenance.go
```diff
@@ -396,6 +396,10 @@ func VerifyTag(prov iface.Provenance, expectedTag string) error {
 	}
 
 	tag, err := utils.TagFromGitRef(ref)
+	if tag == "" {
```
ianlewis, line 399: `TagFromGitRef` validates that the tag isn't empty and starts with `refs/tag` so I don't think this is necessary. If the problem is with the error message itself maybe you can add this check to `utils.ValidateGitRef`?

verifiers/utils/builder.go
laurentsimon: just read the doc and you're right. Will revert.

verifiers/internal/gha/provenance.go
laurentsimon: i ran into an issue where the error message jut gave empty results, because ref is empty. That happened when a user passes `--source-versioned-tag` but the build happened *not* on a release trigger.

ianlewis: Ok, I think you can move this check to `utils.ValidateGitRef` and provide a better error message there.

verifiers/internal/gha/provenance.go
ianlewis: Ah, I see. This maybe when the ref is `refs/tags/` with no actual tag name.

verifiers/internal/gha/provenance.go
```diff
@@ -396,6 +396,10 @@ func VerifyTag(prov iface.Provenance, expectedTag string) error {
...
+		return fmt.Errorf("verifying tag: %w: no tag found in provenance", serrors.ErrorMismatchTag)
```
ianlewis, line 400:
```suggestion
		return fmt.Errorf("verifying tag: %w: empty tag name in provenance", serrors.ErrorMismatchTag)
```

PR #673, opened 2023-07-28
[feat] Non-compulsory BuilderID for BYOB Builders

enteraga6:
/cc @mihaimaruseac 

Based off the prefix of the BuilderID within the provenance, if the builder use to build the artifact is one of the BYOB builders of slsa-framework/slsa-github-generator repo, the --builderid flag is not need and is handled automatically. This was done to increase access to users since before the automatic pickup of the builder-id would get the delegator. 

Test cases that cover verifyProvenance will need to be complete after the v1.8.0 release of slsa-framework/slsa-github-generator.

closes #659

PR #674, opened 2023-07-31
feat: Non-compulsory BuilderID for BYOB Builders

### LINKED ISSUE #659: Make builder ID not compulsory for BYOB builders hosted on our repo
See original discussion in https://github.com/slsa-framework/slsa-verifier/issues/656
/cc @enteraga6

enteraga6:
/cc @mihaimaruseac
/cc @laurentsimon 

Based off the prefix of the BuilderID within the provenance, if the builder use to build the artifact is one of the BYOB builders of slsa-framework/slsa-github-generator repo, the --builderid flag is not need and is handled automatically. This was done to increase access to users since before the automatic pickup of the builder-id would get the delegator. 

Test cases that cover verifyProvenance will need to be complete after the v1.8.0 release of slsa-framework/slsa-github-generator.

The main structure that is changed is the ExpectedBuilderPath is hardcoded now to slsa-framework builders within `/cli/slsa-verifier/verify/verify_artifact.go `. This can later be changed now if needed to be an input like the other fields of `provenanceOpts` populated during `verify_artifact.go`. The added function within `provenance.go`, `verifyBuilderIDPath` is called during `verifyProvenance` to check this path within `provenanceOpts`. Upon failure of this function, expected and received BuilderID's are also outputted.

closes #659 
makes use of discussion on closed pr #673


options/options.go
```diff
@@ -27,10 +28,13 @@ type ProvenanceOpts struct {
 	ExpectedPackageName *string
 
 	ExpectedPackageVersion *string
+
+	ExpectedBuilderPath string
```
laurentsimon, line 32: nit: ExpectedBuilderIDPath

cli/slsa-verifier/verify/verify_artifact.go
```diff
@@ -54,6 +54,7 @@ func (c *VerifyArtifactCommand) Exec(ctx context.Context, artifacts []string) (*
 			ExpectedVersionedTag:   c.SourceVersionTag,
 			ExpectedTag:            c.SourceTag,
 			ExpectedWorkflowInputs: c.BuildWorkflowInputs,
+			ExpectedBuilderPath:    "https://github.com/slsa-framework/slsa-github-generator/.github/workflows/",
```
laurentsimon, line 57: this need not be provided here, otherwise it will become a default for all builders?

verifiers/internal/gha/provenance.go
```diff
@@ -58,6 +58,43 @@ func verifyBuilderIDExactMatch(prov iface.Provenance, expectedBuilderID string)
...
+	}
+
+	provBuilderID, err := utils.TrustedBuilderIDNew(id, false)
+	if err != nil {
+		return err
+	}
+
+	// This is the prefix to check for to verify that the builder is from expected builder path
+	// If this prefix is present, then the --builder-id flag is not needed at CLI command.
+	builder_path_prefix := provenanceOpts.ExpectedBuilderPath
```
laurentsimon, line 78: can we remove this temp variable?

verifiers/internal/gha/provenance.go
```diff
@@ -58,6 +58,43 @@ func verifyBuilderIDExactMatch(prov iface.Provenance, expectedBuilderID string)
...
+
+	if !(strings.HasPrefix(provBuilderID.Name(), builder_path_prefix)) {
```
laurentsimon, line 80: nit: you don't need the extra `(...)`, do you?

verifiers/internal/gha/provenance.go
```diff
@@ -58,6 +58,43 @@ func verifyBuilderIDExactMatch(prov iface.Provenance, expectedBuilderID string)
...
+		return fmt.Errorf("%w: expected slsa-framework BYOB Builder ID in provenance in order to not input --builder-id at command line. Got: %q. Expected: %q", serrors.ErrorInvalidBuilderID, provBuilderID.Name(), builder_path_prefix)
```
laurentsimon, line 81: let's simplify the error message. This part of the code is not aware of the CLI, in fact it may be called via an API. BYOB is also not known to end-users. How about "%w: build path ID mismatch: Got .... expected ..."

verifiers/internal/gha/provenance.go
```diff
@@ -58,6 +58,43 @@ func verifyBuilderIDExactMatch(prov iface.Provenance, expectedBuilderID string)
...
+	}
+
+	// Only set if user input is empty and builder is one of expected builders of Expected Builder Path
+	if strings.HasPrefix(provBuilderID.Name(), builder_path_prefix) && builderOpts.ExpectedID == nil || *builderOpts.ExpectedID == "" {
```
laurentsimon, line 85: this function should only deal with `ExpectedBuilderIDPath`, not `ExpectedID` (the caller is in charge of calling it approprietly).  The caller should also be in charge of setting the `builderOpts.ExpectedID` if needed. So this function probably needs to return `string, error`

verifiers/internal/gha/provenance.go
```diff
@@ -297,7 +336,19 @@ func VerifyProvenance(env *dsselib.Envelope, provenanceOpts *options.ProvenanceO
 		if err := isValidDelegatorBuilderID(prov); err != nil {
 			return err
 		}
+
+		// If the --builder-id flag is empty and the builder used to build artifact is not from ExpectedBuilderPath
```
laurentsimon, line 340: let's not refer to CLI argument in comments, since this part of code is not aware of the CLI. Same for comment below.

options/options.go
laurentsimon: I think we can get rid of this variable entirely, given https://github.com/slsa-framework/slsa-verifier/pull/674/files#r1280023372

verifiers/internal/gha/provenance.go
```diff
@@ -58,6 +58,43 @@ func verifyBuilderIDExactMatch(prov iface.Provenance, expectedBuilderID string)
 	return nil
 }
 
+// Inputs value in place of --BuilderID for builders of ExpectedPath such that the builderID for
+// the builders do not need to be inputted. Example: the expectedBuilderID will default to
+// the delegator builder ID for BYOB, this can verify actual BYOB builder for user without --builder-id flag.
+// Currently slsa-framework path is the only one supported for ExpectedBuilderPath.
+func verifyBuilderIDPath(prov iface.Provenance, builderOpts *options.BuilderOpts, provenanceOpts *options.ProvenanceOpts) error {
```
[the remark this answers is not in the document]
enteraga6: Good catch I just updated the logic!

verifiers/internal/gha/provenance.go
enteraga6: I agree, and I updated the scope of the comments!

verifiers/internal/gha/provenance.go
enteraga6: Setting builderOpts.ExpectedID was based off old logic. I removed it. Success message is based off `provenanceOpts.ExpectedBuilderID` and the new logic reflects this. 

Also realized that this if statement is redundant so I removed it: prefix has already been checked as well as the empty-ness of `builderOpts.ExpectedID` from the caller of this function `verifyBuilderIDPath` so the whole if statement is un-needed.

 > So this function probably needs to return `string, error`.
 
 Completely agree and this made this change.

verifiers/internal/gha/provenance.go
enteraga6: I changed the scope of error message and truncated it with your suggestion!

verifiers/internal/gha/provenance.go
enteraga6: No. I removed the outer parentheses.

verifiers/internal/gha/provenance.go
enteraga6: Yes! Un-needed now as expectedBuilderIDPath is passed in and there is no provenanceOpts (or builderOpts) passed into function anymore.

cli/slsa-verifier/verify/verify_artifact.go
enteraga6: I did this to hardcode the path for our BYOB builders to not have to insert `--builder-id` on the thought that it could later be changed to be input-based.

After our talk earlier today, I made the change we decided on to hardcode the `expectedBuilderIDPath` in `provenance.go`. 

Functionality is still the same: only BYOB builders are supported at this time for the non-compulsory BuilderID feature.

options/options.go
enteraga6: I agree. The variable has been removed entirely since `verifyBuilderIDPath` function within provenance.go only takes in the provenance and the `expectedBuilderIDPath` now.

verifiers/internal/gha/provenance.go
```diff
@@ -58,6 +58,29 @@ func verifyBuilderIDExactMatch(prov iface.Provenance, expectedBuilderID string)
...
+	if err != nil {
+		return "", err
+	}
+
+	provBuilderID, err := utils.TrustedBuilderIDNew(id, false)
+	if err != nil {
+		return "", err
+	}
+
+	// Compare actual BuilderIDPath with the expected
```
laurentsimon, line 78: nit. Add `.` at the end of your comment

verifiers/internal/gha/provenance.go
```diff
@@ -58,6 +58,29 @@ func verifyBuilderIDExactMatch(prov iface.Provenance, expectedBuilderID string)
 	return nil
 }
 
+// Inputs value in place of --BuilderID for builders of ExpectedPath such that the builderID for
```
laurentsimon: This comments should probably live in the caller. This function simply checks the builder ID starts with the provided expectedBuilderIDPath.

verifiers/internal/gha/provenance.go
```diff
@@ -297,7 +322,17 @@ func VerifyProvenance(env *dsselib.Envelope, provenanceOpts *options.ProvenanceO
 		if err := isValidDelegatorBuilderID(prov); err != nil {
 			return err
 		}
+
+		// If ExpectedID is empty, check to see if it is a trusted builder
```
laurentsimon, line 328: nit: end sentence with `.`

verifiers/internal/gha/provenance.go
```diff
@@ -297,7 +322,17 @@ func VerifyProvenance(env *dsselib.Envelope, provenanceOpts *options.ProvenanceO
...
+		if builderOpts.ExpectedID == nil || *builderOpts.ExpectedID == "" {
+			var trustedBuilderRepositoryPath = "https://github.com/slsa-framework/slsa-github-generator/.github/workflows/"
```
laurentsimon, line 330: use the variables frmo builder.go to create this string.

verifiers/internal/gha/provenance.go
enteraga6: Just did.

verifiers/internal/gha/provenance.go
enteraga6: Just did!

verifiers/internal/gha/provenance.go
enteraga6: replaced them just now!

laurentsimon: the comment `If empty and trusted builder, populate with path from provenance` seems off. If ExpectedID is empty, then we assume it's one of the builders on our repo.

verifiers/internal/gha/builder_test.go
```diff
@@ -486,10 +495,13 @@ func Test_verifyTrustedBuilderID(t *testing.T) {
 			byob:     true,
 		},
 		{
+			// This is a BYOB workflow without an id that tests non-compulsory builder-id
```
laurentsimon, line 498: can we add a unit tests to verify that verification works when no builder ID is provided but it's a trusted builder? Similarly, a test that if the builder is not a trusted one and we don't pass an explicit builder, it fails. These extra tests will need to be in provenance_test.go I think

verifiers/internal/gha/builder.go
```diff
@@ -109,11 +109,27 @@ func verifyTrustedBuilderID(certBuilderID, certTag string, expectedBuilderID *st
...
 			return nil, false, err
 		}
+
+		// Check if:
+		// - the builder in the cert is a BYOB builder
+		// - the caller trusts the BYOB builder
+		// If both are true, we don't match the user-provided builder ID
+		// against the certificate. Instead that will be done by the caller.
+		//
+		// If using a trusted delgator builder, then return the builderID and true
```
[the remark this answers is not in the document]
enteraga6: fixed.

verifiers/internal/gha/provenance.go
enteraga6: Moved and improved it.

verifiers/internal/gha/builder_test.go
enteraga6: Yes, but, per our discussion, unit tests will have to wait until the release of v1.8.0 of slsa-github-generator. I will make another PR for them after that release is cut.

laurentsimon: alright so we'll wait until then to merge this PR.

verifiers/internal/gha/provenance.go
```diff
@@ -297,7 +321,25 @@ func VerifyProvenance(env *dsselib.Envelope, provenanceOpts *options.ProvenanceO
 		if err := isValidDelegatorBuilderID(prov); err != nil {
 			return err
 		}
-		// Note: `provenanceOpts.ExpectedBuilderID` is provided by the user.
+
+		// If ExpectedID is empty, check to see if it is a trusted builder.
+		// If empty and trusted builder, populate with path from provenance,
```
laurentsimon, line 326: this comment is off `If empty and trusted builder, populate with path from provenance`. If empty, we expect a trusted builder

verifiers/internal/gha/builder_test.go
laurentsimon: actually, we don't need to cut a gen release for those unit tests. provenance.go only verifies the provenance, so you should be able to create those test.

options/options.go
```diff
@@ -18,7 +18,8 @@ type ProvenanceOpts struct {
 	// ExpectedSourceURI is the expected source URI in the provenance.
 	ExpectedSourceURI string
 
-	// ExpectedBuilderID is the expected builder ID.
+	// ExpectedBuilderID is the expected builder ID that is parsed from the envelope
+	// of the provenance.
```
ianlewis, line 22: The builder ID is parsed from the envelope?

IIUC, This ID is also the one that the user expects and passes via the command line, not one that is parsed out of the provenance.

options/options.go
```diff
@@ -31,6 +32,7 @@ type ProvenanceOpts struct {
 
 // BuildOpts are the options for checking the builder.
 type BuilderOpts struct {
-	// ExpectedID is the expected builder ID.
+	// ExpectedID is the expected builder ID that is inputted
+	// at the command line by the user to be verified.
```
ianlewis, line 36: nit: inputted just sounds weird
```suggestion
	// ExpectedID is the expected builder ID that is provided by the user.
```

verifiers/internal/gha/provenance.go
```diff
@@ -58,6 +58,28 @@ func verifyBuilderIDExactMatch(prov iface.Provenance, expectedBuilderID string)
 	return nil
 }
 
+// Verifies expectedBuilderIDPath by checking to see if the builderID in provenance
+// starts with inputted expectedBuilderIDPath.
```
ianlewis, line 62: nit: start doc comments with the name of the function as if it was public.

```suggestion
// verifyBuilderIDPath verifies that the builder ID in provenance matches the provided expectedBuilderIDPrefix.
```

verifiers/internal/gha/provenance.go
```diff
@@ -58,6 +58,28 @@ func verifyBuilderIDExactMatch(prov iface.Provenance, expectedBuilderID string)
...
+// Returns provenance builderID if verified against inputted expected path.
+func verifyBuilderIDPath(prov iface.Provenance, expectedBuilderIDPath string) (string, error) {
```
ianlewis, line 64: nit: Consider using "Prefix" or "PathPrefix" rather than "Path" here since that's how we're validating the value.
```suggestion
func verifyBuilderIDPrefix(prov iface.Provenance, expectedBuilderIDPrefix string) (string, error) {
```

verifiers/internal/gha/provenance.go
```diff
@@ -297,7 +321,25 @@ func VerifyProvenance(env *dsselib.Envelope, provenanceOpts *options.ProvenanceO
...
 		}
-		// Note: `provenanceOpts.ExpectedBuilderID` is provided by the user.
+
+		// If ExpectedID is empty, check to see if it is a trusted builder.
+		// If empty and trusted builder, populate with path from provenance,
+		// otherwise, populate from user input.
+		//
+		// Example: the expectedBuilderID will default to the delegator builder ID for BYOB.
+		// This can verify actual BYOB builderIDPath against the trusted builderIDPath inputted,
+		// Currently slsa-framework path is the only one supported for ExpectedBuilderPath.
```
ianlewis, line 331: nit: I must be getting old and I'm sure I'll die on the hill that says "inputted" is a word.
```suggestion
		// Example: the expectedBuilderID will default to the delegator builder ID for BYOB.
		// This can verify the actual BYOB builderIDPath against the trusted builderIDPath provided.
		// Currently slsa-framework path is the only one supported for ExpectedBuilderPath.
```

verifiers/internal/gha/provenance.go
```diff
@@ -297,7 +321,25 @@ func VerifyProvenance(env *dsselib.Envelope, provenanceOpts *options.ProvenanceO
...
+		if builderOpts.ExpectedID == nil || *builderOpts.ExpectedID == "" {
+			var trustedBuilderRepositoryPath = httpsGithubCom + trustedBuilderRepository + "/.github/workflows/"
+			if provenanceOpts.ExpectedBuilderID, err = verifyBuilderIDPath(prov, trustedBuilderRepositoryPath); err != nil {
+				return err
+			}
+		} else {
+			provenanceOpts.ExpectedBuilderID = *builderOpts.ExpectedID
+		}
+
+		// Note: `provenanceOpts.ExpectedBuilderID` is provided by the user
```
ianlewis, line 341: nit: NOTE is a kind of label, like "TODO"
```suggestion
		// NOTE: `provenanceOpts.ExpectedBuilderID` is provided by the user
```

verifiers/internal/gha/provenance.go
```diff
@@ -297,7 +321,25 @@ func VerifyProvenance(env *dsselib.Envelope, provenanceOpts *options.ProvenanceO
...
-		// Note: `provenanceOpts.ExpectedBuilderID` is provided by the user.
+
+		// If ExpectedID is empty, check to see if it is a trusted builder.
+		// If empty and trusted builder, populate with path from provenance,
+		// otherwise, populate from user input.
+		//
+		// Example: the expectedBuilderID will default to the delegator builder ID for BYOB.
+		// This can verify actual BYOB builderIDPath against the trusted builderIDPath inputted,
+		// Currently slsa-framework path is the only one supported for ExpectedBuilderPath.
+		if builderOpts.ExpectedID == nil || *builderOpts.ExpectedID == "" {
```
ianlewis, line 332: So we'll default to verifying the builder ID prefix if the user specifies `--builder-id=""` on the command line?

verifiers/internal/gha/builder.go
```diff
@@ -109,11 +109,25 @@ func verifyTrustedBuilderID(certBuilderID, certTag string, expectedBuilderID *st
...
+
+		// Check if:
+		// - the builder in the cert is a BYOB builder
+		// - the caller trusts the BYOB builder
+		// If both are true, we don't match the user-provided builder ID
+		// against the certificate. Instead that will be done by the caller.
+		//
+		// This return of the delegator builderID enables non-compulsory
+		// builderID feature for BYOB builders by setting byob flag to true.
+		if isTrustedDelegatorBuilder(trustedBuilderID, defaultTrustedBuilders) {
```
ianlewis, line 127: nit: This may be a style thing but we could just `return trustedBuilderID, isTrustedDelegatorBuilder(trustedBuilderID, defaultTrustedBuilders), nil` here and unindent the else block below.

verifiers/internal/gha/provenance.go
```diff
@@ -286,7 +309,8 @@ func isValidDelegatorBuilderID(prov iface.Provenance) error {
 }
 
 // VerifyProvenance verifies the provenance for the given DSSE envelope.
-func VerifyProvenance(env *dsselib.Envelope, provenanceOpts *options.ProvenanceOpts, trustedBuilderID *utils.TrustedBuilderID, byob bool) error {
+func VerifyProvenance(env *dsselib.Envelope, provenanceOpts *options.ProvenanceOpts, trustedBuilderID *utils.TrustedBuilderID, byob bool,
+	builderOpts *options.BuilderOpts) error {
```
ianlewis, line 313: nit: Since we only use `builderOpts.ExpectedID`, would it make sense to just pass the expected ID here rather than the entire `BuilderOpts`?

options/options.go
enteraga6: You understand correctly. I think confused it when I wrote that comment. I was trying to differentiate between the two since they had the same comment, and `provenanceOpts.ExpectedBuilderID` could default to the delegator workflow when the `builderOpts.ExpectedBuilderID` is empty or nil.

Here is where `provenanceOpts.ExpectedBuilderID` gets set:
https://github.com/slsa-framework/slsa-verifier/blob/main/verifiers/internal/gha/verifier.go#L58C1-L71C1

That function [`VerifyBuilderIdentity`](https://github.com/slsa-framework/slsa-verifier/blob/main/verifiers/internal/gha/builder.go#L66) gets the builderID that sets `provenanceOpts.ExpectedBuilderID` [here](https://github.com/slsa-framework/slsa-verifier/blob/main/verifiers/internal/gha/builder.go#L86) which can default to the delegator in this conditional [here](https://github.com/slsa-framework/slsa-verifier/blob/main/verifiers/internal/gha/builder.go#L108C1-L110C3)

Perhaps a better comment would be:

`// ExpectedBuilderID is the expected builder ID that is passed from user and verified`

While for `BuilderOpts.ExpectedID' we could change the comment to be:

`// ExpectedBuilderID is the builderID passed in from the user to be verified`

Thoughts?

I think it is important to clear up the two because it can be confusing if reading the code for the first time.

verifiers/internal/gha/provenance.go
enteraga6: Will do!

verifiers/internal/gha/provenance.go
enteraga6: I agree

verifiers/internal/gha/provenance.go
enteraga6: Yes good catch, thanks!

verifiers/internal/gha/provenance.go
enteraga6: Right, I guess what I was trying to say was

`// If empty, then a trusted builder is expected, to populate provenanceOpts.ExpectedBuilderID with that builder`

verifiers/internal/gha/provenance.go
enteraga6: https://www.merriam-webster.com/grammar/is-inputted-a-word

Seems to be a debated word! Lets go with provided :)

options/options.go
enteraga6: changed =)

verifiers/internal/gha/provenance.go
enteraga6: Got it, thanks.

verifiers/internal/gha/provenance.go
enteraga6: That is the current functionality. I suppose most would not go out of there way to type `--builder-id=""` if they do not have to type anything at all.

I will remove that such that this feature only gets triggered if there is no input for `--builder-id` at all.

verifiers/internal/gha/builder_test.go
enteraga6: The current BYOB Builders [Maven and Gradle (not including nodejs since it has its own separate individual logic)] in v1.7.0 are non-functional in my usage and their readme's on how to use them need to be updated.  I am referring to writing unit tests for VerifyProvenance in `verifiers/internal/gha/provenance_test.go`. Thus, I cannot generate a proper provenance to test `verifyProvenance`. Additionally, I also cannot use provenance generated by builders that are not from a tagged release. If I do not generate the provenance from a tagged release of `slsa-github-generator` then it will fail the first check within the byob conditional of `verifyProvenance`, using `isValidDelegatorBuilderID`. `isValidDelegatorBuilderID` checks for a tagged version via `utils.IsValidBuilderTag`. If it does not get a tagged release an error is thrown.

Once the v1.8.0 release gets cut, the Bazel Builder I added can generate provenance from referencing `slsa-framework/slsa-github-generator@v1.8.0` such that I can use the tagged release provenance for test cases. If the other builders are working, we could add test cases for those as well.

I have the logic for the unit tests ready when the release gets cut; I just need a tagged provenance from a BYOB builder.

options/options.go
enteraga6: Changed comment.

verifiers/internal/gha/builder_test.go
enteraga6: Test cases added.

verifiers/internal/gha/builder_test.go
```diff
@@ -486,10 +495,13 @@ func Test_verifyTrustedBuilderID(t *testing.T) {
...
+			// feature of slsa-verifier and expects byob to be true
 			name:     "low perms delegator workflow no ID provided",
 			path:     trustedBuilderRepository + "/.github/workflows/delegator_lowperms-generic_slsa3.yml",
```
ianlewis, line 501: same here
```suggestion
			path:     trustedBuilderRepository + "/.github/workflows/delegator_lowperms-generic_slsa3.yml",
			// NOTE: id is nil.
			id:       nil,
```

verifiers/internal/gha/provenance.go
```diff
@@ -315,7 +338,25 @@ func VerifyProvenance(env *dsselib.Envelope, provenanceOpts *options.ProvenanceO
 		if err := isValidDelegatorBuilderID(prov); err != nil {
 			return err
 		}
-		// Note: `provenanceOpts.ExpectedBuilderID` is provided by the user.
+
+		// If ExpectedID is not provided, check to see if it is a trusted builder.
```
ianlewis, line 342:
```suggestion
		// If expectedID is not provided, check to see if it is a trusted builder.
```

verifiers/internal/gha/provenance.go
```diff
@@ -315,7 +338,25 @@ func VerifyProvenance(env *dsselib.Envelope, provenanceOpts *options.ProvenanceO
...
+		// If not provided, then a trusted builder is expected, to populate provenanceOpts.ExpectedBuilderID
+		// with that builder, otherwise, populate from user input.
+		//
+		// Example: the expectedBuilderID will default to the delegator builder ID for BYOB.
```
ianlewis, line 346: Which `expectedBuilderID` is meant here? I don't see one in this context? Do you mean the `provenanceOpts.ExpectedBuilderID` or

verifiers/internal/gha/provenance_test.go
```diff
@@ -1182,3 +1185,115 @@ func Test_VerifyVersionedTag(t *testing.T) {
...
+				ExpectedDigest:         "caaadba2846905ac477c777e96a636e1c2e067fdf6fed90ec9eeca4df18d6ed9",
+				ExpectedSourceURI:      "github.com/enteraga6/slsa-lvl3-generic-provenance-with-bazel-example",
+				ExpectedBuilderID:      "https://github.com/slsa-framework/slsa-github-generator/.github/workflows/delegator_lowperms-generic_slsa3.yml@refs/tags/v1.8.0",
+				ExpectedWorkflowInputs: map[string]string{},
+			},
+			byob: true,
+
+			trustedBuilderIDName: "https://github.com/slsa-framework/slsa-github-generator/.github/workflows/delegator_lowperms-generic_slsa3.yml@refs/tags/v1.8.0",
+
+			expectedID: nil,
```
ianlewis, line 1225: nit: remove empty lines
```suggestion
			byob: true,
			trustedBuilderIDName: "https://github.com/slsa-framework/slsa-github-generator/.github/workflows/delegator_lowperms-generic_slsa3.yml@refs/tags/v1.8.0",
			expectedID: nil,
```

verifiers/internal/gha/provenance_test.go
```diff
@@ -1182,3 +1185,115 @@ func Test_VerifyVersionedTag(t *testing.T) {
...
+				ExpectedDigest:         "caaadba2846905ac477c777e96a636e1c2e067fdf6fed90ec9eeca4df18d6ed9",
+				ExpectedSourceURI:      "github.com/enteraga6/slsa-lvl3-generic-provenance-with-bazel-example",
+				ExpectedBuilderID:      "https://github.com/slsa-framework/slsa-github-generator/.github/workflows/delegator_lowperms-generic_slsa3.yml@refs/tags/v1.7.0",
+				ExpectedWorkflowInputs: map[string]string{},
+			},
+			byob: true,
+
+			trustedBuilderIDName: "https://github.com/slsa-framework/slsa-github-generator/.github/workflows/delegator_lowperms-generic_slsa3.yml@refs/tags/v1.7.0",
+
+			expectedID: nil,
```
ianlewis, line 1281: nit: Remove empty lines
```suggestion
			byob: true,
			trustedBuilderIDName: "https://github.com/slsa-framework/slsa-github-generator/.github/workflows/delegator_lowperms-generic_slsa3.yml@refs/tags/v1.7.0",

			expectedID: nil,
```

verifiers/internal/gha/provenance.go
enteraga6: That comment I now realize is misleading. Before that conditional, if there was no builder-id inputted, it would think the builder-id was the delegator. I was trying to show the functionality of that conditional but now I see it is just confusing! Let's remove that line.

enteraga6: Removed: "// Example: the expectedBuilderID will default to the delegator builder ID for BYOB."

verifiers/internal/gha/provenance_test.go
```diff
@@ -1182,3 +1185,111 @@ func Test_VerifyVersionedTag(t *testing.T) {
...
...91cmwiOiJodHRwczovL2FwaS5naXRodWIuY29tL3VzZXJzL2VudGVyYWdhNi9yZWNlaXZlZF9ldmVudHMiLCJyZXBvc191cmwiOiJodHRwczovL2FwaS5naXRodWIuY29tL3VzZXJzL2VudGVyYWdhNi9yZXBvcyIsInNpdGVfYWRtaW4iOmZhbHNlLCJzdGFycmVkX3VybCI6Imh0dHBzOi8vYXBpLmdpdGh1Yi5jb20vdXNlcnMvZW50ZXJhZ2E2L3N0YXJyZWR7L293bmVyfXsvcmVwb30iLCJzdWJzY3JpcHRpb25zX3VybCI6Imh0dHBzOi8vYXBpLmdpdGh1Yi5jb20vdXNlcnMvZW50ZXJhZ2E2L3N1YnNjcmlwdGlvbnMiLCJ0eXBlIjoiVXNlciIsInVybCI6Imh0dHBzOi8vYXBpLmdpdGh1Yi5jb20vdXNlcnMvZW50ZXJhZ2E2In0sIndvcmtmbG93IjoiLmdpdGh1Yi93b3JrZmxvd3MvdGVzdC12ZXJpZmllci55YW1sIn19LCJyZXNvbHZlZERlcGVuZGVuY2llcyI6W3sidXJpIjoiZ2l0K2h0dHBzOi8vZ2l0aHViLmNvbS9lbnRlcmFnYTYvc2xzYS1sdmwzLWdlbmVyaWMtcHJvdmVuYW5jZS13aXRoLWJhemVsLWV4YW1wbGVAcmVmcy9oZWFkcy9tYWluIiwiZGlnZXN0Ijp7ImdpdENvbW1pdCI6ImFiOWE0NThmNjU5ODI2MjVkZDgzODg0NDIzZmE0NTlhMjI2NGU0MjkifX1dfSwicnVuRGV0YWlscyI6eyJidWlsZGVyIjp7ImlkIjoiaHR0cHM6Ly9naXRodWIuY29tL3Nsc2EtZnJhbWV3b3JrL3Nsc2EtZ2l0aHViLWdlbmVyYXRvci8uZ2l0aHViL3dvcmtmbG93cy9idWlsZGVyX2JhemVsX3Nsc2EzLnltbEByZWZzL3RhZ3MvdjEuOC4wIn0sIm1ldGFkYXRhIjp7Imludm9jYXRpb25JZCI6Imh0dHBzOi8vZ2l0aHViLmNvbS9lbnRlcmFnYTYvc2xzYS1sdmwzLWdlbmVyaWMtcHJvdmVuYW5jZS13aXRoLWJhemVsLWV4YW1wbGUvYWN0aW9ucy9ydW5zLzU3ODgzNDIzODEvYXR0ZW1wdHMvMSJ9fX19",
```
laurentsimon, line 1204: Can you put these large contents in a file as DSSE format (under the existing testdata folder). Same comment for the  other test.

laurentsimon (commented):
LGTM. Let's just use files to store the large content of the tests

verifiers/internal/gha/provenance_test.go
enteraga6: @laurentsimon Restructured tests!

verifiers/internal/gha/testdata/bazel-trusted-dsseEnvelope.build.slsa
```diff
@@ -0,0 +1,3 @@
...
...lLCJzdGFycmVkX3VybCI6Imh0dHBzOi8vYXBpLmdpdGh1Yi5jb20vdXNlcnMvZW50ZXJhZ2E2L3N0YXJyZWR7L293bmVyfXsvcmVwb30iLCJzdWJzY3JpcHRpb25zX3VybCI6Imh0dHBzOi8vYXBpLmdpdGh1Yi5jb20vdXNlcnMvZW50ZXJhZ2E2L3N1YnNjcmlwdGlvbnMiLCJ0eXBlIjoiVXNlciIsInVybCI6Imh0dHBzOi8vYXBpLmdpdGh1Yi5jb20vdXNlcnMvZW50ZXJhZ2E2In0sIndvcmtmbG93IjoiLmdpdGh1Yi93b3JrZmxvd3MvdGVzdC12ZXJpZmllci1nbG9iLnlhbWwifX0sInJlc29sdmVkRGVwZW5kZW5jaWVzIjpbeyJ1cmkiOiJnaXQraHR0cHM6Ly9naXRodWIuY29tL2VudGVyYWdhNi9zbHNhLWx2bDMtZ2VuZXJpYy1wcm92ZW5hbmNlLXdpdGgtYmF6ZWwtZXhhbXBsZUByZWZzL2hlYWRzL21haW4iLCJkaWdlc3QiOnsiZ2l0Q29tbWl0IjoiN2YyM2FjNDc4M2NiNzAyYTY4YzU1Zjc5NTVjYzI5NmZkZDQyNWRiYiJ9fV19LCJydW5EZXRhaWxzIjp7ImJ1aWxkZXIiOnsiaWQiOiJodHRwczovL2dpdGh1Yi5jb20vc2xzYS1mcmFtZXdvcmsvc2xzYS1naXRodWItZ2VuZXJhdG9yLy5naXRodWIvd29ya2Zsb3dzL2J1aWxkZXJfYmF6ZWxfc2xzYTMueW1sQHJlZnMvdGFncy92MS44LjAifSwibWV0YWRhdGEiOnsiaW52b2NhdGlvbklkIjoiaHR0cHM6Ly9naXRodWIuY29tL2VudGVyYWdhNi9zbHNhLWx2bDMtZ2VuZXJpYy1wcm92ZW5hbmNlLXdpdGgtYmF6ZWwtZXhhbXBsZS9hY3Rpb25zL3J1bnMvNTc5MjEwODc0OC9hdHRlbXB0cy8xIn19fX0=","payloadType":"application/vnd.in-toto+json","signatures":[{"sig":"MEYCIQC5KILL/ouIQVHa10bCUW2VukAff5Towcn7t9wP48Qx2AIhAJBiNPxVT2ug4dKNzV73VR+7Qj6wNW7F+vtJ0ngsyIM4","keyid":""}]
```
laurentsimon, line 2: nit: remove this empty line

verifiers/internal/gha/testdata/bazel-untrusted-dsseEnvelope.sigstore
```diff
@@ -0,0 +1,3 @@
...
...pbiI6ZmFsc2UsInN0YXJyZWRfdXJsIjoiaHR0cHM6Ly9hcGkuZ2l0aHViLmNvbS91c2Vycy9lbnRlcmFnYTYvc3RhcnJlZHsvb3duZXJ9ey9yZXBvfSIsInN1YnNjcmlwdGlvbnNfdXJsIjoiaHR0cHM6Ly9hcGkuZ2l0aHViLmNvbS91c2Vycy9lbnRlcmFnYTYvc3Vic2NyaXB0aW9ucyIsInR5cGUiOiJVc2VyIiwidXJsIjoiaHR0cHM6Ly9hcGkuZ2l0aHViLmNvbS91c2Vycy9lbnRlcmFnYTYifSwid29ya2Zsb3ciOiIuZ2l0aHViL3dvcmtmbG93cy90ZXN0LXZlcmlmaWVyLTIueWFtbCJ9fSwicmVzb2x2ZWREZXBlbmRlbmNpZXMiOlt7InVyaSI6ImdpdCtodHRwczovL2dpdGh1Yi5jb20vZW50ZXJhZ2E2L3Nsc2EtbHZsMy1nZW5lcmljLXByb3ZlbmFuY2Utd2l0aC1iYXplbC1leGFtcGxlQHJlZnMvaGVhZHMvbWFpbiIsImRpZ2VzdCI6eyJnaXRDb21taXQiOiI3ZjIzYWM0NzgzY2I3MDJhNjhjNTVmNzk1NWNjMjk2ZmRkNDI1ZGJiIn19XX0sInJ1bkRldGFpbHMiOnsiYnVpbGRlciI6eyJpZCI6Imh0dHBzOi8vZ2l0aHViLmNvbS9lbnRlcmFnYTYvc2xzYS1naXRodWItZ2VuZXJhdG9yLy5naXRodWIvd29ya2Zsb3dzL2J1aWxkZXJfYmF6ZWxfc2xzYTMueW1sQHJlZnMvdGFncy92MS4wLjEifSwibWV0YWRhdGEiOnsiaW52b2NhdGlvbklkIjoiaHR0cHM6Ly9naXRodWIuY29tL2VudGVyYWdhNi9zbHNhLWx2bDMtZ2VuZXJpYy1wcm92ZW5hbmNlLXdpdGgtYmF6ZWwtZXhhbXBsZS9hY3Rpb25zL3J1bnMvNTc5MjIxODE2My9hdHRlbXB0cy8xIn19fX0=","payloadType":"application/vnd.in-toto+json","signatures":[{"sig":"MEUCIBI2U1FXszYZ2BU1hrtiCgfDsvLBpftB5gvGv0UrobTrAiEAnBPjQHPQHXgGRGokorag5kq9BSDNIGs90v/hHUNY77s=","keyid":""}]
```
laurentsimon, line 2: nit: remove this empty line

verifiers/internal/gha/provenance_test.go
```diff
@@ -1182,3 +1187,113 @@ func Test_VerifyVersionedTag(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			trustedBuilderID, tErr := utils.TrustedBuilderIDNew(tt.trustedBuilderIDName, true)
+			if tErr != nil {
+				t.Errorf("Provenance Verification FAILED. Error: %v", tErr)
+			}
+
+			envelopeBytes, err := os.ReadFile(filepath.Join("testdata", tt.envelopePath))
+			if err != nil {
+				panic(fmt.Errorf("os.ReadFile: %w", err))
```
laurentsimon, line 1230: let's use `t.Errorf` since we're using it in other parts. I know, our existing tests are not consistent about this!

laurentsimon (approved):
Thanks @enteraga6 !

verifiers/internal/gha/provenance_test.go
enteraga6: Switched =)

verifiers/internal/gha/provenance_test.go
```diff
@@ -1182,3 +1186,113 @@ func Test_VerifyVersionedTag(t *testing.T) {
...
+
+			vErr := VerifyProvenance(env, tt.provenanceOpts, trustedBuilderID, tt.byob, tt.expectedID)
+			if vErr != nil {
+				t.Errorf("Provenance Verification FAILED. Error: %v", vErr)
+			}
+		})
+	}
+}
+
+func Test_VerifyUntrustedProvenance(t *testing.T) {
```
laurentsimon, line 1245: sorry I missed something: Can you merge these two tests under `Test_VerifyProvenance()` function? 
Use the `expected` field to check whether the verification failed or passed. You can use `cmp.Equal` to compare errors, like in other parts of the code
Also, add some test with an empty ExpectedBuilderID to verify that the changes work
Also add corresponding test for when the builder tag is empty

verifiers/internal/gha/provenance_test.go
```diff
@@ -1182,3 +1186,113 @@ func Test_VerifyVersionedTag(t *testing.T) {
...
+				t.Errorf("os.ReadFile: %v", err)
+			}
+
+			env, err := EnvelopeFromBytes(envelopeBytes)
+			if err != nil {
+				t.Errorf("unexpected error parsing envelope %v", err)
+			}
+
+			vErr := VerifyProvenance(env, tt.provenanceOpts, trustedBuilderID, tt.byob, tt.expectedID)
+			if vErr != nil {
```
laurentsimon, line 1238: nit: re-use `err`

enteraga6: Done

verifiers/internal/gha/provenance_test.go
enteraga6: Added tests - no test for builder tag after our earlier correspondence

Issue #652, opened 2023-07-07
Share v0.2 / v1.0 interfaces among builders


laurentsimon:
We currently have interfaces for v0.2 and v1.0 SLSA specs for GitHub builders, but they are not shared with other builders like GCB

PR #682, opened 2023-08-10
feat: GCB refactor for v1.0 support

laurentsimon:
In anticipation for GCB's v1.0 support, this PR re-factors the code to look similar to GHA's code


laurentsimon:
@ianlewis friendly ping for a review

verifiers/internal/gcb/provenance.go
```diff
@@ -68,7 +53,7 @@ type gloudProvenance struct {
 type Provenance struct {
 	gcloudProv              *gloudProvenance
 	verifiedProvenance      *provenance
-	verifiedIntotoStatement *v01IntotoStatement
+	verifiedIntotoStatement *iface.Statement
```
ianlewis, line 56: This should definitely not be a pointer to an interface (interface references are always pointers).
```suggestion
	verifiedIntotoStatement iface.Statement
```

verifiers/internal/gcb/provenance.go
```diff
@@ -166,11 +151,23 @@ func (p *Provenance) VerifyTextProvenance() error {
 		return err
 	}
 
-	// Note: there is an additional field `metadata.buildInvocationId` which
-	// is not part of the specs but is present. This field is currently ignored during comparison.
-	unverifiedTextIntotoStatement := v01IntotoStatement{
-		StatementHeader: p.verifiedProvenance.Build.UnverifiedTextIntotoStatement.StatementHeader,
-		Predicate:       p.verifiedProvenance.Build.UnverifiedTextIntotoStatement.SlsaProvenance,
+	statement := p.verifiedIntotoStatement
+	predicate, err := (*statement).Predicate()
```
ianlewis, line 155: If you change the pointer definition above you won't need to do this.
```suggestion
	predicate, err := p.verifiedIntotoStatement.Predicate()
```

verifiers/internal/gcb/provenance.go
```diff
@@ -166,11 +151,23 @@ func (p *Provenance) VerifyTextProvenance() error {
...
+	if err != nil {
+		return err
+	}
+
+	var unverifiedTextIntotoStatement interface{}
+	switch v := predicate.(type) {
+	case v01.ProvenancePredicate:
+		// Note: there is an additional field `metadata.buildInvocationId` which
```
ianlewis, line 163: nit: uppercase NOTE labels
```suggestion
		// NOTE: there is an additional field `metadata.buildInvocationId` which
```

verifiers/internal/gcb/provenance.go
```diff
@@ -166,11 +151,23 @@ func (p *Provenance) VerifyTextProvenance() error {
...
-		StatementHeader: p.verifiedProvenance.Build.UnverifiedTextIntotoStatement.StatementHeader,
-		Predicate:       p.verifiedProvenance.Build.UnverifiedTextIntotoStatement.SlsaProvenance,
+	statement := p.verifiedIntotoStatement
+	predicate, err := (*statement).Predicate()
+	if err != nil {
+		return err
+	}
+
+	var unverifiedTextIntotoStatement interface{}
+	switch v := predicate.(type) {
```
ianlewis, line 161: I wonder if we really need a `Statement` interface and type checks. Couldn't we just check the `predicateType` for the SLSA version?

verifiers/internal/gcb/slsaprovenance/iface/statement.go
```diff
@@ -0,0 +1,29 @@
+package iface
+
+import (
+	intoto "github.com/in-toto/in-toto-golang/in_toto"
+)
+
+// Statement represents statement for a predicate type and build type.
+type Statement interface {
```
ianlewis, line 8: This looks pretty similar to the gha `Provenance` interface. Is it not possible to re-use that interface somehow?

ianlewis (approved):

verifiers/internal/gcb/provenance.go
laurentsimon: I am not sure this works, because we're testing that the point is nil in the code (this pointer is set only when we found a verified statement https://github.com/slsa-framework/slsa-verifier/blob/main/verifiers/internal/gcb/provenance.go#L541). 

Or maybe you're suggesting we replace the nil checks by an "empty interface" check?

verifiers/internal/gcb/provenance.go
laurentsimon: I think you're right. Ill try that

verifiers/internal/gcb/provenance.go
ianlewis: Since `iface.Statement` is an interface any reference is automatically a pointer so you can assign `nil` and check for it. You should almost never need a pointer to an interface. `*iface.Statement` is a pointer to a pointer.

laurentsimon: if `interface == nil` was not compiling when I tried, I think. I'll double check

verifiers/internal/gcb/slsaprovenance/iface/statement.go
laurentsimon: Hoping it is, https://github.com/slsa-framework/slsa-verifier/issues/652 is tracking it. I've been hesitant to change the code too much in one PR, but this is on the TODO list. So I plan on iterating. Do you think I should attempt it in this PR?

ianlewis: As long as we have some idea and we are tracking it I'm ok with that.

PR #689, opened 2023-08-15
feat: Allow byob builders ref at main for e2e tests

laurentsimon:


verifiers/internal/gha/provenance.go
```diff
@@ -294,17 +294,42 @@ func isValidDelegatorBuilderID(prov iface.Provenance) error {
 	if err != nil {
 		return err
 	}
+
+	sourceURI, err := prov.SourceURI()
+	if err != nil {
+		return err
+	}
+
+	uri, _, err := utils.ParseGitURIAndRef(sourceURI)
```
kpk47, line 303: You aren't using the value `uri` before overwriting it at L323. Is that intentional?

laurentsimon: yes. Let me rename the second uri to normalizedUri

laurentsimon: done:
normalizedE2eRepoUri and normalizedUri

kpk47 (approved):

Issue #711, opened 2023-10-03
feat: Rename verifySubjectDigest to verifyPuboishAttestationSubjectDigest


laurentsimon:
See https://github.com/slsa-framework/slsa-verifier/blob/main/verifiers/internal/gha/npm.go#L371

PR #712, opened 2023-10-09
feat: Rename verifySubjectDigest function

laurentsimon:
closes https://github.com/slsa-framework/slsa-verifier/issues/711


trishankatdatadog (commented):
Sorry, I don't see it right now: how exactly does this close #711?

laurentsimon:
> Sorry, I don't see it right now: how exactly does this close #711?

You should see it now :)

cli/slsa-verifier/main_regression_test.go
```diff
@@ -1674,15 +1674,15 @@ func Test_runVerifyNpmPackage(t *testing.T) {
 			name:      "invalid signature provenance npm CLI",
 			artifact:  "provenance-npm-test-cli-v02-prega-invalidsigprov.tgz",
 			source:    "github.com/laurentsimon/provenance-npm-test",
-			pkgName:   PointerTo("@laurentsimon/provenance-npm-test2"),
+			pkgName:   PointerTo("@laurentsimon/provenance-npm-test"),
```
trishankatdatadog, line 1677: Thanks for the new commits: it's better, but why change the pkgName here? Was it a bug?

laurentsimon: This unit test is supposed to check signature verification failure, not package name. So it should fail with the right package name given as input (the package name need not be different from what's in the attestation). I just realized it should be improved, it's _not_ a bug.

trishankatdatadog (approved):

ianlewis (approved):

PR #731, opened 2024-01-10
feat: fixes #547: add npm sigstore-tuf suport

### LINKED ISSUE #547: [npm] Verify publish attestations from npm package signing keys
The npm package signing key will be distributed through Sigstore's TUF root as of next week.

The target will be found at `registry.npmjs.org/keys.json` and has a format like this:https://raw.githubusercontent.com/sigstore/root-signing/f3ebf7f951e021d2b1ad94735d33af7f7834793f/repository/staged/targets/registry.npmjs.org/keys.json

where there is a signatures key and an attestations key.

To get this file in bytes, you can use the sigstore/sigstore TUF client like this:
```
	tufClient, err := tuf.NewFromEnv(context.Background())
	if err != nil {
		check(err)
	}
       keysFileBytes, err := tufClient.GetTarget(registry.npmjs.org/keys.json)
	if err != nil {
		check(err)
	}
```

The format of this file is defined loosely from the protobufs, at least the PublicKeys are from this [proto](https://github.com/sigstore/protobuf-specs/blob/c8a23a43dc1c685d336d112864ba237a15061c53/protos/sigstore_common.proto#L100)

ramonpetgrave64:
Addresses: https://github.com/slsa-framework/slsa-verifier/issues/547
 - [x] Pending: https://github.com/sigstore/sigstore-go/pull/41
Uses the new [sigstore-go@0.2.0](https://github.com/sigstore/sigstore-go/releases/tag/v0.2.0)

Currently slsa-verifier has npmjs' attestation key hardcoded. But sigstore now stores the same key within their own TUF root. 

This PR 
 - dynamically use the keyid specified in the sigstore bundle, rather than the hardcoded keyid.
 - uses an updated ([pending]( https://github.com/sigstore/sigstore-go/pull/41)) sigstore-go library that allows us to fetch a signed and verified copy of the same key.


ramonpetgrave64:
@laurentsimon please take a look

verifiers/internal/gha/npm.go
```diff
@@ -144,6 +128,12 @@ func (n *Npm) verifyPublishAttestationSignature() error {
 		return err
 	}
 
+	// retrieve the key material
```
laurentsimon, line 131: nit: start comment with capital letter and end with dot.

verifiers/internal/gha/npm_sigstore_tuf.go
```diff
@@ -0,0 +1,114 @@
...
+
+type SigstoreTufClient interface {
+	GetTarget(target string) ([]byte, error)
+}
+
+// NewSigstoreTufClient Get a Sigstore TUF client, which itself is a wrapper around the official TUF client.
+func NewSigstoreTufClient() (*sigstoreTuf.Client, error) {
+	_, filename, _, ok := runtime.Caller(1)
+	if !ok {
+		log.Fatal("unable to get path")
```
laurentsimon, line 46: return an error instead of logging

verifiers/internal/gha/npm_sigstore_tuf.go
```diff
@@ -0,0 +1,114 @@
...
+	"fmt"
+	"log"
+	"path"
+	"runtime"
+	"time"
+
+	sigstoreTuf "github.com/sigstore/sigstore-go/pkg/tuf"
+)
+
+const (
```
laurentsimon, line 14: those need not be public, I think. You should be able to use lower case for first letter?
I's under internal directory so not a super big deal :)

verifiers/internal/gha/npm_sigstore_tuf.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	return client, nil
+}
+
+/*
+GetNpmjsKeysTarget Fetch and parse the keys.json file in Sigstore's root for npmjs
+The inner TUF client will verify this "blob" is signed with correct delegate TUF roles
+https://github.com/sigstore/root-signing/blob/5fd11f7ec0a993b0f20c335b33e53cfffb986b2e/repository/repository/targets/registry.npmjs.org/7a8ec9678ad824cdccaa7a6dc0961caf8f8df61bc7274189122c123446248426.keys.json#L4
+*/
+func GetNpmjsKeysTarget(client SigstoreTufClient, targetPath string) (*NpmjsKeysTarget, error) {
```
laurentsimon, line 62: nit: make all these function private

go.mod
```diff
@@ -4,50 +4,59 @@ go 1.20
...
-	github.com/spf13/cobra v1.7.0
-	golang.org/x/mod v0.13.0
-	sigs.k8s.io/release-utils v0.7.4
+	github.com/spf13/cobra v1.8.0
+	golang.org/x/mod v0.14.0
+	sigs.k8s.io/release-utils v0.7.7
 )
 
+// use the pending PR #41 branch tuf-client-2
+replace github.com/sigstore/sigstore-go => github.com/sigstore/sigstore-go v0.0.0-20231222133331-d489b534902f
```
laurentsimon, line 30: note to you: don't forget to replace before merging

laurentsimon (approved):
minor nits, but LGTM. Thank you!

.github/workflows/pre-submit.cli.yml
```diff
@@ -36,4 +36,5 @@ jobs:
       - name: Run tests for verifier
         run: |
           set -euo pipefail
+          make unit-test
```
laurentsimon, line 39: Do we need this change? Unit tests are run in another workflow, no?

ramonpetgrave64: I'm not 100% but it seems like the regression-tests command does cover the unit tests, as well.

* [go test -mod=vendor -tags=regression ...](https://github.com/ramonpetgrave64/slsa-verifier/blob/edb1decd90fa511993ead9c3f7917e5abe371c10/Makefile#L47)
* [go:build](https://github.com/ramonpetgrave64/slsa-verifier/blob/edb1decd90fa511993ead9c3f7917e5abe371c10/cli/slsa-verifier/main_regression_test.go#L1)

verifiers/internal/gha/npm_sigstore_tuf_test.go
```diff
@@ -0,0 +1,120 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			mockClient := mockSigstoreTufClient{localPath: tt.localPath}
+			keys, err := getNpmjsKeysTarget(mockClient, tt.localPath)
+			if err != nil {
+				t.Error(err)
+			}
+			actualKeyData, err := getKeyDataWithNpmjsKeysTarget(keys, tt.keyID, tt.keyUsage)
+			if !tt.expectError {
+				if err != nil {
+					t.Error(err)
```
ianlewis, line 105:
```suggestion
					t.Fatalf("getKeyDataWithNpmjsKeysTarget: %v", err)
```

verifiers/internal/gha/npm_sigstore_tuf_test.go
```diff
@@ -0,0 +1,120 @@
...
+				}
+				if tt.expectedKeyData != actualKeyData {
+					t.Errorf("expected equal values: \nexpected: %v \nactual: %v", tt.expectedKeyData, actualKeyData)
+				}
+			} else {
+				if err == nil {
+					t.Error("expected and error")
```
ianlewis, line 112:
```suggestion
					t.Error("expected an error")
```

verifiers/internal/gha/npm_sigstore_tuf_test.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			mockClient := mockSigstoreTufClient{localPath: tt.localPath}
+			keys, err := getNpmjsKeysTarget(mockClient, tt.localPath)
+			if err != nil {
+				t.Error(err)
+			}
+			actualKeyData, err := getKeyDataWithNpmjsKeysTarget(keys, tt.keyID, tt.keyUsage)
+			if !tt.expectError {
```
ianlewis, line 103: nit: I'm not sure what kind of errors the tuf client returns but if they have consistent error types you could use `!cmp.Equal(t.expectedErr, err, cmpopts.EquateErrors)` to test the error return value matches the expected value just like any other return value. I think we do this in tests elsewhere in the codebase.

`diff := cmp.Diff(t.expectedErr, err, cmpopts.EquateErrors); diff != ""` is maybe better even.

verifiers/internal/gha/npm_sigstore_tuf_test.go
```diff
@@ -0,0 +1,120 @@
...
+		actualKeys, err := getNpmjsKeysTarget(mockClient, testTargetLocalFilePath)
+		if err != nil {
+			t.Error(err)
+		}
+		if err != nil {
+			t.Error(err)
+		}
+		if !cmp.Equal(expectedKeys, *actualKeys) {
+			t.Errorf("expected equal values: \nexpected: %v \nactual: %v", expectedKeys, *actualKeys)
+		}
```
ianlewis, line 55: nit: it's not super important but would be nice to stick to the convention of using `want` and `got` variables for comparisons. 
```golang
		if got, want := *actualKeys, expectedKeys; !cmp.Equal(want, got) {
			t.Errorf("expected equal values: \nwant: %v \ngot: %v", want, got)
		}
```

`cmp.Diff` is maybe even better.

```suggestion
		if diff := !cmp.Diff(expectedKeys, *actualKeys); diff != "" {
			t.Errorf("expected equal values (-want +got):\n%s", diff)
		}
```

verifiers/internal/gha/npm_sigstore_tuf_test.go
```diff
@@ -0,0 +1,120 @@
...
+			t.Errorf("parsing mock file: %s", err)
+		}
+		mockClient := mockSigstoreTufClient{localPath: testTargetLocalFilePath}
+		actualKeys, err := getNpmjsKeysTarget(mockClient, testTargetLocalFilePath)
+		if err != nil {
+			t.Error(err)
+		}
+		if err != nil {
+			t.Error(err)
+		}
```
ianlewis, line 52: nit: You only need one check and it should return early if there is an error. I presume `actualKeys` could be `nil` in that case and would cause a panic later.
```suggestion
		if err != nil {
			t.Fatalf("getNpmjsKeysTarget: %v", err)
		}
```

verifiers/internal/gha/npm_sigstore_tuf_test.go
```diff
@@ -0,0 +1,120 @@
...
+			expectedKeyData: "", // should not be returned in this error case
+			expectError:     true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			mockClient := mockSigstoreTufClient{localPath: tt.localPath}
+			keys, err := getNpmjsKeysTarget(mockClient, tt.localPath)
+			if err != nil {
+				t.Error(err)
```
ianlewis, line 100: nit: `keys` might be undefined if an error is returned so we should fail early.
```suggestion
				t.Fatalf("getNpmjsKeysTarget: %v", err)
```

verifiers/internal/gha/npm_sigstore_tuf_test.go
```diff
@@ -0,0 +1,120 @@
...
+func TestGetNpmjsKeysTarget(t *testing.T) {
+	t.Run("parsing local registry.npmjs.org_keys.json", func(t *testing.T) {
+		content, err := os.ReadFile(testTargetLocalFilePath)
+		if err != nil {
+			t.Errorf("reading local file: %s", err)
+		}
+		var expectedKeys npmjsKeysTarget
+		err = json.Unmarshal(content, &expectedKeys)
+		if err != nil {
+			t.Errorf("parsing mock file: %s", err)
```
ianlewis, line 43:
```suggestion
			t.Errorf("parsing mock file: %v", err)
```

verifiers/internal/gha/npm_sigstore_tuf_test.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+	return content, nil
+}
+
+// TestGetNpmjsKeysTarget ensures we can parse the target file.
+func TestGetNpmjsKeysTarget(t *testing.T) {
+	t.Run("parsing local registry.npmjs.org_keys.json", func(t *testing.T) {
+		content, err := os.ReadFile(testTargetLocalFilePath)
+		if err != nil {
+			t.Errorf("reading local file: %s", err)
```
ianlewis, line 38: nit: Mostly the same, but I think usually you want `%v` when printing errors.
```suggestion
			t.Errorf("reading local file: %v", err)
```

verifiers/internal/gha/npm_sigstore_tuf.go
```diff
@@ -0,0 +1,111 @@
...
+	Start time.Time `json:"start"`
+}
+
+type sigstoreTufClient interface {
+	GetTarget(target string) ([]byte, error)
+}
+
+// newSigstoreTufClient Get a Sigstore TUF client, which itself is a wrapper around the official TUF client.
+func newSigstoreTufClient() (*sigstoreTuf.Client, error) {
+	_, filename, _, ok := runtime.Caller(1)
```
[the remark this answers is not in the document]
ianlewis: Is there a standard place to put the `tufcache` that other clients use?

verifiers/internal/gha/npm_sigstore_tuf_test.go
laurentsimon: +1 on using Diff

verifiers/internal/gha/npm_sigstore_tuf_test.go
laurentsimon: +1 on using Diff here

verifiers/internal/gha/npm_sigstore_tuf_test.go
```diff
@@ -0,0 +1,120 @@
...
+			expectedKeyData: testTargetKeyData,
+			expectError:     false,
+		},
+		{
+			name:            "missing the 'npm:attestations' keyusage",
+			localPath:       "./testdata/wrong_keyusage_registry.npmjs.org_keys.json",
+			keyID:           testTargetKeyID,
+			keyUsage:        testTargetKeyUsage,
+			expectedKeyData: "", // should not be returned in this error case
+			expectError:     true,
```
laurentsimon, line 92: define errors and match in the unit test instead of using a bool

verifiers/internal/gha/npm_sigstore_tuf_test.go
```diff
@@ -0,0 +1,120 @@
...
+		if !cmp.Equal(expectedKeys, *actualKeys) {
+			t.Errorf("expected equal values: \nexpected: %v \nactual: %v", expectedKeys, *actualKeys)
+		}
+	})
+
+	t.Run("parsing non-existent registry.npmjs.org_keys.json", func(t *testing.T) {
+		nonExistantPath := "./testdata/my-fake-path"
+		mockClient := mockSigstoreTufClient{localPath: nonExistantPath}
+		_, err := getNpmjsKeysTarget(mockClient, nonExistantPath)
+		if err == nil {
```
laurentsimon, line 62: re-use Diff here vas well

verifiers/internal/gha/npm_sigstore_tuf_test.go
ianlewis: This is essentially the same as my other comment. Though I think it might depend on the types of errors we get from the tuf client. Maybe we could wrap the errors in stable types ourselves if we don't get stable types from the tuf client?

verifiers/internal/gha/npm_sigstore_tuf.go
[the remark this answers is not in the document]
ianlewis: A singleton `atomic.Value` is probably good since it's a global value and wouldn't depend on any client instance.

verifiers/internal/gha/npm.go
```diff
@@ -144,6 +127,12 @@ func (n *Npm) verifyPublishAttestationSignature() error {
 		return err
 	}
 
+	// Retrieve the key material.
+	npmRegistryPublicKey, err := getKeyDataFromSigstoreTuf(npmRegistryPublicKeyID, attestationKeyUsage)
```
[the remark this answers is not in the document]
ramonpetgrave64: iiuc, the inner tuf client will use a local cache, so it's really only fetched from the internet once per invocation of `getKeyDataFromSigstoreTuf`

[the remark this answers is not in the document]
laurentsimon: by "local", you mean in memory or on disk?

verifiers/internal/gha/npm_sigstore_tuf.go
```diff
@@ -0,0 +1,111 @@
+package gha
```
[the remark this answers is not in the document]
laurentsimon: good idea. We have a larger goal of restructuring long term: sigstore -> github / gitlab -> npm / container.. something like that.
@ramonpetgrave64 is going to help with that as well.

verifiers/internal/gha/npm.go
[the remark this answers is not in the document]
ramonpetgrave64: On disk. But now I think about it, perhaps @ianlewis meant the repeated verifications that GetTarget will do are expensive. Even if it did not do a repeated verification, it may still be better to not rely on the local cache on disk. I'll change it to preserve the final keydata in slsa-verifier memory.

verifiers/internal/gha/npm_sigstore_tuf.go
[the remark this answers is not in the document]
ramonpetgrave64: added. see https://github.com/slsa-framework/slsa-verifier/pull/731#discussion_r1459944529

verifiers/internal/gha/npm.go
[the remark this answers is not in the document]
ianlewis: I was really trying to make sure we don't go to the internet each time, but yeah, if we can avoid hitting the disk too so much the better.

verifiers/internal/gha/npm_sigstore_tuf_test.go
ramonpetgrave64: I found some references that suggest that %w is slightly better than %v because you can also unwrap, so I'll use that.
* https://go.dev/blog/go1.13-errors

verifiers/internal/gha/npm_sigstore_tuf_test.go
ramonpetgrave64: We can't use Diff to compare nil values. But I added a check to ensure that the return value is nil.

laurentsimon: I think you just need to cast the nil to a type error. I've used it in the past. DO you have an example where it does not work? Or some doc?

verifiers/internal/gha/npm_sigstore_tuf_test.go
ramonpetgrave64: Added our own errors to wrap theirs

verifiers/internal/gha/testdata/registry.npmjs.org_keys.json
```diff
@@ -0,0 +1,26 @@
+{
```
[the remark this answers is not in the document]
ramonpetgrave64: added

verifiers/internal/gha/npm_sigstore_tuf_test.go
ianlewis: Pretty sure, you won't be able to use that with `t.Errorf` since that's just recording an error. Only `fmt.Errorf` is the only place where `%w` is supported IIRC.

verifiers/internal/gha/npm_sigstore_tuf_test.go
ianlewis: Sure you can't really do `cmp.Diff(someval, nil)` but you could do:
```go
var nilErr error
if diff := cmp.Diff(nilErr, someErr); diff != "" {
    ...
}```

Comment (ianlewis) [reply verifiers/internal/gha/npm_sigstore_tuf_test.go]: In reality though you probably don't need to check for a nil error. The ideal thing would be to return an specific error type and check for the error type you expect.

```go
if diff := cmp.Diff(wantErrType{}, err, cmpopts.EquateErrors); diff != "" {
    t.Fatalf("unexpected error (-want +got): \n%s", diff)
}
```

ramonpetgrave64: @ianlewis not about nil errors, but about using `Diff` against an nil value and a non-nil struct. 
But never mind it works now! I'm not sure what I was doing wrong, but I Think I remember something about `Diff`'s error saying the types of the arguments weren't comparable.

laurentsimon: you should be able to cast, something like `error(nil)`

ianlewis: > @ianlewis not about nil errors, but about using `Diff` against an nil value and a non-nil struct. But never mind it works now! I'm not sure what I was doing wrong, but I Think I remember something about `Diff`'s error saying the types of the arguments weren't comparable.

Right. That's why I suggested declaring a nil value of `error` type first. This is essentially the same as @laurentsimon's suggestion of using a casted value (`error(nil)`) but maybe a bit more readable IMO. Either way is fine though I think.

I do realize that I should have made the check using an equals though. my bad.

The code looks much much better now that you are checking against a specific error type. GJ!

verifiers/internal/gha/npm_sigstore_tuf.go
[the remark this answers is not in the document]
ianlewis: SG.

verifiers/internal/gha/npm_sigstore_tuf_test.go
```diff
@@ -0,0 +1,187 @@
...
+	testTargetKeyData  = testTargetKey.PublicKey.RawBytes
+)
+
+// mockSigstoreTufClient a mock implementation of sigstoreTufClient.
+type mockSigstoreTufClient struct {
+	fileContentMap map[string]string
+}
+
+// NewMockSigstoreTufClient returns an instance of the mock client,
+// with fileContentMap as input and outputs of the GetTarget() method
```
ianlewis, line 92:
```suggestion
// with fileContentMap as input and outputs of the GetTarget() method.
```

verifiers/internal/gha/npm_sigstore_tuf.go
```diff
@@ -0,0 +1,110 @@
...
+	if err := json.Unmarshal(blob, &keys); err != nil {
+		return nil, fmt.Errorf("%w: %w", ErrorCouldNotParseKeys, err)
+	}
+	return &keys, nil
+}
+
+/*
+getKeyDataWithNpmjsKeysTarget Given our set of keys, return the target key's material.
+We may also want to check the existing ValidFor.Start (and a potential future ValidFor.End).
+*/
```
ianlewis, line 76:
```suggestion
// getKeyDataWithNpmjsKeysTarget Given our set of keys, return the target key's material.
// We may also want to check the existing ValidFor.Start (and a potential future ValidFor.End).
```

verifiers/internal/gha/npm_sigstore_tuf.go
```diff
@@ -0,0 +1,110 @@
...
+/*
+getKeyDataFromSigstoreTuf given a keyid and keyusage, retriive the keyfile from sigstore's TUF root,
+parse the file and return the specific key material.
+See documentation for getNpmjsKeysTarget
+
+example params:
+
+	keyID: "SHA256:jl3bwswu80PjjokCgh0o2w5c2U4LhQAE57gj9cz1kzA"
+	keyUsage: "npm:attestations"
+*/
```
ianlewis, line 95:
```suggestion
// getKeyDataFromSigstoreTuf given a keyid and keyusage, retriive the keyfile from sigstore's TUF root,
// parse the file and return the specific key material.
// See documentation for getNpmjsKeysTarget
// 
// example params:
//
//	keyID: "SHA256:jl3bwswu80PjjokCgh0o2w5c2U4LhQAE57gj9cz1kzA"
//	keyUsage: "npm:attestations"
```

ianlewis (approved):

go.mod
ramonpetgrave64: It's now using sigstore/sigstore-go@0.2.0

renovate.json
```diff
@@ -6,6 +6,8 @@
     "schedule": "before 4am"
   },
   "postUpdateOptions": ["gomodTidy", "gomodUpdateImportPaths"],
+  "postUpgradeTasks": { "commands": ["(cd ./actions/installer && make clean && make package)"] },
+  "allowedPostUpgradeCommands": ["(cd ./actions/installer && make clean && make package)"],
```
ianlewis, line 10: https://docs.renovatebot.com/configuration-options/#postupgradetasks
> "Post-upgrade tasks can only be used on self-hosted Renovate instances."

Unfortunately I don't think we can have rennovate update the `dist/` directory automatically without running a self-hosted instance.

verifiers/internal/gha/npm_sigstore_tuf.go
```diff
@@ -0,0 +1,104 @@
...
+func newSigstoreTufClient() (*sigstoreTuf.Client, error) {
+	opts := sigstoreTuf.DefaultOptions()
+	client, err := sigstoreTuf.New(opts)
+	if err != nil {
+		return nil, fmt.Errorf("creating SigstoreTuf client: %w", err)
+	}
+	return client, nil
+}
+
+// getNpmjsKeysTarget Fetch and parse the keys.json file in Sigstore's root for npmjs
```
ianlewis, line 56: nit
```suggestion
// getNpmjsKeysTarget will fetch and parse the keys.json file in Sigstore's root for npmjs
```

verifiers/internal/gha/npm_sigstore_tuf.go
```diff
@@ -0,0 +1,104 @@
...
+const (
+	attestationKeyUsage = "npm:attestations"
+	attestationKeyID    = "SHA256:jl3bwswu80PjjokCgh0o2w5c2U4LhQAE57gj9cz1kzA"
+	targetPath          = "registry.npmjs.org/keys.json"
+)
+
+var (
+	ErrorMissingNpmjsKeyIDKeyUsage = errors.New("could not find a key with the specified 'keyId' and 'keyUsage'")
+	ErrorCouldNotFindTarget        = errors.New("could not get the target from the tuf root")
+	ErrorCouldNotParseKeys         = errors.New("could not parse keys file content")
```
ianlewis, line 21: Do we use this outside the package or just for tests? If the latter, we could make this unexported.
```suggestion
	errMissingNpmjsKeyIDKeyUsage = errors.New("could not find a key with the specified 'keyId' and 'keyUsage'")
	errCouldNotFindTarget        = errors.New("could not get the target from the tuf root")
	errCouldNotParseKeys         = errors.New("could not parse keys file content")
```

verifiers/internal/gha/npm_sigstore_tuf_test.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			mockClient := NewMockSigstoreTufClient()
+			actualKeys, err := getNpmjsKeysTarget(mockClient, tt.targetPath)
+			if keyDataDiff := cmp.Diff(tt.expectedKeys, actualKeys, cmpopts.EquateComparable()); keyDataDiff != "" {
+				t.Errorf("expected equal values (-want +got):\n%s", keyDataDiff)
+			}
+			if errorDiff := cmp.Diff(tt.expectedErr, err, cmpopts.EquateErrors()); errorDiff != "" {
+				log.Printf("want: %v, got: %v, type:%T", tt.expectedErr, err, err)
```
ianlewis, line 138: nit: Do you need this message if you call `t.Error` right after it?
```suggestion
```

verifiers/internal/gha/npm_sigstore_tuf_test.go
```diff
@@ -0,0 +1,187 @@
...
+				t.Errorf("expected equaivalent errors (-want +got):\n%s", errorDiff)
```
ianlewis, line 139: nit
```suggestion
				t.Errorf("unexpected error (-want +got):\n%s", errorDiff)
```

verifiers/internal/gha/npm_sigstore_tuf_test.go
```diff
@@ -0,0 +1,187 @@
...
+			keys, err := getNpmjsKeysTarget(mockClient, tt.targetPath)
+			if err != nil {
+				t.Fatalf("getNpmjsKeysTarget: %v", err)
+			}
+			actualKeyData, err := getKeyDataWithNpmjsKeysTarget(keys, tt.keyID, tt.keyUsage)
+			if keyDataDiff := cmp.Diff(tt.expectedKeyData, actualKeyData); keyDataDiff != "" {
+				t.Errorf("expected equal values (-want +got):\n%s", keyDataDiff)
+			}
+			if errorDiff := cmp.Diff(tt.expectedErr, err, cmpopts.EquateErrors()); errorDiff != "" {
+				t.Errorf("expected equaivalent errors (-want +got):\n%s", errorDiff)
```
ianlewis, line 183: nit
```suggestion
				t.Errorf("unexpected error (-want +got):\n%s", errorDiff)
```

ianlewis (approved):

renovate.json
ramonpetgrave64: Oh I included this by accident while working on something else.
 - https://github.com/slsa-framework/slsa-verifier/pull/754

And yes I found the same to be true.

verifiers/internal/gha/npm_sigstore_tuf_test.go
ramonpetgrave64: I was debugging something again, after the long pause. I'll convert this PR back to a draft for now, while we decide on how exactly to get the KeyID dynamically.

verifiers/internal/gha/npm_sigstore_tuf_test.go
ramonpetgrave64: changed

verifiers/internal/gha/npm_sigstore_tuf_test.go
ramonpetgrave64: changed

verifiers/internal/gha/npm_sigstore_tuf.go
ramonpetgrave64: changed

ramonpetgrave64:
I've updated the PR to dynamically use the keyid specified in the sigstore bundle, rather than the hardcoded keyid.

verifiers/internal/gha/npm_sigstore_tuf.go
ramonpetgrave64: changed

verifiers/internal/gha/npm.go
```diff
@@ -144,6 +141,15 @@ func (n *Npm) verifyPublishAttestationSignature() error {
 		return err
 	}
 
+	// Get the keyID used to sign the attestaion
```
laurentsimon, line 144: nit: Let's add a note that the hint is untrusted at this point.

verifiers/internal/gha/npm.go
```diff
@@ -144,6 +141,15 @@ func (n *Npm) verifyPublishAttestationSignature() error {
...
+	npmRegistryPublicKeyID := signedPublish.PublicKey.Hint
+
+	// Retrieve the key material.
```
laurentsimon, line 147: nit: add a note that the key is now trusted

verifiers/internal/gha/npm_sigstore_tuf.go
```diff
@@ -0,0 +1,103 @@
...
+	"errors"
+	"fmt"
+	"time"
+
+	sigstoreTuf "github.com/sigstore/sigstore-go/pkg/tuf"
+)
+
+const (
+	attestationKeyUsage = "npm:attestations"
+	attestationKeyID    = "SHA256:jl3bwswu80PjjokCgh0o2w5c2U4LhQAE57gj9cz1kzA"
```
laurentsimon, line 14: is the attestationKeyID needed? Is that not the same as the pub key hint? I don't think the variable is used in the code

verifiers/internal/gha/npm_sigstore_tuf.go
```diff
@@ -0,0 +1,103 @@
...
+
+// getKeyDataWithNpmjsKeysTarget returns the target key's material, given our set of keys, return the target key's material.
+// We may also want to check the existing ValidFor.Start (and a potential future ValidFor.End).
+func getKeyDataWithNpmjsKeysTarget(keys *npmjsKeysTarget, keyID, keyUsage string) (string, error) {
+	for _, key := range keys.Keys {
+		if key.KeyID == keyID && key.KeyUsage == keyUsage {
+			return key.PublicKey.RawBytes, nil
+		}
+	}
+	return "", fmt.Errorf("%w: 'keyId': %s, 'keyUsage':%s", errorMissingNpmjsKeyIDKeyUsage, keyID, keyUsage)
```
laurentsimon, line 79: nit: you can use `%q` instead of `%s` to add `"` around the printed string.

verifiers/internal/gha/npm_sigstore_tuf.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	return &keys, nil
+}
+
+// getKeyDataWithNpmjsKeysTarget returns the target key's material, given our set of keys, return the target key's material.
+// We may also want to check the existing ValidFor.Start (and a potential future ValidFor.End).
+func getKeyDataWithNpmjsKeysTarget(keys *npmjsKeysTarget, keyID, keyUsage string) (string, error) {
+	for _, key := range keys.Keys {
+		if key.KeyID == keyID && key.KeyUsage == keyUsage {
+			return key.PublicKey.RawBytes, nil
```
laurentsimon, line 76: I suppose that's where we should eventually get the TUF validity period and validate against rekor entry.
/cc @haydentherapper

laurentsimon (approved):
LGTM, thanks!

@haydentherapper if you have comments let us know

verifiers/internal/gha/npm_sigstore_tuf.go
haydentherapper: Yep, for example it's done for Rekor like so - https://github.com/sigstore/sigstore-go/blob/main/pkg/tlog/entry.go#L274-L277

you should be able to check the attestation against the npm key validity period too - https://github.com/sigstore/root-signing/blob/main/repository/repository/targets/registry.npmjs.org/7a8ec9678ad824cdccaa7a6dc0961caf8f8df61bc7274189122c123446248426.keys.json#L9-L11

laurentsimon:
@ramonpetgrave64 feel free to merge after addressing the few nits. Please create a tracking issue for the TUF verification discussed in https://github.com/slsa-framework/slsa-verifier/pull/731#discussion_r1560402195. Thanks for the PR!

verifiers/internal/gha/npm_sigstore_tuf.go
ramonpetgrave64: thanks, removed.

verifiers/internal/gha/npm_sigstore_tuf.go
ramonpetgrave64: For now, I created the issue to track https://github.com/slsa-framework/slsa-verifier/issues/757

Issue #642, opened 2023-06-19
[refactor] Don't use go-cmp in non-test code.


ianlewis:
Currently we use go-cmp for non-test code in some verification logic. Since go-cmp is meant to be used in tests, I think we should avoid using it in non-test code.

PR #737, opened 2024-01-17
fix: #642: don't use go-cmp for outputting diff

ramonpetgrave64:
Previously we used the go-cmp's Diff for displaying a human-friendly diff between two structs in an error message.

I had intended to do a json print of the structs and do a line-by-line diff. There is an internal library for calculating text diff, but I don't see any external functions that expose it to make it available for our use: https://pkg.go.dev/golang.org/x/tools/internal/diff

Instead, this we will simply display both structs in their own "actual" and "expected" sections. The user can use their other tools to find a human-friendly diff.

### DIFF
--- a/verifiers/internal/gcb/provenance.go
+++ b/verifiers/internal/gcb/provenance.go
@@ -10,7 +10,6 @@ import (
 
 	"golang.org/x/exp/slices"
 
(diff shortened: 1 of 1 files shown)


laurentsimon (approved):
LGTM, thanks!

PR #760, opened 2024-04-23
feat: workflow to update actions dist

ramonpetgrave64:
Add a new Post-Commit workflow, to make these renovate-bot updates a bit easier.
Previously, we had to clone the PR locally, run `make package`, and then push to the PR.
Now we would just need to use the github UI to invoke this new workflow against the PR number.
We could also copy this over to the slsa-github-generator repo.

> A workflow to run against renovate-bot's PRs,
> such as `make package` after it updates the package.json and package-lock.json files.
> The potentially untrusted code is first run inside a low-privilege Job, and the diff is uploaded as an artifact.
> Then a higher-privilege Job applies the diff and pushes the changes to the PR.
> It's important to only run this workflow against PRs from trusted sources, after also reviewing the changes!

## Testing.

Tested in my own private fork, where when applicable, it pushed a commit of changes to `dist/` folders
- https://github.com/ramonpetgrave64/slsa-verifier/actions/runs/8806815483
  - https://github.com/ramonpetgrave64/slsa-verifier/pull/8/commits
- https://github.com/ramonpetgrave64/slsa-verifier/actions/runs/8806841353
  - https://github.com/ramonpetgrave64/slsa-verifier/pull/16/commits

### DIFF
--- a/.github/workflows/post-commit.yml
+++ b/.github/workflows/post-commit.yml
@@ -1,7 +1,7 @@
 # A workflow to run against renovate-bot's PRs,
 # such as `make package` after it updates the package.json and package-lock.json files.
 # The potentially untrusted code is first run inside a low-privilege Job, and the diff is uploaded as an artifact.
-# Then a higher-privilage Job applies the diff and pushes the changes to the PR.
+# Then a higher-privilege Job applies the diff and pushes the changes to the PR.
 # It's important to only run this workflow against PRs from trusted sources, after also reviewing the changes!
 
 name: Post-Commit
--- a/.github/workflows/post-commit.yml
+++ b/.github/workflows/post-commit.yml
@@ -55,7 +55,7 @@ jobs:
                   [ -z "$(cat ${{ env.ARTIFACT }})" ] && RESULT=false || RESULT=true
                   echo "patch_not_empty=$RESULT" >> "$GITHUB_OUTPUT"
             - name: upload
-              uses: actions/upload-artifact@v4
+              uses: actions/upload-artifact@5d5d22a31266ced268874388b861e4b58bb5c2f3 # v4.3.1
               with:
                   name: ${{ env.ARTIFACT }}
                   path: ${{ env.ARTIFACT }}
@@ -76,7 +76,7 @@ jobs:
               run: |
                   gh pr checkout ${{ inputs.pr_number }}
             - name: download-patch
-              uses: actions/download-artifact@v4
+              uses: actions/download-artifact@c850b930e6ba138125429b7e5c93fc707a7f8427 # v4.1.4
               with:
                   name: ${{ env.ARTIFACT }}
             - id: apply
--- a/.github/workflows/post-commit.yml
+++ b/.github/workflows/post-commit.yml
@@ -22,16 +22,6 @@ on:
                 required: true
                 type: number
 
-env:
-    COMMAND: |
-        (
-            cd ./actions/installer/dist/../ && \
-            make clean && \
-            make package
-        )
-    COMMIT_MESSAGE: "apply post-commit changes"
-    ARTIFACT: changes.patch
-
 jobs:
     diff:
         permissions:
@@ -48,23 +38,27 @@ jobs:
             - name: checkout-pr
               env:
                   GH_TOKEN: ${{ github.token }}
-              run: |
-                  gh pr checkout ${{ inputs.pr_number }}
+              run: gh pr checkout ${{ inputs.pr_number }}
             - name: run-command
-              run: ${{ env.COMMAND }}
+              run: |
+                (
+                    cd ./actions/installer/dist/../ && \
+                    make clean && \
+                    make package
+                )
             - name: diff
               id: diff
               run: |
                   git add .
                   git status
-                  git diff HEAD > ${{ env.ARTIFACT }}
-                  [ -z "$(cat ${{ env.ARTIFACT }})" ] && RESULT=false || RESULT=true
+                  git diff HEAD > changes.patch
+                  [ -z "$(cat changes.patch)" ] && RESULT=false || RESULT=true
                   echo "patch_not_empty=$RESULT" >> "$GITHUB_OUTPUT"
             - name: upload
               uses: actions/upload-artifact@5d5d22a31266ced268874388b861e4b58bb5c2f3 # v4.3.1
               with:
-                  name: ${{ env.ARTIFACT }}
-                  path: ${{ env.ARTIFACT }}
+                  name: changes.patch
+                  path: changes.patch
 
     push:
         if: needs.diff.outputs.patch_not_empty == 'true'
@@ -79,16 +73,15 @@ jobs:
             - name: checkout-pr
               env:
                   GH_TOKEN: ${{ github.token }}
-              run: |
-                  gh pr checkout ${{ inputs.pr_number }}
+              run: gh pr checkout ${{ inputs.pr_number }}
             - name: download-patch
               uses: actions/download-artifact@c850b930e6ba138125429b7e5c93fc707a7f8427 # v4.1.4
               with:
-                  name: ${{ env.ARTIFACT }}
+                  name: changes.patch
             - id: apply
               run: |
-                  git apply ${{ env.ARTIFACT }}
-                  rm ${{ env.ARTIFACT }}
+                  git apply changes.patch
+                  rm changes.patch
             # example from
             # https://github.com/actions/checkout/blob/cd7d8d697e10461458bc61a30d094dc601a8b017/README.md#push-a-commit-using-the-built-in-token
             - name: push
@@ -97,5 +90,5 @@ jobs:
                   git config user.email github-actions@github.com
                   git add .
                   git status
-                  git commit -m "${{ env.COMMIT_MESSAGE }}"
+                  git commit -m "apply post-commit changes"
                   git push
--- a/.github/workflows/post-commit.yml
+++ b/.github/workflows/post-commit.yml
@@ -1,5 +1,8 @@
 # A workflow to run against renovate-bot's PRs,
 # such as `make package` after it updates the package.json and package-lock.json files.
+# The potentially untrusted code is first run inside a low-privilege Job, and the diff is uploaded as an artifact.
+# Then a higher-privilage Job applies the diff and pushes the changes to the PR.
+# It's important to only run this workflow against PRs from trusted sources, after also reviewing the changes!
 
 name: Post-Commit
 
--- a/.github/workflows/post-commit.yml
+++ b/.github/workflows/post-commit.yml
@@ -21,8 +21,7 @@ env:
         (
             cd ./actions/installer/dist/../ && \
             make clean && \
-            make package && \
-            true
+            make package
         )
     COMMIT_MESSAGE: "apply post-commit changes"
     ARTIFACT: changes.patch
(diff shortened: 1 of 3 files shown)


ramonpetgrave64:
@laurentsimon @mihaimaruseac @ianlewis

.github/workflows/post-commit.yml
```diff
@@ -0,0 +1,94 @@
...
+jobs:
+    diff:
+        permissions:
+            pull-requests: read
+        outputs:
+            patch_not_empty: ${{ steps.diff.outputs.patch_not_empty }}
+        runs-on: ubuntu-latest
+        steps:
+            - name: checkout
+              uses: actions/checkout@v4
```
laurentsimon, line 37: nit: pin these actions

.github/workflows/post-commit.yml
```diff
@@ -0,0 +1,94 @@
...
+              with:
+                  repository: ${{ github.repository }}
+                  persist-credentials: false
+            - name: checkout-pr
+              env:
+                  GH_TOKEN: ${{ github.token }}
+              run: |
+                  gh pr checkout ${{ inputs.pr_number }}
+            - name: run-command
+              run: ${{ env.COMMAND }}
```
laurentsimon, line 47: I think it's fine to copy the command instead of using env.COMMAND. Or is there a reason for using an env variable here but not in the diff step?

.github/workflows/post-commit.yml
```diff
@@ -0,0 +1,94 @@
+# A workflow to run against renovate-bot's PRs,
+# such as `make package` after it updates the package.json and package-lock.json files.
+# The potentially untrusted code is first run inside a low-privilege Job, and the diff is uploaded as an artifact.
+# Then a higher-privilege Job applies the diff and pushes the changes to the PR.
+# It's important to only run this workflow against PRs from trusted sources, after also reviewing the changes!
+
+name: Post-Commit
+
```
laurentsimon, line 8: declare permissions as `{}` in the workflow

.github/workflows/post-commit.yml
```diff
@@ -0,0 +1,94 @@
...
+                  name: ${{ env.ARTIFACT }}
+                  path: ${{ env.ARTIFACT }}
+
+    push:
+        if: needs.diff.outputs.patch_not_empty == 'true'
+        needs: diff
+        runs-on: ubuntu-latest
+        permissions:
+            pull-requests: read
+            contents: write
```
laurentsimon, line 68: Add a comment saying we use CODEOWNERS and it's enforced thru Branch rules, so the token cannot be used to push to the main and release branches. 

@mihaimaruseac can you take a look at this PR? We want to update the PR branch for renovatebot PRs

laurentsimon (approved):
LGTM. Added a few nits.

.github/workflows/post-commit.yml
mihaimaruseac: +1

.github/workflows/post-commit.yml
```diff
@@ -0,0 +1,94 @@
...
+                required: true
+                type: number
+
+env:
+    COMMAND: |
+        (
+            cd ./actions/installer/dist/../ && \
+            make clean && \
+            make package && \
+            true
```
mihaimaruseac, line 23: I don't understand why `true` is needed here. If `make package` fails, `true` won't get executed. If `make package` succeeds, `true` doesn't change the return status

mihaimaruseac (commented):
This seems a little bit more complex. Can this be done similar to https://github.com/guacsec/guac/pull/953#issuecomment-1599090212 ?

.github/workflows/post-commit.yml
ramonpetgrave64: No real reason other than it looks nicer to have the COMMAND as a veriable at the top of this document.

.github/workflows/post-commit.yml
ramonpetgrave64: added

.github/workflows/post-commit.yml
ramonpetgrave64: It's not needed. removed

.github/workflows/post-commit.yml
ramonpetgrave64: added

ramonpetgrave64:
> This seems a little bit more complex. Can this be done similar to [guacsec/guac#953 (comment)](https://github.com/guacsec/guac/pull/953#issuecomment-1599090212) ?

I used the `gh` cli because it has some conveniences like setting up the correct remote branch for the PR. If we used `actions/checkout` within a normal `pull_request` event, then we wouldn't need to use `gh` cli.

ramonpetgrave64:
@kpk47 And I added some more Repo rules that make this workflow even safer to use. Safer than cloning the PR locally and running the untrusted code.

.github/workflows/post-commit.yml
ramonpetgrave64: changed.

.github/workflows/post-commit.yml
```diff
@@ -0,0 +1,94 @@
...
+# The potentially untrusted code is first run inside a low-privilege Job, and the diff is uploaded as an artifact.
+# Then a higher-privilege Job applies the diff and pushes the changes to the PR.
+# It's important to only run this workflow against PRs from trusted sources, after also reviewing the changes!
+
+# There have been vulnerabilities with using `git apply` https://github.blog/2023-04-25-git-security-vulnerabilities-announced-4/
+# At this point a compromised git binary cannot modify any of this repo's branches, only the PR fork's branch,
+# due to our branch protection rules and CODEOWNERS.
+# It aslso cannot submit a new release or modify exsiting releases due to tag protection rules.
+
+name: Post-Commit
```
ianlewis, line 13: nit: consider naming this something like 'Update action dist post-commit' since we expect this to be a manual thing. The name could be something better, but I think 'Post-Commit' isn't very descriptive.

.github/workflows/post-commit.yml
```diff
@@ -0,0 +1,94 @@
...
+        inputs:
+            pr_number:
+                description: "The pull request number."
+                required: true
+                type: number
+
+jobs:
+    diff:
+        permissions:
+            pull-requests: read
```
ianlewis, line 28: Similar to comment below, just add a NOTE comment here and below to the `permissions` sections indicating the intention that `diff` should be low permissions since it executes code while `push` has more permissions since it doesn't execute arbitrary code from the repo.

.github/workflows/post-commit.yml
```diff
@@ -0,0 +1,94 @@
+# A workflow to run against renovate-bot's PRs,
```
ianlewis, line 1: nit: add a doc section to the CONTRIBUTING.md that outlines how to use this workflow.

.github/workflows/post-commit.yml
```diff
@@ -0,0 +1,94 @@
...
+        permissions:
+            pull-requests: read
+            contents: write
+        steps:
+            - name: checkout
+              uses: actions/checkout@v4
+            - name: checkout-pr
+              env:
+                  GH_TOKEN: ${{ github.token }}
+              run: gh pr checkout ${{ inputs.pr_number }}
```
ianlewis, line 76: After `git pr checkout`, what is the git remote and branch set to? i.e. where does a `git push` go to?

.github/workflows/post-commit.yml
```diff
@@ -0,0 +1,94 @@
...
+# such as `make package` after it updates the package.json and package-lock.json files.
+
+# The potentially untrusted code is first run inside a low-privilege Job, and the diff is uploaded as an artifact.
+# Then a higher-privilege Job applies the diff and pushes the changes to the PR.
+# It's important to only run this workflow against PRs from trusted sources, after also reviewing the changes!
+
+# There have been vulnerabilities with using `git apply` https://github.blog/2023-04-25-git-security-vulnerabilities-announced-4/
+# At this point a compromised git binary cannot modify any of this repo's branches, only the PR fork's branch,
+# due to our branch protection rules and CODEOWNERS.
+# It aslso cannot submit a new release or modify exsiting releases due to tag protection rules.
```
ianlewis, line 11: Is there any reason this needs to run on the `slsa-framework/slsa-verifier` repo at all? could contributors just run it manually on their fork? That would somewhat limit exposure on our side and reduces toil for reviewers (i.e. if the dist check fails we could just tell them "run this" and point to the doc on how to run it in their fork. better yet, the error from that check could tell them).

 AFAICT, this would just checkout the forked PR and push to the appropriate branch on the forked repo so wouldn't need to interact with (and thus need permissions for) the main repo at all.

Also makes it easier for us to audit since we would basically just makes sure there aren't any runs for this action on the main repo

.github/workflows/post-commit.yml
ramonpetgrave64: The remote and branch get set to the PRs repo and branch, and it pushes there.

.github/workflows/post-commit.yml
ramonpetgrave64: This is mainly intended to run on PRs from renovate-bot's fork. We wouldn't be able to invoke their fork's workflows. 

Example:
 - https://github.com/slsa-framework/slsa-verifier/pull/717
 - https://github.com/renovate-bot/slsa-verifier/tree/renovate/npm
 - https://github.com/renovate-bot/slsa-verifier/actions/workflows/pre-submit.actions.yml

.github/workflows/post-commit.yml
ramonpetgrave64: changed.

.github/workflows/post-commit.yml
ramonpetgrave64: added.

.github/workflows/post-commit.yml
ramonpetgrave64: No existing contributing docs, but I added one.

.github/workflows/update-actions-dist-post-commit.yml
```diff
@@ -0,0 +1,96 @@
...
+    diff:
+        permissions:
+            # This Job executes the PR's untrusted code, so it must how low permissions.
+            pull-requests: read
+        outputs:
+            patch_not_empty: ${{ steps.diff.outputs.patch_not_empty }}
+        runs-on: ubuntu-latest
+        steps:
+            - name: checkout
+              uses: actions/checkout@v4
```
mihaimaruseac, line 35: This should still be pinned by hash

.github/workflows/update-actions-dist-post-commit.yml
```diff
@@ -0,0 +1,96 @@
...
+        if: needs.diff.outputs.patch_not_empty == 'true'
+        needs: diff
+        runs-on: ubuntu-latest
+        permissions:
+            # This Job does not run untrusted code, but it does need to push changes to the PR's branch.
+            pull-requests: read
+            contents: write
+        steps:
+            - name: checkout
+              uses: actions/checkout@v4
```
mihaimaruseac, line 74: Needs to be pinned

mihaimaruseac (approved):

.github/workflows/update-actions-dist-post-commit.yml
ramonpetgrave64: fixed

.github/workflows/update-actions-dist-post-commit.yml
ramonpetgrave64: fixed

mihaimaruseac (approved):

ramonpetgrave64:
Thanks everyone for the reviews

PR #761, opened 2024-04-25
fix: make download-artifacts.sh more flexible

ramonpetgrave64:
Making the `download-artifacts.sh` script be more useful.

Before, it would error upon seeing some zip files that it doesn't expect to be in the GH release. I think the script is just a bit outdated. But for now, I think we should bypass that, since the script is already written to know which of the final files within the archives are actually needed.

related PR https://github.com/slsa-framework/slsa-github-generator/pull/3589


ramonpetgrave64:
@mihaimaruseac

mihaimaruseac (approved):

ramonpetgrave64:
@ianlewis

download-artifacts.sh
```diff
@@ -88,6 +88,13 @@ unzip_files() {
         rm -rf "${tmp_dir}"
         ;;
 
+    # catch-all for any zip file
+    # TODO: With the commands at the end of this script, only the needed files will be copied over.
+    # But we should get specific about which files to ignore.
+    ./*.zip)
+        unzip -o "${zip_path}" -d "${output_path}"
```
ianlewis, line 95: nit: I think this case makes some of the above cases obsolete.

ianlewis (approved):

PR #768, opened 2024-05-08
[draft] feat: VerifyNpmPackage API

ramonpetgrave64:
[Draft]

- Add a new `VerifyNpmPackageWithSigstoreTufClient` method that allows supplying a SigstoreTufClient
  - TODO: confirm it works with the inner tuf clients
  - TODO: confirm it works with local tuf roots?
- TODO: offline rekor verification
  - TODO: investigate existing functionality
    - https://github.com/ramonpetgrave64/slsa-verifier/blob/a2edf56f612f6ef4baef66fb07cd62988432d357/verifiers/internal/gha/provenance.go#L217-L229
  - TODO check prior arts
    - https://github.com/sigstore/cosign/blob/main/README.md#verify-a-container-in-an-air-gapped-environment
    - https://github.com/sigstore/sigstore-go/blob/75b405ded9aee87afa3c29facd08effbd35e43b2/pkg/verify/signed_entity.go#L39-L40


ramonpetgrave64:
@slugclub

verifiers/verifier_regression_test.go
```diff
@@ -0,0 +1,132 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+
+	sigstoreTuf "github.com/sigstore/sigstore-go/pkg/tuf"
+	serrors "github.com/slsa-framework/slsa-verifier/v2/errors"
+	"github.com/slsa-framework/slsa-verifier/v2/options"
+)
+
+const TEST_DIR = "../cli/slsa-verifier/testdata"
```
slugclub, line 21: It might be worth creating a new `testdata/` folder in the folder this test file is in and copying the subset of test data you need from `.../cli/slsa-verifier/testdata` into it.

Having this test depend directly on the test data in `.../cli/slsa-verifier/testdata` could make maintenance more difficult in the future e.g. if someone wants to change the test data for the cli or move the `cli/slsa-verifier` folder it will break these tests. It might seem a bit wasteful to have two separate copies of the data but as the Go proverb goes "A little copying is better than a little dependency." (https://go-proverbs.github.io/).

verifiers/internal/gha/verifier.go
```diff
@@ -332,12 +333,32 @@ func (v *GHAVerifier) VerifyNpmPackage(ctx context.Context,
...
+}
+
+// VerifyNpmPackage verifies an npm package tarball.
+func (v *GHAVerifier) VerifyNpmPackageWithSigstoreTufClient(ctx context.Context,
+	attestations io.Reader, tarballHash string,
+	provenanceOpts *options.ProvenanceOpts,
+	builderOpts *options.BuilderOpts,
+	sigstoreTufClient utils.SigstoreTufClient,
+) ([]byte, *utils.TrustedBuilderID, error) {
+	atts, err := io.ReadAll(attestations)
```
slugclub, line 351: Sorry I know we had chatted previously about using an `io.Reader` here (and there are many cases where that is the right thing to do). But the first thing this function does with the reader is to call `io.ReadAll`, so it actually might be better to just pass in a `[]byte` array. What's here now essentially converts a `[]byte` array to an `io.Reader` to call the function and then the function converts it back to a `[]byte` array. So we can avoid the middle step and just use `[]byte` as the function argument.

slugclub (commented):
This is looking great, thanks so much for working on this. I have a few minor nits but overall it's looking good.

ramonpetgrave64:
> This is looking great, thanks so much for working on this. I have a few minor nits but overall it's looking good.

Thanks for the review! I was also looking into logging in https://github.com/slsa-framework/slsa-verifier/pull/772,

verifiers/internal/gha/verifier.go
```diff
@@ -332,12 +333,32 @@ func (v *GHAVerifier) VerifyNpmPackage(ctx context.Context,
 	provenanceOpts *options.ProvenanceOpts,
 	builderOpts *options.BuilderOpts,
 ) ([]byte, *utils.TrustedBuilderID, error) {
+	atts := strings.NewReader(string(attestations))
```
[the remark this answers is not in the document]
ramonpetgrave64: alright. changed it back to bytes

docs/API-Library.md
```diff
@@ -0,0 +1,73 @@
...
+)
+
+func main() {
+	doVerify()
+}
+
+func doVerify() (*apiUtils.TrustedBuilderID, error) {
+	packageVersion := "0.1.127"
+	packageName := "@ianlewis/actions-test"
+	builderId := "https://github.com/slsa-framework/slsa-github-generator/.github/workflows/builder_nodejs_slsa3.yml"
```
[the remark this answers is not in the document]
ramonpetgrave64: changed

docs/API-Library.md
```diff
@@ -0,0 +1,73 @@
...
+		ExpectedID: &builderId,
+	}
+	// example: force using the embedded root, without going online for a refresh
+	// opts := sigstoreTuf.DefaultOptions().WithForceCache()
+	// example: supply your own root
+	// opts := sigstoreTuf.DefaultOptions().WithRoot([]byte(`{"signed":{"_type":"root","spec_version":"1.0","version":9,"expires":"2024-09-12T06:53:10Z","keys":{"1e1d65ce98b10 ...`)).WithForceCache()
+	opts := sigstoreTuf.DefaultOptions()
+	client, err := sigstoreTuf.New(opts)
+	if err != nil {
+		fmt.Printf("creating SigstoreTuf client: %v", err)
```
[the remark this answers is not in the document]
ramonpetgrave64: changed.

docs/API-Library.md
```diff
@@ -0,0 +1,73 @@
...
+	"fmt"
+
+	sigstoreTuf "github.com/sigstore/sigstore-go/pkg/tuf"
+	options "github.com/slsa-framework/slsa-verifier/v2/options"
+	apiVerify "github.com/slsa-framework/slsa-verifier/v2/verifiers"
+	apiUtils "github.com/slsa-framework/slsa-verifier/v2/verifiers/utils"
+)
+
+func main() {
+	doVerify()
```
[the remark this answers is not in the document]
ramonpetgrave64: I'll do it this way.

verifiers/verifier_regression_test.go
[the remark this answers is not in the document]
ramonpetgrave64: changed.

ramonpetgrave64: added.

verifiers/internal/gha/verifier.go
ramonpetgrave64: changed

verifiers/internal/gha/npm_sigstore_tuf_test.go
```diff
@@ -87,9 +90,9 @@ type mockSigstoreTufClient struct {
 	fileContentMap map[string]string
 }
 
-// NewMockSigstoreTufClient returns an instance of the mock client,
+// newMockSigstoreTufClient returns an instance of the mock client,
```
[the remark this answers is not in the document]
ramonpetgrave64: I'll be considering that. right now, the mock is very simple, so I guess folks can just copy-paste it.

slugclub (approved):
Looks good. A few minor nits but overall LGTM

docs/API-Library.md
```diff
@@ -27,19 +29,22 @@ import (
...
+	fmt.Println("Verifying npm package: PASSED")
 }
 
 func doVerify() (*apiUtils.TrustedBuilderID, error) {
 	packageVersion := "0.1.127"
 	packageName := "@ianlewis/actions-test"
-	builderId := "https://github.com/slsa-framework/slsa-github-generator/.github/workflows/builder_nodejs_slsa3.yml"
-    attestationsReader, err := os.Open("../my/attestations.json")
+	builderID := "https://github.com/slsa-framework/slsa-github-generator/.github/workflows/builder_nodejs_slsa3.yml"
+	attestations, err := os.ReadFile("../stuff/attestations.json")
```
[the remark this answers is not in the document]
ramonpetgrave64: changed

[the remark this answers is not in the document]
ramonpetgrave64: changed

verifiers/internal/gcb/verifier.go
```diff
@@ -48,6 +48,16 @@ func (v *GCBVerifier) VerifyNpmPackage(ctx context.Context,
 	return nil, nil, serrors.ErrorNotSupported
 }
 
+// VerifyNpmPackage verifies an npm package tarball.
```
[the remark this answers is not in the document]
ramonpetgrave64: fixed

verifiers/utils/sigstore_tuf.go
```diff
@@ -0,0 +1,21 @@
+package utils
+
+import (
+	"fmt"
+
+	sigstoreTuf "github.com/sigstore/sigstore-go/pkg/tuf"
+)
+
+type SigstoreTufClient interface {
```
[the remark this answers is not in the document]
ramonpetgrave64: changed.

docs/API-Library.md
```diff
@@ -27,19 +29,22 @@ import (
 )
 
 func main() {
-	doVerify()
+	builderID, err := doVerify()
+	if err != nil {
+		log.Fatalf("Verifying npm package: FAILED: %w", err)
+	}
+	fmt.Printf("builderID: %s\n", builderID.Name())
+	fmt.Println("Verifying npm package: PASSED")
```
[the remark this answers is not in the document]
ramonpetgrave64: changed

ramonpetgrave64:
@slugclub thanks again. @ianlewis @laurentsimon , please take a look

register/register.go
```diff
@@ -34,6 +34,12 @@ type SLSAVerifier interface {
 		provenanceOpts *options.ProvenanceOpts,
 		builderOpts *options.BuilderOpts,
 	) ([]byte, *utils.TrustedBuilderID, error)
+
+	VerifyNpmPackageWithSigstoreTUFClient(ctx context.Context,
```
haydentherapper, line 38: As we chatted about offline, we may want this API to take the npm verification material directly and another API responsible for fetching this target file from TUF. Then this API could verify metadata offline, and the latter API would be responsible for network calls. I think this would also make testing easier since you don't have to mock the Sigstore TUF client, just need the verification material struct initialized.

Another change, outside of the scope of this PR, would be to update https://github.com/slsa-framework/slsa-verifier/blob/main/verifiers/internal/gha/trusted_root.go to use [TrustedMaterial](https://github.com/sigstore/sigstore-go/blob/main/pkg/root/trusted_material.go) rather than Cosign's APIs. Like you're already doing in https://github.com/slsa-framework/slsa-verifier/blob/main/verifiers/internal/gha/bundle.go#L162, rather than passing the TUF client, you would pass in the trusted root material.

verifiers/internal/gha/npm.go
```diff
@@ -134,7 +134,7 @@ func (n *Npm) verifyProvenanceAttestationSignature() error {
 	return nil
 }
 
-func (n *Npm) verifyPublishAttestationSignature() error {
+func (n *Npm) verifyPublishAttestationSignature(sigstoreTUFClient utils.SigstoreTUFClient) error {
```
haydentherapper, line 137: Rather than change the function, could the `Npm` struct be initialized with a TUF client (or the npm verification metadata)?

verifiers/utils/sigstore_tuf.go
```diff
@@ -0,0 +1,21 @@
...
+	sigstoreTUF "github.com/sigstore/sigstore-go/pkg/tuf"
+)
+
+type SigstoreTUFClient interface {
+	GetTarget(target string) ([]byte, error)
+}
+
+// NewSigstoreTUFClient gets a Sigstore TUF client, which itself is a wrapper around the official TUF client.
+func NewSigstoreTUFClient() (*sigstoreTUF.Client, error) {
+	opts := sigstoreTUF.DefaultOptions()
```
haydentherapper, line 15: I'd recommend we add `CacheValidity` to the default so that repeated verifications in a short time window won't result in constant TUF repo refreshes.

register/register.go
ramonpetgrave64: Speaking about this offline again, we have our own struct to parse the material, but the spec/layout of the material isn't defined in npmjs docs. Until we can be sure that npmjs will commit to this layout, I think it's better to hide this detail from the user, or let the user stub out their own `SigstoreTUFClient`.
 - https://docs.npmjs.com/about-registry-signatures#supporting-signatures-on-third-party-registries

verifiers/internal/gha/npm.go
ramonpetgrave64: We could, but it doesn't save much space, in terms on code.

verifiers/utils/sigstore_tuf.go
ramonpetgrave64: I want to keep this as the default, because like we've had in other conversations, the TUF spec requires updates every time you fetch. So I'd prefer to be compliant by default. But also we keep an in-memory cache of the target key, so it already does avoid repeated fetches.

verifiers/internal/gha/npm.go
haydentherapper: Given there are multiple functions that take in a TUF client, I think it would be more idiomatic to have the `Npm` struct include the TUF client, though `getAttestationKey` would need to be a method on the struct. Not strongly opinionated on this.

register/register.go
haydentherapper: I don't think we should recommend implementing a `SigstoreTUFClient` to bypass TUF. I think this API is fine for the current use case, though taking in the verification material (keys) directly could be another way to implement this API while not requiring knowledge of the npm struct.

verifiers/utils/sigstore_tuf.go
haydentherapper: To confirm, how long is the in-memory cache?

register/register.go
laurentsimon: in the long term this function should use variadic options / arguments. For a quick turnaround, maybe you could add a pointer to a tuf client in the existing builderOpts https://github.com/slsa-framework/slsa-verifier/blob/main/options/options.go#L36.

verifiers/internal/gcb/verifier.go
```diff
@@ -48,6 +48,16 @@ func (v *GCBVerifier) VerifyNpmPackage(ctx context.Context,
...
+func (v *GCBVerifier) VerifyNpmPackageWithSigstoreTufClient(ctx context.Context,
+	attestations []byte, tarballHash string,
+	provenanceOpts *options.ProvenanceOpts,
+	builderOpts *options.BuilderOpts,
+	sigstoreTufClient utils.SigstoreTufClient,
+) ([]byte, *utils.TrustedBuilderID, error) {
```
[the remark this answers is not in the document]
laurentsimon: I think so.

> if it should return the data in a more structured manner

@ramonpetgrave64 That's maybe what we discussed a few weeks ago, to have a layered approach to verification where the inner layer returns a structured "condense" / summary for callers to use.

docs/API-Library.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Verifers
+
+We have exported functions for using slsa-verifier within your own golang packages
+
+- slsa-verifier/verifiers/verifier.go
+
+### Npmjs
+
+With `VerifyNpmPackageWithSigstoreTUFClient`, you can pass in your own TUF client with custom options.
```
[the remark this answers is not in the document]
ramonpetgrave64: I wouldn't call it a hack. But your other comment got me thinking I could expose the `Npm` struct externally, or make a new `NpmVerifier` struct. And we could still use that TUFClient to let the user pass in their own [TUF root](https://github.com/haydentherapper/sigstore-go/blob/1bb1ba649570efaa5e7f0c6bb06d82449e2b406c/pkg/root/trusted_root.go#L292).

```
type NpmVerifier struct {
	Ctx              context.Context
	AttestationBytes []byte
	PovenanceOpts    *options.ProvenanceOpts
	BuilderOpts      *options.BuilderOpts
	VerifierOpts     *options.NpmVerifierOpts // hypothetical new type

}

type VerifierOpts struct { // hypothetical new type
	Logger  *log.Logger
	offline bool // whether to force offline verification
}

type NpmVerifierOpts struct { // hypothetical new type
	*VerifierOpts
	TUFClient *utils.TUFClient
}
```

register/register.go
ramonpetgrave64: If you're referring to using variadic options like in this blog post, I decided on a bit of a hybrid approach in my latest commits.
- https://petomalina.medium.com/dealing-with-optional-parameters-in-go-9780f9bfbd1d

docs/API-Library.md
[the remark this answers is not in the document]
ramonpetgrave64: @laurentsimon Alright I decided to do both this new struct and the variadic arguments in options.go
 - https://github.com/slsa-framework/slsa-verifier/pull/768/files#diff-76f121a3a81c8945dc0e0c81a4b95e941d252522eba8704a1aa6f6a1722a8b68R45-R61

docs/API-Library.md
```diff
@@ -0,0 +1,77 @@
...
+	// client, err := apiUtils.NewSigstoreTUFClient()
+	opts := sigstoreTUF.DefaultOptions()
+	client, err := sigstoreTUF.New(opts)
+	if err != nil {
+		return nil, fmt.Errorf("creating SigstoreTUF client: %w", err)
+	}
+	verifierOptioners := []options.VerifierOptioner{
+		options.WithSigstoreTUFClient(client),
+	}
+	_, outBuilderID, err := apiVerify.VerifyNpmPackageWithSigstoreTUFClient(context.Background(), attestations, tarballHash, provenanceOpts, builderOpts, verifierOptioners...)
```
ramonpetgrave64, line 71: @slugclub small change in the interface. We're passing in some variadic arguments, instead of passing in the the client directly.

slugclub: +1 to using options to set the TUF client. Is there a reason to pass this in as a variadic `options.VerifierOptioner` rather than something like a `*options.VerifierOpts` like what we have on lines 332 and 333 for `provenanceOpts`/`builderOpts`?

verifiers/internal/gha/npm.go
```diff
@@ -66,25 +71,62 @@ func (n *Npm) ProvenanceLeafCertificate() *x509.Certificate {
...
 		ctx:  ctx,
 		root: root,
 
 		provenanceAttestation: prov,
 		publishAttestation:    pub,
+		verifierOpts:          verifierOpts,
 	}, nil
 }
 
+// getDefaultSigstoreTUFClient returns the default Sigstore TUF client.
```
slugclub, line 98: It might be good to export a function that gives the default options so its more explicit to an interface user what's going on? Like what TUF does: https://github.com/sigstore/sigstore-go/blob/v0.4.0/pkg/tuf/options.go#L121.

options/options.go
```diff
@@ -37,3 +41,21 @@ type BuilderOpts struct {
...
+
+// VerifierOpts are the options for the verifier, created with the VerifierOptioner functions.
+// In the future, this can include a logger and a rekor client,
+// or be a standalone argument to the verifier functions.
+type VerifierOpts struct {
+	// SigstoreTufClient is the Sigstore TUF client, used for retrieving the Npmjs public keys
+	SigstoreTUFClient apiUtils.SigstoreTUFClient
+}
+
+// VerifierOptioner is a function that sets options for the verifier.
```
slugclub, line 53: Is there any reason to use this approach for creating options as opposed to something like:
```
verifierOpts := &options.VerifierOpts{
  SigstoreTUFClient: client,
  ... // Anything else that needs to be set up
}
```
in the client (and then potentially a `DefaultVerifierOptions` func if needed - see comment below)? I think that would be more consistent with `builderOpts`/`provenanceOpts` and clearer for clients to understand?

docs/API-Library.md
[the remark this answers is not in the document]
slugclub: Another question following from Laurent's - how much leeway do we have to change this API? Does it have many users currently? Would it be possible to make a new release with a breaking change? (Not saying we necessarily have to do that just wondering what the constraints are here).

options/options.go
ramonpetgrave64: I agree it would be more clear to understand, but my reason is so that we won't me merging in a breaking change at this time.

verifiers/internal/gha/npm.go
ramonpetgrave64: Added

haydentherapper:
cc @loosebazooka

docs/API-Library.md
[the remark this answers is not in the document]
ramonpetgrave64: I suppose we could have lots of leeway. Personally, I'd rather we get this into v2, and then later on add the breaking changes in a new v3.

docs/API-Library.md
```diff
@@ -0,0 +1,82 @@
...
+	// example: supply your own default client
+	opts := sigstoreTUF.DefaultOptions()
+	client, err := sigstoreTUF.New(opts)
+	if err != nil {
+		return nil, fmt.Errorf("creating SigstoreTUF client: %w", err)
+	}
+	verifierOptioners := []options.VerifierOptioner{
+		options.WithSigstoreTUFClient(client),
+	}
+	_, outBuilderID, err := apiVerify.VerifyNpmPackageWithSigstoreTUFClient(context.Background(), attestations, tarballHash, provenanceOpts, builderOpts, verifierOptioners...)
```
haydentherapper, line 76: Need to update doc, `VerifyNpmPackageWithSigstoreTUFClient` no longer exists

verifiers/utils/sigstore_tuf.go
```diff
@@ -0,0 +1,31 @@
...
+func GetDefaultSigstoreTUFClient() (SigstoreTUFClient, error) {
+	value := defaultSigstoreTUFClientAtomicValue.Load()
+	if value != nil {
+		return value.(SigstoreTUFClient), nil
+	}
+	sigstoreTUFClient, err := sigstoreTUF.DefaultClient()
+	if err != nil {
+		return nil, err
+	}
+	defaultSigstoreTUFClientAtomicValue.Store(sigstoreTUFClient)
```
haydentherapper, line 29: Do you want an atomic or https://pkg.go.dev/sync#Once? Before the TUF client is initialized, if you had multiple threads calling `GetDefaultSigstoreTUFClient`, then `Load` may return `nil` for each thread, and multiple clients will be created.

docs/API-Library.md
ramonpetgrave64: fixed

verifiers/utils/sigstore_tuf.go
ramonpetgrave64: Thanks, I think this is a nicer way

verifiers/internal/gcb/verifier.go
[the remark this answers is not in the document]
ramonpetgrave64: Added it in a new commit
 - https://github.com/slsa-framework/slsa-verifier/pull/768/commits/b4381c50060c8ae0f01e24e2ac61b0640c6ef9f6

options/options.go
slugclub: Using variadic seems OK if we need to not make a breaking change to the api. In terms of using an `Optioner` vs something like `options.VerifierOpts`, using the latter might be better to keep consistent with the way we provide the other options (possibly including a `DefaultVerifierOptions` function to make it easy for clients to get the standard set of options). Or are there pros for using an `Optioner` that I'm missing? (I'm not very familiar with that pattern sorry)

ramonpetgrave64: I'm also new to the pattern, but I see it's used a bit in sigstore-go for making optional arguments. (I prefer to call the returned func an "optioner", instead of an "option".)
 - https://github.com/sigstore/sigstore-go/blob/87bfa5b1c9fcf161c773559f3b470a36ddea4552/pkg/verify/signed_entity.go#L123
 - https://github.com/sigstore/sigstore-go/blob/87bfa5b1c9fcf161c773559f3b470a36ddea4552/cmd/conformance/main.go#L306

I could change it to accept a variadic `options.VerifierOpts`, but then I'd need code to check that the receiver has no more than one of `options.VerifierOpts`. This seems problematic. And If you recall, that was my original approach, but now it seems we could agree the variadic option funcs seem to be a good compromise to making a whole new `VerifyNpmPackageWithSigstoreTUFClient`.

```
func (v *GCBVerifier) VerifyNpmPackage(ctx context.Context,
	attestations []byte, tarballHash string,
	verifierOpts ...options.VerifierOpts,
) ([]byte, *utils.TrustedBuilderID, error) {
    if len(verifierOpts) > 1 {
        panic("There should be at most one instance of verifierOpts supplied.")
    }
    ...
}
```

slugclub: Hmm yeah I agree that checking there's at most one `options.VerifierOpts` isn't the nicest.

Where I'm getting stuck is that ideally the API function and its arguments would provide a "well-lit path" to users i.e. the function signature would make it easy for clients to:
a) understand what the function is doing
b) call the function correctly/with their desired arguments.

I think right now it's more complex than it needs to be. There's three different sets of options you have to pass in and the options themselves are specified two different ways. I know that's a by-product of hacking on the existing API (which I don't believe was necessarily designed to be part of an external library).

I don't want to block this change from merging - I know it's been in the works for a while now and I definitely think this restructuring to allow a user to pass in a TUF client is a big improvement. Maybe for the future it be good to either:

- Introduce an additional function with a simpler signature (I'm curious to hear Laurent explain why that might be a bad idea)
- Make a breaking change to this API

We should also make sure the example documentation is super clear.

ramonpetgrave64:
@loosebazooka

loosebazooka:
On the sigstore clients we actually just accept a [trusted_root.json](https://github.com/sigstore/root-signing/blob/main/targets/trusted_root.json) as an input to the verifier -- this would require depending on sigstore-go instead of cosign and I'm sure where we're at here. However this eliminates the need for consumers to define a whole TUF client, they may compute `trusted_root.json` some other way.

haydentherapper:
I could see a need for both - a user may gather their own root material out of band, at which point they should pass that material via a trusted root file, to align with the other Sigstore clients (and sigstore-go has an API for this). A developer integrating with the API may be opinionated on TUF settings and construct their own client (or pass options).

ramonpetgrave64:
I just don't think slsa-verifier should be involved in the lower-level details of the TUF client.

options/options.go
ramonpetgrave64: @loosebazooka  It might be easier to see this example from an earlier commit
 - https://github.com/slsa-framework/slsa-verifier/pull/768/commits/e4c034adebe6b3e2fb72f7a6e200f38370217105#diff-2b75bff58ca87cf24fae495867d0eb3a039b63b9505b885b705ea55720777cbcR330-R339

Here I was accepting a variadic `...utils.SigstoreTufClient`, which is really a slice of clients. And so it needed an awkward input validation, to make sure the caller was only passing in exactly one client.

```
func (v *GHAVerifier) VerifyNpmPackage(ctx context.Context,
	attestations []byte, tarballHash string,
	sigstoreTufClient ...utils.SigstoreTufClient,
) ([]byte, *utils.TrustedBuilderID, error) {
	if len(sigstoreTufClient) != 1 {
		return nil, nil, fmt.Errorf("%w: length of variadic sigstoreTufClient must be exactl 1", serrors.ErrorInternal)
	}
    ...
```

And in the previous post `options.VerifierOpts` is a struct containing the actual options. Instead of a variadic `...options.VerifierOpts`, which requires another awkward check, I chose to pass in a variadic `...options.VerifierOptioner`, which is a slice of functions that will set the actual options in a `options.VerifierOpts` structs that is internal to the receiver.

haydentherapper:
> I just don't think slsa-verifier should be involved in the lower-level details of the TUF client.

Could you say by what you mean on lower-level details? Do you mean the configuration of the TUF client?

In terms on an API, I think we could see a few different options:

* No changes, accept default - This will be for most users imo.
* Providing trust root material directly - This is useful for private deployments or private PKI, and the API take in Sigstore's trust root format. I think this would be an alternative to the proposed API that involves implementing a client that fetches targets.
* A user who wants to configure certain settings for the TUF client - This would be a user who very much understands the implementation details of TUF and wants to control caching, which I don't think is likely, but could happen. An alternative is providing a TUF client.

options/options.go
loosebazooka: oh right, I forgot go doesn't allow function overloading. But are we only using `...` so that we can send 0 or 1 options and not break old users ? It seems like maybe we should be okay breaking the API for this. It'll get caught at compile time and is an easy fix? Get rid of the variadic and force users to provide options or empty.

docs/API-Library.md
```diff
@@ -0,0 +1,84 @@
+# Api Library
+
+## Verifers
+
+We have exported functions for using slsa-verifier within your own golang packages
```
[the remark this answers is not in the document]
ramonpetgrave64: fixed

ramonpetgrave64:
> Could you say by what you mean on lower-level details? Do you mean the configuration of the TUF client?

I mean the trusted_root.json file, and managing potential refreshes in a long-running process.

> * A user who wants to configure certain settings for the TUF client - This would be a user who very much understands the implementation details of TUF and wants to control caching, which I don't think is likely, but could happen. An alternative is providing a TUF client.

But a user still has to know how to prepare a trusted_root.json properly. Ideally, the Sigstore TUF client can handle those cache-control options and make it easy for the user.

I would rather the user instead [supply](https://pkg.go.dev/github.com/sigstore/sigstore-go@v0.5.1/pkg/tuf#Options.WithRoot) the trusted_root.json to their own TUF client, and then supply that to slsa-verifier.

For the CLI, if we wanted the same kind of customization, then supplying the trusted_root.json directly makes more sense, but we may still have to worry about refreshes. We can consider that feature later if there is community ask for it.

haydentherapper:
I think there may be some conflation of "roots". https://pkg.go.dev/github.com/sigstore/sigstore-go@v0.5.1/pkg/tuf#Options.WithRoot refers to the TUF root, as in the [root of trust](https://github.com/sigstore/root-signing/blob/main/repository/repository/root.json) for the TUF metadata. The ["trusted root"](https://pkg.go.dev/github.com/sigstore/sigstore-go@v0.5.1/pkg/root#NewTrustedRootProtobuf) is the [Sigstore spec-compliant file](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_trustroot.proto) that contains all of the roots of trust for the services within Sigstore's infra.

Regarding "we may still have to worry about refreshes", in the case that someone would supply a trusted root file directly, that effectively says "bypass TUF, don't worry about keeping metadata up to date".

For a user, they could either supply a TUF client which goes and fetches a "trusted root", or they supply the "trusted root" directly bypassing TUF (if they don't maintain their own TUF repo, for example, or want offline verification).

As for TUF caching options, I think that's mostly up to the library, and I don't think we must expose that right of the bat. Some users might be opinionated, so for example in Cosign, we allow disabling the cache, but we don't expose every TUF client option.

ramonpetgrave64:
Alright. For now, it's simpler and more flexible to the user for slsa-verifier to accept a sigstore TUF client.

options/options.go
slugclub: Big +1, I think that would make the API much cleaner.

ramonpetgrave64: I agree, but we should wait to do a more thorough redesign before breaking the API for this feature.

loosebazooka: Ah gotcha, okay so if we're being strict about API breakages, I think just using variadic with VerifierOpts kind of leads into the v3 api better, even if we *must* to check for for 0 or 1 at runtime right now (and we feel it's gross)

Using `VerifierOptioner` temporarily and then moving to a new API that accepts VerifierOpts or Empty would be a more significant change?

ramonpetgrave64: I could do that. @slugclub wdyt?

slugclub: Yep I agree with @loosebazooka that if we need to be strict about API breakages, the best option would be to accept a variadic `VerifierOpts` and check for 0 or 1 at runtime.

ramonpetgrave64: @slugclub Thanks! I've added it and renamed it to `ClientOpts`, because now we have a separate `VerificationOpts`. Although, I could fold it into `VerificationOpts`, instead. That might make more sense. wdyt @loosebazooka ?

 - https://github.com/slsa-framework/slsa-verifier/pull/768/files#diff-76f121a3a81c8945dc0e0c81a4b95e941d252522eba8704a1aa6f6a1722a8b68R62

loosebazooka: Yeah sounds good. Thanks everyone for the fruitful discussion here too.

ramonpetgrave64: Just making sure: you'd prefer it folded into `ClientOpts` folded into `VerificationOpts`?

haydentherapper: I don't feel super strongly about this, but I think these configurations as separate structs seems reasonable, as we can grow `ClientOpts` to include configuration for other service dependencies of slsa-verifier while `VerificationOpts` is focused on cryptographic verification options.

Issue #161, opened 2022-07-22
Proper logging


laurentsimon:
It would be nice to use a logging library or a more standard way to log errors, especially with a good formatting

PR #772, opened 2024-05-17
fix: #161: replace prints with logs

ramonpetgrave64:
Addresses #161, and followup to #768

This PR replaces all `fmt` print statements with `log/slog` logs at appropriate levels.

for successes
```
2024/05/17 17:43:33 INFO Verified build using builder https://github.com/slsa-framework/slsa-github-generator/.github/workflows/delegator_lowperms-generic_slsa3.yml@refs/tags/v1.6.0 at commit b12088652ee70b25fcd64efd002aa851c6e247fe
2024/05/17 17:43:33 INFO Verifying npm package ../stuff/actions-test.tgz: PASSED

2024/05/17 17:43:33 INFO PASSED: Verified SLSA provenance
```

and for failures

```
2024/05/17 17:44:54 ERROR Verifying npm package ../stuff/actions-test.tgz: FAILED: builderID does not match provenance: expected name 'https://github.com/slsa-framework/slsa-github-generator/.github/workflows/builder_nodejs_slsa3.yml', got 'https://github.com/slsa-framework/slsa-github-generator/.github/workflows/builer_nodejs_slsa3.yml'

2024/05/17 17:44:54 ERROR FAILED: SLSA verification failed: builderID does not match provenance: expected name 'https://github.com/slsa-framework/slsa-github-generator/.github/workflows/builder_nodejs_slsa3.yml', got 'https://github.com/slsa-framework/slsa-github-generator/.github/workflows/builer_nodejs_slsa3.yml'
exit status 1
```

If you're using slssa-verifier as a library, you can suppress INFO logs by invoking

```
slog.SetLogLoggerLevel(slog.LevelError)
```


cli/slsa-verifier/verify/verify_npm_package.go
```diff
@@ -74,22 +75,22 @@ func (c *VerifyNpmPackageCommand) Exec(ctx context.Context, tarballs []string) (
...
 			return nil, err
 		}
 
 		if c.PrintProvenance {
-			fmt.Fprintf(os.Stdout, "%s\n", string(verifiedProvenance))
+			slog.Info(fmt.Sprintf("%s\n", string(verifiedProvenance)))
 		}
 
 		builderID = outBuilderID
-		fmt.Fprintf(os.Stderr, "Verifying npm package %s: PASSED\n\n", tarball)
```
laurentsimon, line 92: where does slog write the result? We need to ensure compatibility and ensure it's written to stderr when `fmt.Fprintf(os.Stderr`. Otherwise folks won't be able to pipe the result into jq or other tools: jq will pick not just the printed provenance, but the other messages too.

I think before changing this API we should define what the interface is for our logger. And we should provide an interface as logger, so that callers can customize how they log. Here's a fast logging library https://github.com/uber-go/zap, among others, that would need to be instantiable by callers

ramonpetgrave64: I confirmed that by default slog will send to stderr. And then I removed the changes against `fmt.Fprintf(os.Stdout, ...` statements.

I'll look into how to best allow caller-provided loggers.

ianlewis: FWIW, I'm personally in favor of using the stdlib logging packages rather than including a dependency.

PR #777, opened 2024-06-20
feat: vsa support

### LINKED ISSUE #542: feat: Verification of VSA
See VSA https://slsa.dev/verification_summary/v0.2

High-level verification in CLI:

```shell
$ slsa-verifier verify-vsa <artifact> --vsa-path <vsa-file> verifier-id google.com [--resource-uri <purl>] --policy-level X
```

ramonpetgrave64:
Fixes #542

Adds support for VSAs.

## Testing process

- added some unit an end-to-end tests
- manually invoking

    ```
    go run ./cli/slsa-verifier/ verify-vsa \
    --subject-digest gce_image_id:8970095005306000053 \
    --attestation-path ./cli/slsa-verifier/testdata/vsa/gce/v1/gke-gce-pre.bcid-vsa.jsonl \
    --verifier-id https://bcid.corp.google.com/verifier/bcid_package_enforcer/v0.1 \
    --resource-uri gce_image://gke-node-images:gke-12615-gke1418000-cos-101-17162-463-29-c-cgpv1-pre \
    --verified-level BCID_L1 \
    --verified-level SLSA_BUILD_LEVEL_2 \
    --public-key-path ./cli/slsa-verifier/testdata/vsa/gce/v1/vsa_signing_public_key.pem \
    --public-key-id keystore://76574:prod:vsa_signing_public_key \
    --print-attestation


    {"_type":"https://in-toto.io/Statement/v1","predicateType":"https://slsa.dev/verification_summary/v1","predicate":{"timeVerified":"2024-06-12T07:24:34.351608Z","verifier":{"id":"https://bcid.corp.google.com/verifier/bcid_package_enforcer/v0.1"},"verificationResult":"PASSED","verifiedLevels":["BCID_L1","SLSA_BUILD_LEVEL_2"],"resourceUri":"gce_image://gke-node-images:gke-12615-gke1418000-cos-101-17162-463-29-c-cgpv1-pre","policy":{"uri":"googlefile:/google_src/files/642513192/depot/google3/production/security/bcid/software/gce_image/gke/vm_images.sw_policy.textproto"}},"subject":[{"name":"_","digest":{"gce_image_id":"8970095005306000053"}}]}
    Verifying VSA: PASSED
    
    PASSED: SLSA verification passed
    ```

TODOS:
 - open issue on the in_toto attestations repo about the incorrect json [fields](https://github.com/in-toto/attestation/blob/36c11295429a997d5bb520b4e80a1d0c16845f9c/go/predicates/vsa/v1/vsa.pb.go#L26-L40) for vsa 1.0


ramonpetgrave64:
@haydentherapper @laurentsimon

verifiers/internal/vsa/verifier.go
```diff
@@ -0,0 +1,201 @@
...
+	sigstoreDSSE "github.com/sigstore/sigstore/pkg/signature/dsse"
+	serrors "github.com/slsa-framework/slsa-verifier/v2/errors"
+	"github.com/slsa-framework/slsa-verifier/v2/options"
+	vsa10 "github.com/slsa-framework/slsa-verifier/v2/verifiers/internal/vsa/v1.0"
+	"github.com/slsa-framework/slsa-verifier/v2/verifiers/utils"
+)
+
+// VerifyVSA verifies the VSA attestations.
+func VerifyVSA(ctx context.Context,
+	attestations []byte,
```
haydentherapper, line 19: Should this be singular?

verifiers/internal/vsa/verifier.go
```diff
@@ -0,0 +1,201 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// 2. match the subject digests,
+	// 4. match the verifier ID,
+	// 5. match the expected valuesmatch resourceURI,
+	// 6. confirm the slsaResult is PASSED,
+	// 7. match the verifiedLevels,
+	// no other feields are checked.
```
haydentherapper, line 52:
```suggestion
	// no other fields are checked.
```

cli/slsa-verifier/verify/options.go
```diff
@@ -127,6 +127,59 @@ func (o *VerifyNpmOptions) AddFlags(cmd *cobra.Command) {
...
+	PublicKeyHashAlgo string
+	PrintAttestation  bool
+}
+
+var _ Interface = (*VerifyVSAOptions)(nil)
+
+// AddFlags implements Interface.
+func (o *VerifyVSAOptions) AddFlags(cmd *cobra.Command) {
+	cmd.Flags().StringArrayVar(&o.SubjectDigests, "subject-digest", []string{},
+		"the digests to be verified. Pass multiple digests by repeating the flag.")
```
haydentherapper, line 148: Do we need to document that format for each digest, that it's `alg:digest`?

verifiers/utils/dsse.go
```diff
@@ -42,13 +43,17 @@ func PayloadFromEnvelope(env *dsselib.Envelope) ([]byte, error) {
...
+// but it parses json fields in snake case, while the official spec uses camel case
+// https://github.com/in-toto/attestation/blob/v1.0/spec/v1.0/statement.md.
 func StatementFromBytes(payload []byte) (*intoto.Statement, error) {
 	var statement intoto.Statement
 	if err := json.Unmarshal(payload, &statement); err != nil {
 		return nil, fmt.Errorf("%w: %w", serrors.ErrorInvalidDssePayload, err)
 	}
 
-	if statement.Type != intoto.StatementInTotoV01 {
+	if statement.Type != intoto.StatementInTotoV01 && statement.Type != intotoAttestattions.StatementTypeUri {
```
haydentherapper, line 56: Can we use `intot.StatementInTotoV1` to avoid pulling in another package? https://github.com/in-toto/in-toto-golang/blob/72262c0fd3a9c392440040c04336174eddac427b/in_toto/attestations.go#L18C2-L18C19

cli/slsa-verifier/verify/verify_vsa.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	verifiedProvenance, outProducerID, err := verifiers.VerifyVSA(ctx, attestations, vsaOpts, VerificationOpts)
+	if err != nil {
+		printFailed(err)
+		return nil, err
+	}
+	if c.PrintAttestation {
+		fmt.Fprintf(os.Stdout, "%s\n", string(verifiedProvenance))
+	}
+	fmt.Fprintf(os.Stderr, "Verifying VSA: PASSED\n\n")
```
loosebazooka, line 99: Similarly, I think limiting the places where we "print" messages might be nice. Could `verifyVSACmd` just handle all the printing, and these inner packages/files just return state? (except for PrintAttestation)

verifiers/utils/dsse.go
ramonpetgrave64: That's not yet released in the latest v0.9.0, but that would be ideal
 - https://github.com/in-toto/in-toto-golang/blob/v0.9.0/in_toto/attestations.go#L10-L22

haydentherapper: Ah that makes sense. We could also hardcode the string, but this LGTM as well.

verifiers/internal/vsa/verifier.go
```diff
@@ -0,0 +1,201 @@
...
+	verificationOpts *options.VerificationOpts,
+) ([]byte, *utils.TrustedAttesterID, error) {
+	// following steps in https://slsa.dev/spec/v1.1/verification_summary#how-to-verify
+
+	envelope, err := utils.EnvelopeFromBytes(attestations)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// 1. verify the envelope signature,
```
loosebazooka, line 30: nit: These step numbers need an update potentially? (out of order)

verifiers/verifier.go
```diff
@@ -74,3 +75,12 @@ func VerifyNpmPackage(ctx context.Context,
 	return verifier.VerifyNpmPackage(ctx, attestations, tarballHash,
 		provenanceOpts, builderOpts)
 }
+
+// VerifyVSA verifies the VSA attestations.
+func VerifyVSA(ctx context.Context,
+	attestations []byte,
+	vsaOpts *options.VSAOpts,
+	verificationOpts *options.VerificationOpts,
+) ([]byte, *utils.TrustedAttesterID, error) {
```
loosebazooka, line 84: I can't seem to tell where these returned values are consumed? `[]byte` and the `trustedAttesterID`? Can we update the comment?

loosebazooka (commented):
First pass looks good. I'm still trying to wrap my head around some of it.

README.md
```diff
@@ -481,6 +481,40 @@ The verified in-toto statement may be written to stdout with the
...
 
+### Verification Summary Attestations (VSA)
+
+We have experimental support for [verifying](https://slsa.dev/spec/v1.1/verification_summary#how-to-verify) VSAs.
+Rather than passing in filepaths as arguments, we allow passing in mulitple `--subject-digest` cli options, to
+accomodate subjects that are not simple-files.
+
+This experimental support does not work yet with VSAs wrapped in Sigstore bundles, only with simple DSSE envelopes.
```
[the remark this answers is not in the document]
loosebazooka: Not early, probably much later. And we're aware of it. We can "unwrap" in `bazel-audit` before we pass it in.

verifiers/internal/vsa/verifier.go
ramonpetgrave64: That's the right order. Step 4 is done by two functions.

verifiers/verifier.go
ramonpetgrave64: the `[]byte` is consumed when doing `PrintAttestation`, and for anyone using the `VerifyVSA()` as a library function.

cli/slsa-verifier/verify/verify_vsa.go
```diff
@@ -0,0 +1,107 @@
...
+	"SHA256": crypto.SHA256,
+	"SHA384": crypto.SHA384,
+	"SHA512": crypto.SHA512,
+}
+
+// Exec executes the verifiers.VerifyVSA.
+func (c *VerifyVSACommand) Exec(ctx context.Context) (*utils.TrustedAttesterID, error) {
+	if !options.ExperimentalEnabled() {
+		err := errors.New("feature support is only provided in SLSA_VERIFIER_EXPERIMENTAL mode")
+		printFailed(err)
```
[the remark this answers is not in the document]
ramonpetgrave64: I was just following the norms, but you make a good point.

README.md
```diff
@@ -481,6 +481,40 @@ The verified in-toto statement may be written to stdout with the
...
+With that, we allow the user to pass in the public key.
+Note that if the DSSE Envelope `signatures` specifies a `keyid` that is not a simple hash of the key, then you
```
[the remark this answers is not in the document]
ramonpetgrave64: I just meant like not using a well-known identifier. I reworded it a bit.

README.md
```diff
@@ -481,6 +481,40 @@ The verified in-toto statement may be written to stdout with the
...
+--verifier-id https://bcid.corp.google.com/verifier/bcid_package_enforcer/v0.1 \
+--resource-uri gce_image://gke-node-images:gke-12615-gke1418000-cos-101-17162-463-29-c-cgpv1-pre \
+--verified-levels "BCID_L1, SLSA_BUILD_LEVEL_2" \
+--public-key-path ./cli/slsa-verifier/testdata/vsa/gce/v1/vsa_signing_public_key.pem \
+--public-key-id keystore://76574:prod:vsa_signing_public_key \
+--public-key-hash-algo SHA256 \
+--print-attestation
+```
+
+For multiple subhects, use:
```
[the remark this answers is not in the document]
ramonpetgrave64: fixed

options/options.go
```diff
@@ -37,3 +39,29 @@ type BuilderOpts struct {
 	// ExpectedBuilderID is the builderID passed in from the user to be verified
 	ExpectedID *string
 }
+
+// VSAOpts are the options for checking the VSA.
+type VSAOpts struct {
+	// ExpectedDigests are the digests expected to be in the VSA.
+	ExpectedDigests *[]string
+
+	// ExpectedVerifierID is the verifier ID that is passed from user and not verified.
```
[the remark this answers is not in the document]
ramonpetgrave64: I meant that is hasn't yet been matched in the provenance. Reworded.

cli/slsa-verifier/verify/verify_vsa.go
```diff
@@ -0,0 +1,107 @@
...
+		return nil, err
+	}
+	pubKey, err := cryptoutils.UnmarshalPEMToPublicKey(pubKeyBytes)
+	if err != nil {
+		err = fmt.Errorf("%w: %w", serrors.ErrorInvalidPublicKey, err)
+		printFailed(err)
+		return nil, err
+	}
+	hashHalgo, ok := hashAlgos[*c.PublicKeyHashAlgo]
```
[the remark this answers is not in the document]
ramonpetgrave64: fixed

verifiers/internal/vsa/verifier.go
ramonpetgrave64: fixed

README.md
```diff
@@ -481,6 +481,40 @@ The verified in-toto statement may be written to stdout with the
...
+SLSA_VERIFIER_EXPERIMENTAL=1 \
+go run ./cli/slsa-verifier/ verify-vsa \
+--subject-digest gce_image_id:8970095005306000053 \
+--attestations-path ./cli/slsa-verifier/testdata/vsa/gce/v1/gke-gce-pre.bcid-vsa.jsonl \
+--verifier-id https://bcid.corp.google.com/verifier/bcid_package_enforcer/v0.1 \
+--resource-uri gce_image://gke-node-images:gke-12615-gke1418000-cos-101-17162-463-29-c-cgpv1-pre \
+--verified-levels "BCID_L1, SLSA_BUILD_LEVEL_2" \
+--public-key-path ./cli/slsa-verifier/testdata/vsa/gce/v1/vsa_signing_public_key.pem \
+--public-key-id keystore://76574:prod:vsa_signing_public_key \
+--public-key-hash-algo SHA256 \
```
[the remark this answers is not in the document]
ramonpetgrave64: changed

README.md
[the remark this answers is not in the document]
ramonpetgrave64: thanks, i added that definition

cli/slsa-verifier/verify/options.go
```diff
@@ -127,6 +127,59 @@ func (o *VerifyNpmOptions) AddFlags(cmd *cobra.Command) {
...
+		"path to a file containing the attestations")
+
+	cmd.Flags().StringVar(&o.VerifierID, "verifier-id", "",
+		"the unique verifier ID who created the attestations")
+
+	cmd.Flags().StringVar(&o.ResourceURI, "resource-uri", "",
+		"the resource URI to be verified")
+
+	cmd.Flags().StringSliceVar(&o.VerifiedLevels, "verified-levels", []string{},
+		"the levels of verification to be performed")
```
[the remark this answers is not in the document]
ramonpetgrave64: added

cli/slsa-verifier/verify/options.go
```diff
@@ -127,6 +127,59 @@ func (o *VerifyNpmOptions) AddFlags(cmd *cobra.Command) {
...
+		"[optional] print the verified attestations to stdout")
+
+	cmd.Flags().StringVar(&o.PublicKeyPath, "public-key-path", "",
+		"path to a public key file")
+
+	cmd.Flags().StringVar(&o.PublicKeyID, "public-key-id", "",
+		"the ID of the public key")
+
+	cmd.Flags().StringVar(&o.PublicKeyHashAlgo, "public-key-hash-algo", "SHA256",
+		"the hash algorithm used to hash the public key, one of SHA256, SHA384, or SHA512")
```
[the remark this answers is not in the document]
ramonpetgrave64: added

verifiers/internal/vsa/verifier.go
ramonpetgrave64: changed

cli/slsa-verifier/verify/options.go
ramonpetgrave64: added

cli/slsa-verifier/verify/options.go
```diff
@@ -127,6 +127,59 @@ func (o *VerifyNpmOptions) AddFlags(cmd *cobra.Command) {
...
+		"the levels of verification to be performed")
+
+	cmd.Flags().BoolVar(&o.PrintAttestation, "print-attestation", false,
+		"[optional] print the verified attestations to stdout")
+
+	cmd.Flags().StringVar(&o.PublicKeyPath, "public-key-path", "",
+		"path to a public key file")
+
+	cmd.Flags().StringVar(&o.PublicKeyID, "public-key-id", "",
+		"the ID of the public key")
```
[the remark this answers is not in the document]
ramonpetgrave64: added

cli/slsa-verifier/verify/verify_vsa.go
```diff
@@ -0,0 +1,107 @@
...
+var hashAlgos = map[string]crypto.Hash{
+	"":       crypto.SHA256, // default to SHA256
+	"SHA256": crypto.SHA256,
+	"SHA384": crypto.SHA384,
+	"SHA512": crypto.SHA512,
+}
+
+// Exec executes the verifiers.VerifyVSA.
+func (c *VerifyVSACommand) Exec(ctx context.Context) (*utils.TrustedAttesterID, error) {
+	if !options.ExperimentalEnabled() {
```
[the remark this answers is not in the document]
ramonpetgrave64: I think we can remove it when we're ready to merge this PR.

cli/slsa-verifier/main_regression_test.go
```diff
@@ -1795,6 +1795,99 @@ func Test_runVerifyNpmPackage(t *testing.T) {
...
 }
 
+func Test_runVerifyVSA(t *testing.T) {
+	// We cannot use t.Setenv due to parallelized tests.
+	os.Setenv("SLSA_VERIFIER_EXPERIMENTAL", "1")
+	t.Parallel()
+
+	tests := []struct {
+		name              string
+		AttestationPath   *string
```
haydentherapper, line 1805: Can this be `attestationPath` as it doesn't need to be exported?

cli/slsa-verifier/main_regression_test.go
```diff
@@ -1795,6 +1795,99 @@ func Test_runVerifyNpmPackage(t *testing.T) {
...
+		},
+		{
+			name:              "fail: gke, wrong key id",
+			AttestationPath:   PointerTo("gce/v1/gke-gce-pre.bcid-vsa.jsonl"),
+			publicKeyPath:     PointerTo("gce/v1/vsa_signing_public_key.pem"),
+			publicKeyID:       PointerTo("my_key_id"),
+			publicKeyHashAlgo: PointerTo("SHA256"),
+			err:               serrors.ErrorNoValidSignature,
+		},
+		// TODO: Add more tests for different scenarios.
```
haydentherapper, line 1860: Do you have other tests you want to add? Otherwise I'd remove.

cli/slsa-verifier/verify/options.go
```diff
@@ -127,6 +127,58 @@ func (o *VerifyNpmOptions) AddFlags(cmd *cobra.Command) {
...
+		"[optional] print the verified attestations to stdout")
+
+	cmd.Flags().StringVar(&o.PublicKeyPath, "public-key-path", "",
+		"path to a public key file")
+
+	cmd.Flags().StringVar(&o.PublicKeyID, "public-key-id", "",
+		"[optional] the ID of the public key")
+
+	cmd.Flags().StringVar(&o.PublicKeyHashAlgo, "public-key-hash-algo", "SHA256",
+		"[optional] the hash algorithm used to hash the public key, one of SHA256 [efault], SHA384, or SHA512")
```
haydentherapper, line 172:
```suggestion
		"[optional] the hash algorithm used to compute the digest to be signed, one of SHA256 [default], SHA384, or SHA512")
```

cli/slsa-verifier/verify/options.go
```diff
@@ -127,6 +127,58 @@ func (o *VerifyNpmOptions) AddFlags(cmd *cobra.Command) {
...
+		"[optional] the levels of verification to be performed, comma-separated. e.g., 'SLSA_BUILD_LEVEL_2,FEDRAMP_LOW'")
+
+	cmd.Flags().BoolVar(&o.PrintAttestation, "print-attestation", false,
+		"[optional] print the verified attestations to stdout")
+
+	cmd.Flags().StringVar(&o.PublicKeyPath, "public-key-path", "",
+		"path to a public key file")
+
+	cmd.Flags().StringVar(&o.PublicKeyID, "public-key-id", "",
+		"[optional] the ID of the public key")
```
haydentherapper, line 169:
```suggestion
		"[optional] the ID of the public key, defaults to the SHA256 digest of the base64 encoded")
```

haydentherapper: Missed from the previous comment

cli/slsa-verifier/verify/verify_vsa.go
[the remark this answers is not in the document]
haydentherapper: SG, once this leaves draft, can you remove it?

verifiers/internal/vsa/verifier.go
haydentherapper, line 19:
```suggestion
	attestation []byte,
```

verifiers/internal/vsa/verifier.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	// 3. parse the VSA, verifying the predicateType.
+	vsa, err := vsa10.VSAFromStatement(statement)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// 2. match the subject digests,
+	// 4. match the verifier ID,
+	// 5. match the expected valuesmatch resourceURI,
```
haydentherapper, line 49:
```suggestion
	// 5. match the expected values, match resourceURI,
```

verifiers/utils/builder.go
```diff
@@ -9,6 +9,23 @@ import (
 	serrors "github.com/slsa-framework/slsa-verifier/v2/errors"
 )
 
+// TrustedAttesterID represents an identifier that has been explicitly trusted.
+type TrustedAttesterID struct {
+	TrustedBuilderID
+}
+
+// TrustedAttesterIDNew creates a new AttesterID structure.
+func TrustedAttesterIDNew(attesterID string, needVersion bool) (*TrustedAttesterID, error) {
```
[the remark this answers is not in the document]
haydentherapper: +1, `New` should proceed the struct being created

verifiers/verifier.go
haydentherapper: +1, comment should document what the returned values are.

verifiers/internal/vsa/vsa_test.go
```diff
@@ -0,0 +1,187 @@
...
+		panic(err)
+	}
+	pubKey, err := cryptoutils.UnmarshalPEMToPublicKey(pubKeyBytes)
+	if err != nil {
+		panic(err)
+	}
+	return pubKey
+}
+
+func PointerTo[K any](object K) *K {
```
haydentherapper, line 185:
```suggestion
func pointerTo[K any](object K) *K {
```

verifiers/utils/builder.go
```diff
@@ -9,6 +9,25 @@ import (
 	serrors "github.com/slsa-framework/slsa-verifier/v2/errors"
 )
 
+// TrustedAttesterID represents an identifier that has been explicitly trusted.
+// TODO: don't embed TrustedBuilderID, use a whole new type that implements the
```
haydentherapper, line 13: As someone not familiar with the codebase, why is this?

verifiers/internal/vsa/vsa_test.go
```diff
@@ -0,0 +1,187 @@
...
+const testDir = "./testdata"
+
+func Test_VerifyVSA(t *testing.T) {
+	t.Parallel()
+
+	ctx := context.Background()
+
+	tests := []struct {
+		name             string
+		AttestationPath  string
```
haydentherapper, line 27:
```suggestion
		attestationPath  string
```

haydentherapper: Same here, can this be lower case?

haydentherapper (commented):
Just a few stylistic and spelling nits, otherwise LGTM

kaixu-google:
Could you consider adding well-known public key pems for VSAs to the SLSA verifier, at least in the short term? There are a large number of clients needs to verify Google-issued VSAs so it'll be easier to update the public key in case Google does a key rotation. Longer term, we could consider how to integrate Google's public key pem into sigstore.

cli/slsa-verifier/verify/options.go
ramonpetgrave64: added

verifiers/internal/vsa/verifier.go
ramonpetgrave64: fixed

verifiers/internal/vsa/vsa_test.go
ramonpetgrave64: changed

verifiers/verifier.go
ramonpetgrave64: added

cli/slsa-verifier/main_regression_test.go
ramonpetgrave64: changed

cli/slsa-verifier/verify/options.go
ramonpetgrave64: changed

verifiers/internal/vsa/vsa_test.go
ramonpetgrave64: changed

haydentherapper:
@kaixu-google The public key doesn't necessarily have to be distributed through Sigstore, it could also be distributed through GCS, through a well known endpoint, etc. We don't want to block development of this feature on figuring out how best to support key rotation. If we were to include the keys and later remove them, this would be a breaking feature. Additionally it would mean that slsa-verifier releases are tied to the key - if the key is revoked, then we must cut a new slsa-verifier release and revoke all previous releases as well (which we have no mechanism to do so, and that would be very problematic for all users).

AdamZWu:
> Could you consider adding well-known public key pems for VSAs to the SLSA verifier, at least in the short term? There are a large number of clients needs to verify Google-issued VSAs so it'll be easier to update the public key in case Google does a key rotation. Longer term, we could consider how to integrate Google's public key pem into sigstore.

Agree with Kai that the slew of key related parameters may feel a bit overwhelming to the end users:

```
slsa-verifier verify-vsa \
...
--public-key-path ./cli/slsa-verifier/testdata/vsa/gce/v1/vsa_signing_public_key.pem \
--public-key-id keystore://76574:prod:vsa_signing_public_key \
--public-key-hash-algo SHA256 \
```

Ideally, there could be a "registry" that allows trusted organizations to register info that are needed to verify their signatures. And then the end user only needs to declare their trusted organizations. Something like:

```
slsa-verifier verify-vsa \
...
--trusted_issuer "Google, LLC."
```

But that is probably more fitted as a separate UX improvement FR. :)

haydentherapper:
`public-key-hash-algo` is optional and defaults to sha256 which is almost always the case (except for ecdsa-p521 and ed25519 keys) - we could also rename that to `signing-hash-algo` to signify that it's related to the signature, not the public key. WDYT @ramonpetgrave64?

Long-term, we were talking about a mechanism to retrieve keys from well-known endpoints such that rotation is abstracted away. Then `public-key-path` would be optional as well. A registry for a large entity would be tricky since it's likely there will be multiple keys within an organization, so I suspect `public-key-id` will always be necessary.

cli/slsa-verifier/main_regression_test.go
ramonpetgrave64: Added some more tests

verifiers/internal/vsa/testdata/gce/v1/gke-gce-pre.bcid-vsa.jsonl
```diff
@@ -0,0 +1,10 @@
+{
```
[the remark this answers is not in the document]
ramonpetgrave64: I changed the tests to use embedded data, instead

verifiers/utils/builder.go
ramonpetgrave64: I decided to remove it. It seems like it was originally for semver-like capabilities. Instead, I'll just refer the user in comments to a another semver package.
 - https://github.com/slsa-framework/slsa-verifier/pull/256
 - https://pkg.go.dev/golang.org/x/mod/semver

ramonpetgrave64:
@kaixu-google Since Google VSAs are not yet publicly available, I hope that the plan is that distribution of Google's VSAs will be equally convenient as distributing their Public Keys. If that will be the case, I think we should stick with not embedding the key, given the cons that @haydentherapper spoke of. 

@AdamZWu Ideally, Google's public key could be distributed and rotated within a Sigstore TUF Rooot, similar to how Npmjs distributes their signing keys. 
 - ref: https://github.com/slsa-framework/slsa-verifier/pull/731/files#diff-e838304ec724aef1ba7a4db69d32ff0e02fcf541a96136f9790253c628b589f0R55-R57

ramonpetgrave64:
Thanks everyone for the first passes! I think we're ready for review.

verifiers/utils/builder.go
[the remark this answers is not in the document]
ramonpetgrave64: removed it

cli/slsa-verifier/verify/verify_vsa.go
ramonpetgrave64: I see what you mean. In this case, I'll stick with tradition.

kaixu-google:
Thank you Hayden and Ramon! LGTM. I will try to file a FR to consider how
to improve the UX.

--
Best,
Kai Xu

ramonpetgrave64:
@kaixu-google

kaixu-google:
LGTM.

--
Best,
Kai Xu

README.md
```diff
@@ -481,6 +485,69 @@ The verified in-toto statement may be written to stdout with the
...
+--verifier-id https://bcid.corp.google.com/verifier/bcid_package_enforcer/v0.1 \
+--resource-uri gce_image://gke-node-images:gke-12615-gke1418000-cos-101-17162-463-29-c-cgpv1-pre \
+--verified-levels "BCID_L1, SLSA_BUILD_LEVEL_2" \
+--public-key-path ./cli/slsa-verifier/testdata/vsa/gce/v1/vsa_signing_public_key.pem \
+--public-key-id keystore://76574:prod:vsa_signing_public_key \
+--public-key-signing-hash-algo SHA256 \
+--print-attestation
+```
+
+For multiple subhects, use:
```
haydentherapper, line 532:
```suggestion
For multiple subjects, use:
```

cli/slsa-verifier/main_regression_test.go
```diff
@@ -1795,6 +1795,97 @@ func Test_runVerifyNpmPackage(t *testing.T) {
...
 }
 
+// Test_runVerifyVSA tests the CLI inputes of verify-vsa. More extensive tests are in
+// slsa-verifier/verifiers/internal/vsa/verifier_test.go
+func Test_runVerifyVSA(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name              string
+		AttestationPath   *string
```
haydentherapper, line 1805: Looks like `attestationPath` was missed

cli/slsa-verifier/main_regression_test.go
```diff
@@ -1795,6 +1795,97 @@ func Test_runVerifyNpmPackage(t *testing.T) {
...
+			}
+
+			err := cmd.Exec(context.Background())
+			if diff := cmp.Diff(tt.err, err, cmpopts.EquateErrors()); diff != "" {
+				t.Fatalf("unexpected error (-want +got): \n%s", diff)
+			}
+		})
+	}
+}
+
```
haydentherapper, line 1888: Also `pointerTo`, don't need to export

haydentherapper (approved):
Looks great, great work on this! Just a few tiny style changes

cli/slsa-verifier/main_regression_test.go
ramonpetgrave64: fixed

README.md
ramonpetgrave64: fixed

cli/slsa-verifier/main_regression_test.go
ramonpetgrave64: fixed

haydentherapper (approved):
🥳

ramonpetgrave64:
@laurentsimon @ianlewis

verifiers/utils/dsse.go
```diff
@@ -42,13 +43,17 @@ func PayloadFromEnvelope(env *dsselib.Envelope) ([]byte, error) {
...
+// but it parses json fields in snake case, while the official spec uses camel case
+// https://github.com/in-toto/attestation/blob/v1.0/spec/v1.0/statement.md.
 func StatementFromBytes(payload []byte) (*intoto.Statement, error) {
 	var statement intoto.Statement
 	if err := json.Unmarshal(payload, &statement); err != nil {
 		return nil, fmt.Errorf("%w: %w", serrors.ErrorInvalidDssePayload, err)
 	}
 
-	if statement.Type != intoto.StatementInTotoV01 {
+	if statement.Type != intoto.StatementInTotoV01 && statement.Type != intotoAttestations.StatementTypeUri {
```
laurentsimon, line 56: do you intend to be a `&&` or an `||`?

README.md
```diff
@@ -481,6 +485,69 @@ The verified in-toto statement may be written to stdout with the
...
 Note that `--source-uri` supports GitHub repository URIs like `github.com/$OWNER/$REPO` when the build was enabled with a Cloud Build [GitHub trigger](https://cloud.google.com/build/docs/automating-builds/github/build-repos-from-github). Otherwise, the build provenance will contain the name of the Cloud Storage bucket used to host the source files, usually of the form `gs://[PROJECT_ID]_cloudbuild/source` (see [Running build](https://cloud.google.com/build/docs/running-builds/submit-build-via-cli-api#running_builds)). We recommend using GitHub triggers in order to preserve the source provenance and valiate that the source came from an expected, version-controlled repository. You _may_ match on the fully-qualified tar like `gs://[PROJECT_ID]_cloudbuild/source/1665165360.279777-955d1904741e4bbeb3461080299e929a.tgz`.
 
+## Verification Summary Attestations (VSA)
+
+We have support for [verifying](https://slsa.dev/spec/v1.1/verification_summary#how-to-verify) VSAs.
+Rather than passing in filepaths as arguments, we allow passing in mulitple `--subject-digest` cli options, to
+accomodate subjects that are not simple-files.
+
+
+The verify-vsa command
```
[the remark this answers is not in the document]
haydentherapper: Why do you think this is dangerous? If you get to the point where you have a VSA you need to verify, I assume you should understand what the VSA represents. So maybe adding a link to https://slsa.dev/spec/v1.1/verification_summary#purpose to make sure users understand one?

README.md
```diff
@@ -481,6 +485,69 @@ The verified in-toto statement may be written to stdout with the
...
+Verifies SLSA VSAs for the given subject-digests
+
+Usage:
+  slsa-verifier verify-vsa [flags] subject-digest [subject-digest...]
+
+Flags:
+      --attestations-path string      path to a file containing the attestations
+  -h, --help                          help for verify-vsa
+      --print-attestation             [optional] print the contents of attestation to stdout
+      --public-key-signing-hash-algo string   [optional] the hash algorithm used to compute the digest to be signed, one of SHA256 [default], SHA384, or SHA512
```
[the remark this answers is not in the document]
haydentherapper: The reason I had suggested this is because the key parsing libraries we use support all of Go's signature algorithms.
Thinking more about this now, maybe we can have the best of both worlds, supporting both all signature algorithms and also have a mapping between signature alg and the expected hash alg, without the need for the flag.

RSA -> Always SHA256 (RSA2048-SHA256, RSA3072-SHA256, RSA4096-SHA256 are all typical)
ECDSA-P256 -> SHA256
ECDSA-P384 -> SHA384
ECDSA-P512 -> SHA512
Ed25519 -> SHA512

This aligns with what is supported by Sigstore's spec - https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_common.proto#L69-L86

Should be really easy to add:

```
var h crypto.Hash
switch pk := pub.(type) {
case *rsa.PublicKey:
    h = crypto.SHA256
case *ecdsa.PublicKey:
    switch pk.Curve {
    case elliptic.P256():
	    h = crypto.SHA256
    case elliptic.P384():
	    h = crypto.SHA384
    case elliptic.P521():
	    h = crypto.SHA512
    default:
	    h = crypto.SHA256 (or err out)
    }
case ed25519.PublicKey:
    h = crypto.SHA512
}
```

Thoughts? We can then drop the flag, but support all signature algorithms.

verifiers/utils/dsse.go
ramonpetgrave64: Definitely a `&&`, so we can say "if it's neither a V01 or V10, then error."But we can change it to a more proper NOR or use a map like we've done in other parts of the code.

README.md
```diff
@@ -481,6 +485,69 @@ The verified in-toto statement may be written to stdout with the
...
+      --public-key-id string          [optional] the ID of the public key, defaults to the SHA256 digest of the base64-encoded public key
+      --public-key-path string        path to a public key file
+      --resource-uri string           the resource URI to be verified
+      --subject-digest stringArray    the digests to be verified. Pass multiple digests by repeating the flag. e.g. <digest type>:<digest value>
+      --verified-levels strings       [optional] the levels of verification to be performed, comma-separated. e.g., 'SLSA_BUILD_LEVEL_2,FEDRAMP_LOW'
```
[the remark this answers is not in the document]
ramonpetgrave64: It's not a SLSA level, but the specs leave open the possibility of additional "tracks" that can be included in verifiedLvels.

README.md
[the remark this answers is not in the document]
ramonpetgrave64: @haydentherapper From the CLI, I think we could use your proposal, but only if the user does not specify the hashing algorithm, and then error if it can't be automatically determined by the switch-case.

Just one question: Is it uncommon to use RSA with SHA512? Unless we start having better logging, It could be a bad UX for the user if verifying the signature is failing because the hashing algorithm was incorrectly assumed by our tool to be SHA256.

README.md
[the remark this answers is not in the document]
haydentherapper: I'm not super opinionated on whether to use comma-delimited strings vs multiple instances of the flag name, though from a stylistic point of view, being consistent might be better.

README.md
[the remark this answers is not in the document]
haydentherapper: My suggestion would be to drop the flag and only configure the hash alg based on the public key.

We could later add the flag back if we find that we need to support uncommon use cases like RSA-SHA512. At least within the PKI space, it's not common. We can document what key types we support and ask users to file an issue if they need additional support.

(BTW, I missed the case of `elliptic.P224()`, so the default case of sha256 seems fine for `*ecdsa.PublicKey`).

README.md
```diff
@@ -481,6 +485,69 @@ The verified in-toto statement may be written to stdout with the
...
+
+Flags:
+      --attestations-path string      path to a file containing the attestations
+  -h, --help                          help for verify-vsa
+      --print-attestation             [optional] print the contents of attestation to stdout
+      --public-key-signing-hash-algo string   [optional] the hash algorithm used to compute the digest to be signed, one of SHA256 [default], SHA384, or SHA512
+      --public-key-id string          [optional] the ID of the public key, defaults to the SHA256 digest of the base64-encoded public key
+      --public-key-path string        path to a public key file
+      --resource-uri string           the resource URI to be verified
+      --subject-digest stringArray    the digests to be verified. Pass multiple digests by repeating the flag. e.g. <digest type>:<digest value>
```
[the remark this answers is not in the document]
ramonpetgrave64: That's just the way cobra command shows the help menu, since it accepts multiple instances of `--subject-digest`.

README.md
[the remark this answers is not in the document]
ramonpetgrave64: @laurentsimon Instead of `verify-artifact --vsa-path`, I was thinking to continue with `verify-vsa path1 path2 ...`, or `verify-vsa --artifact-path path1 --artifact-path path2 ...`

README.md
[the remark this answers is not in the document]
ramonpetgrave64: Ok I removed the --public-key-hash-algo option and added your switch-case

README.md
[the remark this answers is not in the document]
ramonpetgrave64: changed

README.md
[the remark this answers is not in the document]
haydentherapper: LGTM, thanks!

README.md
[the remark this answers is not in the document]
laurentsimon: i don't think re-using `verify-vsa` is consistent with the existing API. We currently use `verify-artifact` and `verify-image`, so I think it's the right way to build on existing APIs. `verify-vsa` is a low-level API for the rare use case that the artifact is not available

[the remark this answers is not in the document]
laurentsimon: re: why it's dangerous API. TOCTOU is an example of a vuln https://github.com/slsa-framework/slsa-verifier?tab=readme-ov-file#containers-1. We've been very careful to design APIs that prevent users from shooting themselves ini the foot. Leaving it up to users to calculate a sha is both poor UX and prone to mistakes.

README.md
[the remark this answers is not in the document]
laurentsimon: those additional tracks do _not_ exist today. I don't see value in committing to something that's not landed yet. If the specs said "arbitrary values can be used, not just verified levels", then I'd make sense. In particular FEDRAMP is _not_ a valid track level, so should ,not be accepted since it does not conform to the current specs.
What the use case for this atm?

README.md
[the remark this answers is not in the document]
laurentsimon: + on the discussion. Give safe defaults and only provide an overwrite if someone complains.

verifiers/utils/dsse.go
laurentsimon: my bad, mis-read the code. Thanks for the exaplanation

README.md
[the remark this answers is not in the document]
laurentsimon: ah I see.

README.md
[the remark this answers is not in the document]
ramonpetgrave64: > Users MAY use custom values here but MUST NOT use custom values starting with SLSA_.

It's purely an example. I know FEDRAMP isn't a public track. But we still have to allow custom, non-public tracks. I just chose FEDRAMP because it seems like a [plausible](https://www.fedramp.gov/baselines/) track that an org might have in private.

I could change this example to say `MY_CUSTOM_TRACK_3`. wdyt?

README.md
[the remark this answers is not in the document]
haydentherapper: OK, I see what you're saying. This seems more like a not ideal UX rather than dangerous, since we are asking for a digest, and the digest _should_ be an immutable reference - if it's not, then it's technically non-compliant with SLSA. If we were to add a comment, I might say that we'd like to refactor this at some point to move VSA verification under `verify-artifact`.

README.md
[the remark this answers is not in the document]
haydentherapper: I'd just drop FEDRAMP and stick with the defined SlsaResults, and maybe add a later comment that it can be a custom result as well.

README.md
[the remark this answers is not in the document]
ramonpetgrave64: I think of the CLI options more like 

`verify-artifact` actually means "verify-provenance-with-artifact", or "verify-artifact-provenance"

With this thinking, a VSA is not actually a provenance (build provenance). And I'm open to making a new command "verify-artifact-vsa". To avoid the danger of the user not realizing that they're not enjoying the benefits of an actual build provenance.

re: TOCU, you're referring to how `verify-image` relies on the user using external tools to calculate digests (docs incorrectly says "tarball")
 - https://github.com/slsa-framework/slsa-verifier?tab=readme-ov-file#the-verify-image-command

README.md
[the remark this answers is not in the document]
laurentsimon: > Users MAY use custom values here but MUST NOT use custom values starting with SLSA_.

Good point. Somehow I always end up on v0.1 of the specs because they are the default page shown and this line was not there :/

sidenote: the levels have some ordering we need to understand. If the user passes BUILD_LEVEL2 and the VSA says it's BUILD_LEVEL3, then it should succeed, because 3 >= 2. With unknown arbitrary names, we won't be able do that (nothing we can do about it)

README.md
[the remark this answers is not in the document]
laurentsimon: > verify-artifact actually means "verify-provenance-with-artifact", or "verify-artifact-provenance"

That was not the original intention, and that's why the term 'provenance' is not in the command's name. It's intended to verify artifacts with whatever attestation. It so happens the only one currently supported is provenance, but it need not be. If you ever need to verify an artifact with several attestations, you'll be able to do `verify-artifact --provenance-path --source-path etc`... unless they are packed into a bundle and you'll likely use `--attestations-path` like in npm command.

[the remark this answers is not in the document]
laurentsimon: > OK, I see what you're saying. This seems more like a not ideal UX rather than dangerous, since we are asking for a digest, and the digest _should_ be an immutable reference - if it's not, then it's technically non-compliant with SLSA. If we were to add a comment, I might say that we'd like to refactor this at some point to move VSA verification under `verify-artifact`.

just have to be careful that users don't use the non-immutable image for pulling

verifiers/internal/vsa/verifier.go
```diff
@@ -0,0 +1,215 @@
...
+// confirmVerificationResult checks that the policy verification result is "PASSED".
+func confirmVerificationResult(vsa *vsa10.VSA) error {
+	if normalizeString(vsa.Predicate.VerificationResult) != "PASSED" {
+		return fmt.Errorf("%w: verification result is not Passed: %s", serrors.ErrorInvalidVerificationResult, vsa.Predicate.VerificationResult)
+	}
+	return nil
+}
+
+// matchVerifiedLevels checks if the verified levels in the VSA match the expected values.
+func matchVerifiedLevels(vsa *vsa10.VSA, vsaOpts *options.VSAOpts) error {
```
laurentsimon, line 199: does this fail if user passes level 2 and vsa says level 3? iiuc, it should succeed.

verifiers/internal/vsa/verifier.go
```diff
@@ -0,0 +1,215 @@
...
+	// 4. match the verifier ID
+	if err := matchVerifierID(vsa, vsaOpts); err != nil {
+		return err
+	}
+	// 5. match the expected resourceURI
+	if err := matchResourceURI(vsa, vsaOpts); err != nil {
+		return err
+	}
+	// 6. confirm the verificationResult is Passed
+	if err := confirmVerificationResult(vsa); err != nil {
```
laurentsimon, line 112: nit: be consustent with function names. Other functions are called `match*`, this one is called `confirm*`

verifiers/internal/vsa/verifier.go
```diff
@@ -0,0 +1,215 @@
...
+	if normalizeString(vsa.Predicate.VerificationResult) != "PASSED" {
+		return fmt.Errorf("%w: verification result is not Passed: %s", serrors.ErrorInvalidVerificationResult, vsa.Predicate.VerificationResult)
+	}
+	return nil
+}
+
+// matchVerifiedLevels checks if the verified levels in the VSA match the expected values.
+func matchVerifiedLevels(vsa *vsa10.VSA, vsaOpts *options.VSAOpts) error {
+	vsaLevels := make(map[string]bool)
+	for _, level := range vsa.Predicate.VerifiedLevels {
```
laurentsimon, line 201: nit: iirc, only one value per track ios allowed in the VSA. You may want to validate that's the case. (+ unit tests)

verifiers/internal/vsa/verifier.go
ramonpetgrave64: Only because, unlike the others, it doesn't compare the VSAs value with any user input.

verifiers/internal/vsa/verifier.go
ramonpetgrave64: The specs says

> Users MUST NOT include more than one level per SLSA track.

I was originally going to leave out this inference because it didn't seem necessary, similar to how it also says

> The result of evaluating an artifact (or set of artifacts) against SLSA. SHOULD be one of these values:
> SLSA_BUILD_LEVEL_UNEVALUATED
> SLSA_BUILD_LEVEL_0
> ...

I see now that MUST and SHOULD are different, according to [how the spec use those words](https://www.rfc-editor.org/rfc/rfc2119). 

I change my mind now, I'll add it in.

ramonpetgrave64: added it.

verifiers/internal/vsa/verifier.go
ramonpetgrave64: changed.

README.md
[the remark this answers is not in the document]
ramonpetgrave64: Talked offline, its okay to keep the FEDRAMP exampe.

verifiers/internal/vsa/verifier.go
laurentsimon: gotcha. just a nit. up to you :)

laurentsimon (approved):
Thanks!

haydentherapper (approved):
Great work!!

laurentsimon (approved):

PR #785, opened 2024-06-28
chore: fix vuln: override autolinker ^4.0.0

ramonpetgrave64:
fixes https://github.com/slsa-framework/slsa-verifier/security/code-scanning/11

markdown-toc's latest v1.2.0 is still vulnerable via a transitive dependency, but hasn't received updates in a long time.

This PR overrides one of the other transitive dependencies to a non-vulnerable version.

more info here https://github.com/jonschlinkert/markdown-toc/issues/156#issuecomment-2197630000

# Testing process

 - Manually invoked `make markdown-toc` and it did succeed, while also adding a missing header in the README. 
 - Made a few typos in the headers and markdown-toc did fix them.
 - Cloned markdown-toc, added the override, and its unit tests passed


package.json
```diff
@@ -6,5 +6,8 @@
   "devDependencies": {
     "markdown-toc": "1.2.0",
     "renovate": "37.374.1"
+  },
+  "overrides": {
+    "autolinker": ">= 4.0.0 ^4.0.0"
```
slugclub, line 11: I believe this can be just `^4.0.0` as the set of versions selected by `^4.0.0` is a subset of those selected by `>=4.0.0` (you can play around with this on https://semver.npmjs.com/).

slugclub (approved):

package.json
ramonpetgrave64: You're right. changed.

loosebazooka (approved):

loosebazooka:
sorry I was logged in as distrolessBot.

PR #791, opened 2024-07-18
feat: refactor: use sigstore-go for fetching TrustedRoot

ramonpetgrave64:
Uses the `sigstore-go` library for fetching the `TrustedRoot`, which contains the Sigstore infrastructure certificates needed to validate the leaf ephemeral certificates used to sign artifacts.

Refactors:

- replace `TrustedRootSingleton()` with `getDefaultCosignCheckOpts()`, since only `VerifyImage()` will now need that data.
-  replace `cosign.ValidateAndUnpackCert` with`sigstoreVerify.VerifyLeafCertificate()`
- use `sync.Once` for sigstore and rekor clients, and the `TrustedRoot`

## Testing

- existing tests continue to pass
  - [negative tests ](https://github.com/slsa-framework/slsa-verifier/blob/d96b9777090694fa5096ee1b9c710a46b5a66f5e/cli/slsa-verifier/main_regression_test.go#L450-L471) against rekor TLogs
- manual invocations of `verify-artifact`.


ramonpetgrave64:
@haydentherapper @laurentsimon

ramonpetgrave64:
@loosebazooka

go.mod
```diff
@@ -20,7 +20,7 @@ require (
 	github.com/gorilla/mux v1.8.1
 	github.com/in-toto/attestation v1.1.0
 	github.com/sigstore/cosign/v2 v2.2.4
-	github.com/sigstore/sigstore-go v0.2.0
+	github.com/sigstore/sigstore-go v0.4.0
```
haydentherapper, line 23: You can bump to 0.5.1 now, which does have a few breaking changes in API calls, but I don't think you're affected.

verifiers/utils/sigstore_tuf.go
```diff
@@ -0,0 +1,64 @@
...
+// GetSigstoreTrustedRoot returns the trusted root for the Sigstore TUF client.
+func GetSigstoreTrustedRoot() (*sigstoreRoot.TrustedRoot, error) {
+	var err error
+	trustedRootOnce.Do(func() {
+		client, err := GetDefaultSigstoreTUFClient()
+		if err != nil {
+			trustedRootOnce = new(sync.Once)
+			return
+		}
+		trustedRoot, err = sigstoreRoot.GetTrustedRoot(client)
```
haydentherapper, line 54: Should we handle slsa-verifier used in any long-running (> 1 day) processes? If so, then we should look at using [LiveTrustedRoot](https://github.com/sigstore/sigstore-go/blob/2b6fc6df45dc26c329f4da6c8157340e0f435a0d/pkg/root/trusted_root.go#L310) to refresh the root periodically.

haydentherapper (approved):
Nice work on this! Just one remark about long-running processes.

verifiers/internal/gha/rekor.go
```diff
@@ -339,40 +375,28 @@ func SearchValidSignedAttestation(ctx context.Context, artifactHash string, prov
...
-				SubjectRegExp: certSubjectRegexp,
-			},
-		},
-		CTLogPubKeys: trustedRoot.CTPubKeys,
-	}
-	verifier, err := cosign.ValidateAndUnpackCert(signedAtt.SigningCert, co)
-	if err != nil {
+	// Verify the certificate chain, and that the certificate was valid at the time of signing.
+	if err := sigstoreVerify.VerifyLeafCertificate(signatureTimestamp, *cert, trustedRoot); err != nil {
+		fmt.Fprintf(os.Stderr, "error verifying leaf certificate with sisgtore-go: %v\n", err)
```
haydentherapper, line 388: I missed that this is not a 1-1 swap between APIs. `ValidateAndUnpackCert` also checks the SCT on the certificate, so you'll need to call `VerifySignedCertificateTimestamp(cert, 1 /*threshold*/, trustedRoot` as well.

haydentherapper: This is also missing the identity check, which happened in a deeper call in `CheckCertificatePolicy` in `ValidateAndUnpackCert`.

For sigstore-go, you'll need to create an identity, `certID := NewShortCertificateIdentity(expected-oidc-issuer, expected-oidc-issuer-regex, expected-san, expected-san-regex)`, and you can use `""` for any that you don't want to check. You can then call `certID.Verify(summary)`, where `summary := certificate.SummarizeCertificate(leafCert)` [code](https://github.com/sigstore/sigstore-go/blob/main/pkg/verify/signed_entity.go#L531)

haydentherapper (changes requested):

go.mod
ramonpetgrave64: added

verifiers/internal/gha/rekor.go
ramonpetgrave64: Thanks. Added

verifiers/utils/sigstore_tuf.go
ramonpetgrave64: It makes sense to do that, but I feel we need to change the `NewLiveTrustedRoot` to accept an existing TUFClient, rather than make a new one. wdyt?

- https://github.com/sigstore/sigstore-go/blob/2b6fc6df45dc26c329f4da6c8157340e0f435a0d/pkg/root/trusted_root.go#L310

from 

```
func NewLiveTrustedRoot(opts *tuf.Options) (*LiveTrustedRoot, error)
```

to

```
func NewLiveTrustedRoot(c *tuf.Client) (*LiveTrustedRoot, error)
```

ramonpetgrave64: IIUC, we don't need to explicitly cache TrustedRoot, because the TUFClient will already cache it with GetTarget(). Is that right?

ramonpetgrave64: following up in https://github.com/sigstore/sigstore-go/issues/249

haydentherapper: `GetTarget` caches the target file to prevent redownloading it. Separately, the TUF client will still check if the local TUF repository metadata needs to be updated. If caching is configured (https://github.com/sigstore/sigstore-go/blob/main/pkg/tuf/options.go#L47), then the client will only fetch new targets once the local timestamp has expired or the cache validity window has passed.

For an API to accept a TUF client, one issue is that the underlying Updater can only have Refresh called once (https://github.com/theupdateframework/go-tuf/blob/f95222bdd22d2ac4e5b8ed6fe912b645e213c3b5/metadata/updater/updater.go#L105) (I'm not sure why that is the decision, we can ask Fredrik on the issue). If the live updater accepts an existing TUF client, it will need to reinitialize the `Updater`, which will need the options.

haydentherapper: Since the current implementation just supports verifying from the public instance, WDYT about proceeding with using `NewLiveTrustedRoot` as-is, since a custom TUF client shouldn't be necessary?

ramonpetgrave64: Added. 

It seems counter to our other plans for allowing slsa-verifier users to pass their own TUF client. But slsa-verifier needs to do live lookups against the public Sigstore infra, yes, I would be okay with not using the user-provided TUF client.

- https://github.com/slsa-framework/slsa-verifier/pull/768

haydentherapper (approved):
Thanks, looks great! I’ll follow up about the API and TUF clients on the other PR.

Issue #487, opened 2023-02-16
[feature] harden verification


laurentsimon:
we currently don't verify the cert in the bundle is the same as the one in the rekor entry, we only verify the signatures are the same https://github.com/slsa-framework/slsa-verifier/blob/main/verifiers/internal/gha/bundle.go#L175-L183

We should probably verify the certs match as well.

ianlewis:
Is there a real possibility that the two signatures would be exactly the same but that multiple certificates would be able to verify it? I suppose that there aren't necessarily any cryptographic guarantees but I wonder if this is actually exploitable in practice.

asraa:
Yes, technically the signatures matching is the only requirement - although this is a "SHOULD" as an extra robustness pass. Verifying the sigs match implicitly verifies the public key of the cert, since, except at very statistically improbable events, no two pks should give the same sigs on the same given message (the attestation). 

It is possible the cert content may be different with the public key the same, and that's indicating a key compromise during the 10 min window where the Fulcio cert is valid (which we cannot protect against) or Fulcio misbehavior.

laurentsimon:
> Verifying the sigs match implicitly verifies the public key of the cert, since, except at very statistically improbable events, no two pks should give the same sigs on the same given message (the attestation).

I don't think that's necessarily true. It's not a property of signature schemes, in particular if an attacker can select the public key. With ECDSA for example, it's easy to compute signature collision if you select the public key, I think it's just a matter of re-arranging the equation with the pubKey as the unknown variable.

PR #799, opened 2024-08-13
feat: handle dssev001 tlog entry types

ramonpetgrave64:
re: https://github.com/slsa-framework/slsa-github-generator/issues/3750

Rekor TLog entries can now be of the type dsse v0.0.1, as when what's returned when using sigstore-go's `Bundle()`.

This is to support eventual Sigstore Bundles produced by slsa-github-generator's "generic" generator, which will likely use sigstore-go's Bundle to produce attestations

 - https://github.com/slsa-framework/slsa-github-generator/compare/main...ramonpetgrave64-internal-builder-sigstore-bundlev2#diff-b186a0c5d9ae459b11b694f05455568453699670926d21cad06cafec3dbf895eR101
 - https://github.com/slsa-framework/slsa-github-generator/actions/runs/10359750833

## Tesing

- Added unit tests with stub data
- manual invocations to very both new and old attestations and bundles, with some modifications for testing purposes
  - https://github.com/slsa-framework/slsa-verifier/compare/main...verify-sigstore-go-Bundlev3#diff-94741068472ee694a12811cd704179dd478a9fa20a3bf45cf6ea2d4406214dc2R179

## Followup

Finish the work to produce bundles from the generic generators
 - https://github.com/slsa-framework/slsa-github-generator/compare/main...ramonpetgrave64-internal-builder-sigstore-bundlev2#diff-b186a0c5d9ae459b11b694f05455568453699670926d21cad06cafec3dbf895eR101


verifiers/internal/gha/bundle.go
```diff
@@ -156,6 +164,41 @@ func matchRekorEntryWithEnvelope(tlogEntry *v1.TransparencyLogEntry, env *dsseli
...
+	}
+	if err := json.Unmarshal(specMarshal, &dsseSchemaObj); err != nil {
+		return fmt.Errorf("%w: %s", ErrorUnexpectedEntryType, err)
+	}
+	if len(env.Signatures) != len(dsseSchemaObj.Signatures) {
+		return fmt.Errorf("expected %d sigs in canonical body, got %d",
+			len(env.Signatures),
+			len(dsseSchemaObj.Signatures))
+	}
+	// TODO(#487): verify the certs match.
```
ramonpetgrave64, line 187: Copied from the original method. Not all envelopes will have certificates, so that implementation would be opportunistic.

verifiers/internal/gha/bundle.go
```diff
@@ -156,6 +164,41 @@ func matchRekorEntryWithEnvelope(tlogEntry *v1.TransparencyLogEntry, env *dsseli
...
 
+// matchRekorEntryWithEnvelopeDSSEv001 handles matchRekorEntryWithEnvelope for the dsse v0.0.1 type version.
+func matchRekorEntryWithEnvelopeDSSEv001(tlogEntry *v1.TransparencyLogEntry, env *dsselib.Envelope) error {
+	canonicalBody := tlogEntry.GetCanonicalizedBody()
+	var dsseObj models.DSSE
+	if err := json.Unmarshal(canonicalBody, &dsseObj); err != nil {
+		return fmt.Errorf("%w: %s", ErrorUnexpectedEntryType, err)
+	}
+	var dsseSchemaObj models.DSSEV001Schema
+	specMarshal, err := json.Marshal(dsseObj.Spec)
```
loosebazooka, line 175: this feels odd? are we using json as an intermediate layer while converting between objects? I may not know enough about to go to understand if this is normal though.

verifiers/internal/gha/bundle.go
```diff
@@ -156,6 +164,41 @@ func matchRekorEntryWithEnvelope(tlogEntry *v1.TransparencyLogEntry, env *dsseli
...
+	}
+	if len(env.Signatures) != len(dsseSchemaObj.Signatures) {
+		return fmt.Errorf("expected %d sigs in canonical body, got %d",
+			len(env.Signatures),
+			len(dsseSchemaObj.Signatures))
+	}
+	// TODO(#487): verify the certs match.
+	for _, sig := range env.Signatures {
+		var matchCanonical bool
+		for _, canonicalSig := range dsseSchemaObj.Signatures {
```
loosebazooka, line 190: can we use the new slices.contains functionality here? https://pkg.go.dev/slices#ContainsFunc

```
slices.ContainsFunc(dsseSchemaObj.Signatures, func(canonicalSig <type>) bool {
		canonicalSig.Signature == sig.Sig
	})
```

or something? I'm not super sure how this works.

loosebazooka (commented):
Just some minor things

verifiers/internal/gha/bundle.go
ramonpetgrave64: Yeah, the docs say that the verifier has to parse out the objects to match the sigatures

 - https://pkg.go.dev/github.com/sigstore/protobuf-specs/gen/pb-go/rekor/v1#TransparencyLogEntry
 
And then If you see the test cases for `CanonicalizedBody`, the objects with the signaures are wrapped within another object.

 - github.com/sigstore/rekor/pkg/generated/model

But then the "Spec"'s `DSSESchema` is just an interface with no methods, so we have to marshall that again to get the data.

 - https://pkg.go.dev/github.com/sigstore/rekor@v1.3.6/pkg/generated/models#DSSESchema
 - https://pkg.go.dev/github.com/sigstore/rekor@v1.3.6/pkg/generated/models#DSSEV001Schema

verifiers/internal/gha/bundle.go
ramonpetgrave64: Added. It's not super clean, because I would have to reference the types of the inner objects, but it's a bit more readable, I think.

verifiers/internal/gha/bundle.go
```diff
@@ -156,6 +164,41 @@ func matchRekorEntryWithEnvelope(tlogEntry *v1.TransparencyLogEntry, env *dsseli
 	return nil
 }
 
+// matchRekorEntryWithEnvelopeDSSEv001 handles matchRekorEntryWithEnvelope for the dsse v0.0.1 type version.
+func matchRekorEntryWithEnvelopeDSSEv001(tlogEntry *v1.TransparencyLogEntry, env *dsselib.Envelope) error {
+	canonicalBody := tlogEntry.GetCanonicalizedBody()
+	var dsseObj models.DSSE
+	if err := json.Unmarshal(canonicalBody, &dsseObj); err != nil {
+		return fmt.Errorf("%w: %s", ErrorUnexpectedEntryType, err)
```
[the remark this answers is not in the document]
ramonpetgrave64: fixed. Added a few errors and tests for them.

verifiers/internal/gha/bundle.go
```diff
@@ -111,12 +111,20 @@ func getLeafCertFromBundle(bundle *bundle_v1.Bundle) (*x509.Certificate, error)
...
+
+	if kindVersion.Kind == "intoto" && kindVersion.Version == "0.0.2" {
+		return matchRekorEntryWithEnvelopeIntotov002(tlogEntry, env)
+	}
+
+	if kindVersion.Kind == "dsse" && kindVersion.Version == "0.0.1" {
+		return matchRekorEntryWithEnvelopeDSSEv001(tlogEntry, env)
 	}
 
+	return fmt.Errorf("%w: %s: %s", ErrorUnexpectedEntryType, kindVersion.Kind, kindVersion.Version)
```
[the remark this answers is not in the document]
ramonpetgrave64: added

loosebazooka (approved):
Looks good to me. Dunno if you want a more "go-y" person to do a review too

ramonpetgrave64:
@cmurphy, please take a look

verifiers/internal/gha/bundle.go
haydentherapper: Can we reuse https://github.com/slsa-framework/slsa-verifier/blob/1694bbf872a0a51f32de251779e4533231a53342/verifiers/internal/gha/rekor.go#L146 to do this comparison?
It's best practice to compare the canonicalized entry for Rekor to what you have (e.g the artifact hash, certificate, and signature, or computing the leaf hash yourself and comparing), since there is a risk of someone swapping out an inclusion proof that is valid but for the wrong entry. Of course, the chance of two signatures being the same is very low, so there isn't really a threat here.

haydentherapper (approved):

ramonpetgrave64:
@haydentherapper We can't yet because for the older attestations produce by slsa-github-generator, the certificate was not embedded within the envelope. 

But that seems like another good reason for https://github.com/slsa-framework/slsa-verifier/issues/487

Issue #806, opened 2024-09-05
Installer Action Node16 Deprecation Notice


IAreKyleW00t:
I noticed that the installer Action is still using Node16 and throws a warning at users since it has been deprecated. Would it be possible to bump this to Node20? Looking at the code, I don't think you're doing anything funky that would be a problem with an upgrade.

Side-note, but somewhat related, I noticed the Action doesn't support other platforms even though slsa-verifier does, and is a bit dated in general. I would be happy to volunteer to give it some love and have things align with some newer standards/practices. 😄

ramonpetgrave64:
Yes, we would welcome a PR. By other platforms, do you mean other OSs?

IAreKyleW00t:
@ramonpetgrave64 - Yes, I mean support for Windows and macOS. I see binaries are built for these platforms so I figured it would make sense to have the Action also support them.

ramonpetgrave64:
Alright, it could take a bit of work to get it to work for other OSs, but the node20 upgrade seems simple enough. 

 - https://github.com/slsa-framework/slsa-verifier/blob/767ecf9e0a63f5b7699d023609bd01978ff00d75/actions/installer/action.yml#L26

You're welcome to test it on your fork and submit a PR. I'll be happy to merge.

ramonpetgrave64:
Thanks for testing it out. 

I can't say for sure if it's intentional. But looking at Github's example, it seems like `tc.downloadTool()` is being used properly, and the other method calls aren't necessary. Does this sound right?

 - https://docs.github.com/en/actions/sharing-automations/creating-actions/developing-a-third-party-cli-action

Regarding the ref, it will actually work if the user wants to install older versions of slsa-verifier.

> Users can still pin to a release tag if they want. The version input is optional, and pinning to a release tag will install whichever version was latest at the time of release (i.e. @1.2.3 will install v1.2.3).

IAreKyleW00t:
> I can't say for sure if it's intentional. But looking at Github's example, it seems like `tc.downloadTool()` is being used properly, and the other method calls aren't necessary. Does this sound right?
> 
> * https://docs.github.com/en/actions/sharing-automations/creating-actions/developing-a-third-party-cli-action

[...]

PR #807, opened 2024-09-17
feat(action): Node20, multi-OS/Arch, caching, tests, and Workflows

IAreKyleW00t:
Hi!

This PR relates to the discussion from #806 regarding the Node16 deprecation notice.
During this we talked about adding support for multiple OSes, as well as addressing
the caching logic. There was quite a bit of change required to make this work,
so I took the liberty of updating others files to better reflect most
TypeScript Actions now-a-days and adding many more test cases.

**There are no breaking changes.** Everything will continue to work without
change for existing users, just with a few enhancements. I was actually able to
plug in my fork on some personal projects without any changes.

I hope I'm not overstepping with a PR of this size but I wanted to give a "complete package" - I had some fun fleshing this
out so I may have gone a bit overboard too haha 😇 Happy to make any adjustments

Action CI: https://github.com/IAreKyleW00t/slsa-verifier/actions/runs/10911993495

Test PR w/ bad dist/ changes: https://github.com/IAreKyleW00t/slsa-verifier/actions/runs/10911978180
Update dist/ with Workflow: https://github.com/IAreKyleW00t/slsa-verifier/actions/runs/10911989035
Re-test PR: https://github.com/IAreKyleW00t/slsa-verifier/actions/runs/10912006068

### Node20

Obviously this is on Node20, but wanted to make sure it was called out that the
`actions/installer/.node-version` file is what defines the
version to be used by CI and developer locals.

### README updates

The README has been updated with more information on the Action itself. Since
users are already directed there from the main README, I figured it would be
worth updating this to include details like Inputs, Outputs, usage, and even
Developer guides.

### Multi-OS and Arch support

As mentioned, since the main SLSA Verifier project builds binaries for Linux,
macOS, and Windows, I thought it would be good to have this action cover this
too.

The `actions/installer/src/bootstrap.ts` file defines updated SHA256
values and includes values for Windows, macOS, and Linux, along with their amd64
and arm64 variants.

The only caveat to this support is that the SLSA Verifier only started releasing
builds for Windows and macOS as of `v2.1.0`, so attempting to use this Action
with an older version would result in an error (which is tested and called out
in the README).

### Versioning

There is now a `version` input to allow the user to install specific version of
the SLSA verifier. By default this with use the `GITHUB_ACTION_REF` environment
variable like the previous verison of this Action would. There is a test
included to ensure this sort of logic will work.

Inputs that are allow are:

- `latest` - Will look up the latest Release and use that.
- _none_ - (Default) Uses the `GITHUB_ACTION_REF` env, like mentioned above.
- `vX.Y.Z` - a valid Semantic version tag with the `v` prefix. This will be
  validated to ensure the version _exists_.
- SHA - A full 40 character SHA can be used to support the existing
  functionality with the `GITHUB_ACTION_REF`. This will look through all
  Semantic Version tags with pagination for a matching commit. It will also make
  sure the tag it matches is valid version _and_ has a valid release.

### Caching

I saw that the `@action/tool-cache` package was being used but it was missing a
crutial step to actually utilize the Runners cache. I've updated the logic to
actually cache files, so existing binaries on the runners will be used to speed up the process.
**Cached binaries will always be validated.**

I also added logic to skip re-installing the SLSA Verifier if the version is the
same as the bootstrap version. Since we already validate that directly with it's
SHA256 we can simply re-use it.

The bootstrap binary is cached the same as the main binary.

You can use the `cache` input to manually override the caching behavior. If you
use `false` it will still use the Runner tool-cache location, but will
**always** overwrite the cache. This is true for the bootstrapper too.

### Installation PATH

Because we're properly using the Runner tool-cache the final binary no longer
needs to be in the users `HOME` directory. Instead the PATH will be updated with
the Runners cache location, like mentioned above.

This should have no impact on anyone unless they were strictly relying on having
the bin in their `HOME` vs in their `PATH`.

### Testing, Linting, Formating

The standard `eslint` Linter and `prettier` Formatter (that are used with most modern TypeScript
Actions) have been included to keep things aligned automatically. Most of the code
aligned to this already since I don't think too much changed in that regards,
but the tooling has been updated and validated.

This is included as part of the `npm run all` command and the Workflows, but can
be triggered individually with `npm run format:write` and `npm run lint`

Test coverage has been completely overhauled to include more test cases and
coverage and much more thorough integration testing with the Action itself.
There are **30+ TypeScript tests** (~5s), including **30+ Integration tests**
(60s) with a workflow where it is not easy to mock locally.

This is incorporated into the `npm run all` command and the Workflows, but can
be triggered individually with `npm run test`.

The `Makefile` was updated to include the equivalent commands.

### Workflows

Both the `e2e.schedule.installer.yml` and `update-actions-dist-post-commit.yml`
have been completely updated to support this new setup. They will "just work".
The E2E Workflow is where the Integration tests take place, which include a
number of matrixed operations to fully validate the Actions various cases across
every supported platform.

The E2E Installer Action CI will trigger on any changes to the Workflow file itself
or on any file changes in the `actions/installer/` directory, on the `main` branch.

The `update-actions-dist-post-commit.yml` Workflow no longer generates a `.patch` file. Instead,
it will move the entire `dist/` folder and `git add` them in the job with higher access.

The previous Worklfows appeared to have been failing, so
I needed to fix them in order to properly test things anyways.

### Misc

Other files in the root of the `actions/installer` that were changes were
related to linters, testing, etc. They are just part of a TypeScript Action.

The source files are updated with the SLSA Apache-2.0 LICENSE notice at the top
of each source `.ts` file.

`jasmine` testing suite replaced with `jest` and `lib` is replaced with the newer `dist` naming convention


IAreKyleW00t:
Now noticing that the `pre-submit.actions.yml` may be a better place for some of the workflow jobs. Let me know if any of those should be moved around.

hivoltazh007 (approved):

ramonpetgrave64:
Thanks for submitting the PR. To make reviewing easier, please revert the formatting changes across the files.
Also let's deal with one problem at a time: Let's have one PR about only fixing the node16 deprecation, and then we can followup about multi-os, caching, and other issues.

PR #831, opened 2025-02-11
fix: shell completion improvements

scop:
See individual commits and their messages for details.


cli/slsa-verifier/verify/options.go
```diff
@@ -145,27 +162,34 @@ var _ Interface = (*VerifyVSAOptions)(nil)
 func (o *VerifyVSAOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().StringArrayVar(&o.SubjectDigests, "subject-digest", []string{},
 		"the digests to be verified. Pass multiple digests by repeating the flag. e.g. --subject-digest <digest type>:<digest value> --subject-digest <digest type>:<digest value>")
+	cmd.RegisterFlagCompletionFunc("subject-digest", cobra.NoFileCompletions)
 
 	cmd.Flags().StringVar(&o.AttestationPath, "attestation-path", "",
 		"path to a file containing the attestation")
+	cmd.MarkFlagFilename("attestation-path", "intoto.jsonl")
```
ramonpetgrave64, line 169: The filename can actually have any extension.

cli/slsa-verifier/verify/options.go
```diff
@@ -145,27 +162,34 @@ var _ Interface = (*VerifyVSAOptions)(nil)
 func (o *VerifyVSAOptions) AddFlags(cmd *cobra.Command) {
 	cmd.Flags().StringArrayVar(&o.SubjectDigests, "subject-digest", []string{},
 		"the digests to be verified. Pass multiple digests by repeating the flag. e.g. --subject-digest <digest type>:<digest value> --subject-digest <digest type>:<digest value>")
+	cmd.RegisterFlagCompletionFunc("subject-digest", cobra.NoFileCompletions)
```
ramonpetgrave64, line 165: If we don't add this line, would the behavior be the same?

scop: No, it will suggest all filenames in the current dir as completions.

cli/slsa-verifier/verify/options.go
scop: Ok, removed this filter.

ramonpetgrave64: Thanks, also remove it for the other subcommands.

ramonpetgrave64: I see we can actually specify multiple extensions, so how about using the common ones: `.sigstore`, `.intoto`, `.intoto.jsonl`, `.json`.

 - https://github.com/spf13/cobra/blob/680936a2200be363c61feda8cd29287f0726a48c/shell_completions.go#L42-L46

scop: Done for both related flags. Thanks for the guidance, I'm familiar with cobra but not that much with all things SLSA.

I suppose it should be noted that at least with bash (well readline more accurately), the user can invoke completion of all filenames by hitting Meta-/ (typically Esc-/) instead of Tab. Even though that's a bit inconvenient and not too well known combo, at least there is an escape hatch if we get some of the file extension filters wrong.

cli/slsa-verifier/verify.go
```diff
@@ -72,6 +72,7 @@ func verifyArtifactCmd() *cobra.Command {
 	o.AddFlags(cmd)
 	// --provenance-path must be supplied when verifying an artifact.
 	cmd.MarkFlagRequired("provenance-path")
+	cmd.MarkFlagFilename("provenance-path", "intoto.jsonl")
```
ramonpetgrave64, line 75: Extra extensions here, as well.

cli/slsa-verifier/verify/options.go
```diff
@@ -50,26 +50,34 @@ func (o *VerifyOptions) AddFlags(cmd *cobra.Command) {
...
+	cmd.RegisterFlagCompletionFunc("source-tag", cobra.NoFileCompletions)
 
 	cmd.Flags().StringVar(&o.SourceVersionTag, "source-versioned-tag", "",
 		"[optional] expected version the binary was compiled from. Uses semantic version to match the tag")
+	cmd.RegisterFlagCompletionFunc("source-versioned-tag", cobra.NoFileCompletions)
 
 	/* Other options */
 	cmd.Flags().StringVar(&o.ProvenancePath, "provenance-path", "",
 		"path to a provenance file")
+	cmd.MarkFlagFilename("provenance-path", "intoto.jsonl")
```
ramonpetgrave64, line 76: Extra extensions here, as well.

cli/slsa-verifier/verify.go
```diff
@@ -136,6 +138,12 @@ func verifyNpmPackageCmd() *cobra.Command {
 			}
 			return nil
 		},
+		ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+			if len(args) != 0 {
+				return nil, cobra.ShellCompDirectiveNoFileComp
+			}
```
ramonpetgrave64, line 144: This works like a shortcut to say that all of options shouldn't offer completion?

scop: It means that if we already have one positional argument, do not offer any more, as `verify-npm-package` takes only one positional argument.

cli/slsa-verifier/verify/options.go
scop: Done.

cli/slsa-verifier/verify/options.go
```diff
@@ -45,31 +45,41 @@ type VerifyOptions struct {
 
 var _ Interface = (*VerifyOptions)(nil)
 
+var CommonFilenameExtensions = []string{"sigstore", "intoto", "intoto.jsonl", "json"}
```
ramonpetgrave64, line 48:
```suggestion
var CommonAttestationFilenameExtensions = []string{"sigstore", "intoto", "intoto.jsonl", "json"}
```

ramonpetgrave64 (changes requested):
Looks okay so far

cli/slsa-verifier/verify/options.go
scop: This is also being used for provenance filenames per earlier comments. Do we still want to rename it as suggested?

scop:
> Remember to signoff your commits.

Done. This would be useful info to have in docs/CONTRIBUTING.md

cli/slsa-verifier/verify/options.go
ramonpetgrave64: Yes, provenance and attestation are mostly interchangeable terms, but attestation is more generic.

scop: Done.

ramonpetgrave64 (approved):

scop:
Hi @ramonpetgrave64, others, thank you for the review and approval. I'm wondering if this is pending something still from me or otherwise, or if it could be merged before it could grow conflicts or become stale some other way?

PR #840, opened 2025-03-11
feat: verify provenance for bcr modules produced by trusted reusable workflows

loosebazooka:
@fweikert these are the changes I think might be needed to get this to work (it's somewhat hacky, I'm not sure I've fully covered what's needed).

@ramonpetgrave64 is this kinda what's needed?

This now adds the `verify-github-attestation` sub command. Use this instead of `verify-artifact`.


loosebazooka:
well we're currently on the dev branch of the builder, so I don't think we want to submit this in anyway. Once we finalize the builder I'll actually add tests.

alexeagle:
FYI we have the first Bazel module published with attestations, and the presubmit fails since the reusable workflow isn't allow-listed here: https://github.com/bazelbuild/bazel-central-registry/pull/4060

ramonpetgrave64 (changes requested):

cli/slsa-verifier/main_regression_test.go
```diff
@@ -1510,6 +1510,66 @@ func Test_runVerifyGHAContainerBased(t *testing.T) {
 	}
 }
 
+func Test_runVerifyBCRModule(t *testing.T) {
+	t.Parallel()
+	os.Setenv("SLSA_VERIFIER_EXPERIMENTAL", "1")
```
ianlewis, line 1515: Use `t.Setenv` so that the environment variable is restored automatically up after the test. Do not use `t.Parallel` as it will cause race condition with the environment variables. See [`t.Setenv`](https://pkg.go.dev/testing#T.Setenv)
```suggestion
	t.Setenv("SLSA_VERIFIER_EXPERIMENTAL", "1")
```

cli/slsa-verifier/main_regression_test.go
```diff
@@ -1510,6 +1510,66 @@ func Test_runVerifyGHAContainerBased(t *testing.T) {
...
+			source:   "github.com/aspect-build/rules_lint",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			artifactPath := filepath.Clean(filepath.Join(TEST_DIR, "bcr", tt.artifact))
+			// TODO: this only handles the single attestation case in tests, so *.intoto.jsonl is essentially *.json
```
ianlewis, line 1556: nit: Reference an issue # in the todo. e.g. `// TODO(#1234): this only ...`

cli/slsa-verifier/main.go
```diff
@@ -35,6 +35,7 @@ For more information on SLSA, visit https://slsa.dev`,
 	}
 	c.AddCommand(version.Version())
 	c.AddCommand(verifyArtifactCmd())
+	c.AddCommand(verifyBCRModuleCmd())
```
ianlewis, line 38: It looks like the workflow is just using `actions/attest-build-provenance` to create an attestation.

https://github.com/bazel-contrib/.github/blob/eeb03fb7518fb0036e4bb0713134c3b6531c68ba/.github/workflows/release_ruleset.yaml#L86-L91

What about this requires specific command for bcr modules? Is it just that we are trusting a specific builder ID? Could this be a more generic command? If we want to limit verifying provenance generated by `actions/attest-build-provenance` to specific trusted builder IDs, could we just have an option like `slsa-verifier verify-github-attestation --builder bazel-central-registry` or something along those lines?

cli/slsa-verifier/main_regression_test.go
```diff
@@ -1510,6 +1510,66 @@ func Test_runVerifyGHAContainerBased(t *testing.T) {
...
+		{
+			name:     "source archive no builder id",
+			artifact: "rules_lint-v1.3.1.tar.gz",
+			source:   "github.com/aspect-build/rules_lint",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
```
ianlewis, line 1553: `t.Parallel` can't be used with environment variables.
```suggestion
```

verifiers/internal/gha/verifier.go
```diff
@@ -242,6 +242,32 @@ func (v *GHAVerifier) VerifyArtifact(ctx context.Context,
...
+
+	var signedAtt *SignedAttestation
+	/* Verify signature on the intoto attestation. */
+	signedAtt, err = VerifyProvenanceBundle(ctx, provenance, trustedRoot)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return verifyEnvAndCert(signedAtt.Envelope, signedAtt.SigningCert,
+		provenanceOpts, builderOpts, defaultBCRReusableWorkflows)
```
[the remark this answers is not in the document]
ianlewis: I'm curious what about this is specific to BCR since it looks to be just using the GitHub's attestation action. I tend to agree that this seems like this functionality could be added a bit more generically.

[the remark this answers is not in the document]
loosebazooka: Yeah that was the initial pathway I took. I wanted to flag behind experimental and only allow specific builderids. The initial path was very generic, and maybe I can go back to that. I also want to only allow sigstore signed attestations

cli/slsa-verifier/testdata/bcr/MODULE.bazel
```diff
@@ -0,0 +1,34 @@
+"Bazel dependencies"
```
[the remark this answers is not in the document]
loosebazooka: Similar to a java pom.xml/go.mod or whatever. It lets bcr determine the dependency tree for a module.

cli/slsa-verifier/main.go
loosebazooka: I can change this to `verify-github-attestation` , I'm not deep enough into slsa land to understand the consequences here. Is the `builderId` spoofable in any way?

loosebazooka: oh right, builderId nees to match the SAN in the certificate

ianlewis: Yes. It's not spoofable because we check it against the certificate SAN (at least for sigstore which is all we support currently).

If we implemented private PKI in the future the consumer would presumably trust that the private key is only available to the build system that set the builder ID in that case.

cli/slsa-verifier/main_regression_test.go
```diff
@@ -1510,6 +1510,75 @@ func Test_runVerifyGHAContainerBased(t *testing.T) {
...
+		err       error
+	}{
+		{
+			name:      "module.bazel using publishing builder",
+			artifact:  "MODULE.bazel",
+			source:    "github.com/aspect-build/rules_lint",
+			builderID: bcrPublisherBuilderID,
+		},
+		{
+			name:      "source archive using release builder",
```
ramonpetgrave64, line 1534: For NPM, they put their build attestation and "publish" attestation together in one .jsonl file. Does it make sense for the bazel workflow(s) to do similarly?

cli/slsa-verifier/verify/verify_github_attestation.go
```diff
@@ -0,0 +1,78 @@
+// Copyright 2022 SLSA Authors
```
ramonpetgrave64, line 1:
```suggestion
// Copyright 2025 SLSA Authors
```

cli/slsa-verifier/verify/verify_github_attestation.go
```diff
@@ -0,0 +1,78 @@
...
+	"fmt"
+	"os"
+
+	"github.com/slsa-framework/slsa-verifier/v2/options"
+	"github.com/slsa-framework/slsa-verifier/v2/verifiers"
+	"github.com/slsa-framework/slsa-verifier/v2/verifiers/utils"
+)
+
+type VerifyGithubAttestationCommand struct {
+	AttestationPath     string
```
fweikert, line 30: Would it be useful to allow specification of branch and tag, like verify-artifact currently does?

loosebazooka: I don't think source-branch and source-tag work with github generated attestations at the moment

cli/slsa-verifier/verify/verify_github_attestation.go
```diff
@@ -0,0 +1,78 @@
...
+	BuilderID           *string
+	SourceURI           string
+	BuildWorkflowInputs map[string]string
+	PrintAttestation    bool
+}
+
+func (c *VerifyGithubAttestationCommand) Exec(ctx context.Context, artifact string) (*utils.TrustedBuilderID, error) {
+	if !options.ExperimentalEnabled() {
+		err := errors.New("feature support is only provided in SLSA_VERIFIER_EXPERIMENTAL mode")
```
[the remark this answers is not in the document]
loosebazooka: I think we should leave it for now. If we want to change the api surface or anything, I would like to do so with minimal deprecation windows.

loosebazooka:
@ianlewis any other thoughts here?

bobcallaway (approved):

ramonpetgrave64 (approved):

fweikert:
Thanks everyone! Do you have a planned date for the next release?

PR #850, opened 2025-06-10
feat: Bazel not experimental

### LINKED ISSUE #849: move `verify github attestation` out of SLSA_EXPERIMENTAL
The github attestation spec is stable. So there really shouldn't be too many issues here. If we want BCR users to use slsa-verifier, it would make sense to make this a GA feature.

The one issue with this is that we still have a pre-set list of blessed "github attestation" providers, but we could change [this section](https://github.com/slsa-framework/slsa-verifier/blob/f5b9bade6612a567f8f0ebf9b5494d567c1eef58/verifiers/internal/gha/slsaprovenance/v1.0/provenance.go#L84-L92) a bit to allow ANY builder for github attestations if we want wider coverage.

ramonpetgrave64:
Followup to #840 

Resolves #849 

Removes the experimental flag for verifying bazel attestations.

TODO:

- [ ] add example invocation for bazel https://github.com/slsa-framework/slsa-verifier/pull/858#pullrequestreview-2947145690
- [ ] create a new release


cli/slsa-verifier/verify.go
```diff
@@ -196,7 +196,7 @@ func verifyGithubAttestation() *cobra.Command {
 			}
 			return nil
 		},
-		Short: "Verifies SLSA provenance for a github attestation [experimental]",
+		Short: "Verifies SLSA provenance for a github attestation",
```
haydentherapper, line 199: github -> GitHub

I'd also call these "artifact attestations"

loosebazooka (approved):

PR #855, opened 2025-06-18
fix: no parallel regression tests

ramonpetgrave64:
Make the regression tests that do online access not run in parallel.
Besides VSA tests, all regression tests do some kind of online access of either TUF or Rekor.


ramonpetgrave64:
@slsa-framework/slsa-tool-reviewers

haydentherapper (approved):

PR #856, opened 2025-06-18
fix: rekor retries to 5

ramonpetgrave64:
Sets the retries for our rekor client to 5, not the default of 3.

Hopefully this makes tests less brittle.


haydentherapper (approved):

haydentherapper:
Closing, should be unnecessary now.

PR #863, opened 2025-07-01
feat: Add source tag verification to github attest based verifier

loosebazooka:


cli/slsa-verifier/main_regression_test.go
```diff
@@ -1526,12 +1526,57 @@ func Test_runVerifyGithubAttestation(t *testing.T) {
...
+			artifact:         "MODULE-on-tag.bazel",
+			source:           "github.com/aspect-build/rules_lint",
+			sourceVersionTag: pString("v1.3.4"),
+			builderID:        bcrPublisherBuilderID,
+		},
+		{
+			name:             "module.bazel using publishing builder and partial source versioned tag",
+			artifact:         "MODULE-on-tag.bazel",
+			source:           "github.com/aspect-build/rules_lint",
+			sourceVersionTag: pString("v1.3"),
```
ramonpetgrave64, line 1569: I realize there aren't any tests that check for tags that are pre-releases. But can you add a couple, so we can be sure about the behavior.

loosebazooka: hrmm.. okay lemme try to generate one

